PR #22, opened 2023-05-16
bash script to compile proto files

jjleng:
A bash script that is able to install grpc tools and compile .proto files. 

`./grpc-gen.sh typescript ~/Workspace/aptos-core/crates/aptos-protos/proto` generates files under ./generated/typescript
`./grpc-gen.sh python ~/Workspace/aptos-core/crates/aptos-protos/proto` generates files under ./generated/python
`./grpc-gen.shrust ~/Workspace/aptos-core/crates/aptos-protos/proto` generates files under ./generated/rust

### DIFF
--- a/scripts/grpc-gen.sh
+++ b/scripts/grpc-gen.sh
@@ -100,11 +100,21 @@ generate_python_code() {
   echo "Python code generated successfully!"
 }
 
+ensure_protobuf_codegen_installed() {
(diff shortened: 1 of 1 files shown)


bowenyang007:
can we finish this? Or what's the plan here @jjleng

jjleng:
The rust variant might work. I don't have time to fix it. Let's close it.

PR #24, opened 2023-05-18
Added aptos_ambassador_token example

junkil-park:
Added a minimal example of indexer processor for tracking the changes in the Aptos ambassador tokens' levels

### DIFF
--- a/python/processors/README.md
+++ b/python/processors/README.md
@@ -20,3 +20,18 @@ This example demonstrates the simplest possible variant of a processor. There ar
 It also demonstrates a parser function that outputs operations to create, update, and delete rows.
 
 This processor is part of a full Aptos project, so you can see the processor in context alongside the Move module and frontend.
+
+## Aptos Ambassador Token
+
+This example demonstrates how to track changes on the attribute (e.g., ambassador level) of the [Aptos ambassador token](https://github.com/aptos-labs/aptos-core/tree/main/aptos-move/move-examples/token_objects/ambassador/move). This example includes
+
+- Processor: Processes transactions and detects changes on the ambassador level in the events & writesets of the transactions
+- Database: No database operation is demonstrated for the simplicity sake. The changes found are simply printed out in the console.
+
+When an existing token's level changes, the smart contract is programmed to emit an event (e.g., LevelUpdateEvent). So, the level change can be observed both in the transaction's event and writeset redundantly.
+
+To run:
+```
+$ cd aptos-indexer-processors/python
+$ poetry run python -m processors.aptos_ambassador_token.processor -c config.yaml
+```
--- a/python/processors/aptos_ambassador_token/processor.py
+++ b/python/processors/aptos_ambassador_token/processor.py
@@ -1,0 +1,88 @@
+import argparse
+import grpc
+import ast
+import json
+
+from aptos.indexer.v1 import raw_data_pb2_grpc
+from aptos.indexer.v1 import raw_data_pb2
+from aptos.transaction.testing1.v1 import transaction_pb2
+from utils.config import Config
+from utils.general_utils import standardize_address
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "-c", "--config", help="Path to config file", required=True)
+args = parser.parse_args()
+config = Config.from_yaml_file(args.config)
+starting_version = 0
+if config.starting_version_default != None:
+    starting_version = config.starting_version_default
+if config.starting_version_override != None:
+    starting_version = config.starting_version_override
+metadata = (
+    ("x-aptos-data-authorization", config.indexer_api_key),
+    ("x-aptos-request-name", "ambassador token indexer"),
+)
+options = [("grpc.max_receive_message_length", -1)]
+
+qualified_module_name = "0x9bfdd4efe15f4d8aa145bef5f64588c7c391bcddaf34f9e977f59bd93b498f2a::ambassador"
+qualified_event_name = qualified_module_name + "::LevelUpdateEvent"
+qualified_resource_name = qualified_module_name + "::AmbassadorLevel"
+
+with grpc.insecure_channel(config.indexer_endpoint, options=options) as channel:
+    stub = raw_data_pb2_grpc.RawDataStub(channel)
+    current_transaction_version = starting_version
+
+    for response in stub.GetTransactions(
+        raw_data_pb2.GetTransactionsRequest(starting_version=starting_version),
+        metadata=metadata,
+    ):
+        assert response.chain_id == config.chain_id
+        for transaction in response.transactions:
+            assert transaction.version == current_transaction_version
+
+            found = False
+            if transaction.type == transaction_pb2.Transaction.TRANSACTION_TYPE_USER:
+                for event in transaction.user.events:
+                    addr = standardize_address(event.key.account_address)
+                    type = event.type_str
+                    data = event.data
+                    if type == qualified_event_name:
+                        found = True
+                        print('-------------')
+                        print('Event emitted')
+                        print('-------------')
+                        print('ambassador token addr:', addr)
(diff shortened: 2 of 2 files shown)


python/processors/aptos_ambassador_token/processor.py
```diff
@@ -0,0 +1,83 @@
...
+from aptos.indexer.v1 import raw_data_pb2
+from aptos.transaction.testing1.v1 import transaction_pb2
+from utils.config import Config
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "-c", "--config", help="Path to config file", required=True)
+args = parser.parse_args()
+config = Config.from_yaml_file(args.config)
+starting_version = 0
```
rtso, line 16: You can get the starting version from the Config object now. [Code pointer](https://github.com/aptos-labs/aptos-indexer-processors/blob/main/python/utils/config.py#L38)

python/processors/README.md
```diff
@@ -20,3 +20,12 @@ This example demonstrates the simplest possible variant of a processor. There ar
...
 This processor is part of a full Aptos project, so you can see the processor in context alongside the Move module and frontend.
+
+## Aptos Ambassador Token
+
+This example demonstrates how to track changes on a certain token attribute (e.g., ambassador level) using transactions' events and writesets. This example includes
+
+- Processor: Processes transactions and detects changes on the ambassador level in the events & writesets of the transactions
+- Database: No database operation is demonstrated for the simplicity sake. The changes found are simply printed out in the console.
+
+When an existing token's level changes, the smart contract is programmed to emit an event (e.g., LevelUpdateEvent). So, the level change can be observed both in the transaction's event and writeset redundantly.
```
rtso, line 31: Is the smart contract code public? If so, we can link here so people can see both contract and processor side by side

python/processors/aptos_ambassador_token/processor.py
```diff
@@ -0,0 +1,83 @@
...
+        metadata=metadata,
+    ):
+        assert response.chain_id == config.chain_id
+        for transaction in response.transactions:
+            assert transaction.version == current_transaction_version
+
+            found = False
+            if transaction.type == transaction_pb2.Transaction.TRANSACTION_TYPE_USER:
+                for event in transaction.user.events:
+                    addr = "0x" + event.key.account_address
```
rtso, line 43: For standardizing addresses, we should always use this [helper function](https://github.com/aptos-labs/aptos-indexer-processors/pull/9/files#diff-fb72abdf04c829b4d8b5a8108bd0b604f4c63adefe683dd3b154098e596eee90R15). I haven't merged it yet but feel free to add it in this PR.

python/processors/aptos_ambassador_token/processor.py
```diff
@@ -0,0 +1,83 @@
...
+                        print('-------------')
+                        print('ambassador token addr:', addr)
+                        data_dic = ast.literal_eval(data)
+                        print('old_level:', data_dic["old_level"])
+                        print('new_level:', data_dic["new_level"])
+                        print()
+
+                for change in transaction.info.changes:
+                    if change.type == transaction_pb2.WriteSetChange.TYPE_WRITE_RESOURCE:
+                        addr = change.write_resource.address
```
rtso, line 59: Need to standardize this address, just in case

python/processors/aptos_ambassador_token/processor.py
junkil-park: I couldn't use it because this example doesn't need to establish the DB connection which this line of code does: https://github.com/aptos-labs/aptos-indexer-processors/blob/main/python/utils/config.py#LL39C28-L39C28

python/processors/README.md
junkil-park: Yes, it's public. I will add a link.

rtso (approved):

python/processors/aptos_ambassador_token/processor.py
```diff
@@ -0,0 +1,85 @@
...
+from utils.config import Config
+from utils.general_utils import standardize_address
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "-c", "--config", help="Path to config file", required=True)
+args = parser.parse_args()
+config = Config.from_yaml_file(args.config)
+starting_version = 0
+if config.starting_version_default != None:
```
movekevin, line 18: @rtso Can all of this boilerplate config -> metadata/options conversion be done in a library we provide? This way people wouldn't need to write this every time they write a processor

python/processors/aptos_ambassador_token/processor.py
```diff
@@ -0,0 +1,85 @@
...
+if config.starting_version_override != None:
+    starting_version = config.starting_version_override
+metadata = (
+    ("x-aptos-data-authorization", config.indexer_api_key),
+    ("x-aptos-request-name", "ambassador token indexer"),
+)
+options = [("grpc.max_receive_message_length", -1)]
+
+
+with grpc.insecure_channel(config.indexer_endpoint, options=options) as channel:
```
movekevin, line 29: @rtso Same here. I think this can be abstracted into a library as well

python/processors/aptos_ambassador_token/processor.py
```diff
@@ -0,0 +1,85 @@
...
+        for transaction in response.transactions:
+            assert transaction.version == current_transaction_version
+
+            found = False
+            if transaction.type == transaction_pb2.Transaction.TRANSACTION_TYPE_USER:
+                for event in transaction.user.events:
+                    addr = standardize_address(event.key.account_address)
+                    type = event.type_str
+                    data = event.data
+                    if type == "0x9bfdd4efe15f4d8aa145bef5f64588c7c391bcddaf34f9e977f59bd93b498f2a::ambassador::LevelUpdateEvent":
```
movekevin, line 47: nit: can we make this a const?

python/processors/README.md
```diff
@@ -20,3 +20,12 @@ This example demonstrates the simplest possible variant of a processor. There ar
 It also demonstrates a parser function that outputs operations to create, update, and delete rows.
 
 This processor is part of a full Aptos project, so you can see the processor in context alongside the Move module and frontend.
+
+## Aptos Ambassador Token
```
movekevin, line 24: Can you add some instructions on how to run the processor code?

movekevin (approved):
This is awesome!

python/processors/README.md
junkil-park: Added.

python/processors/aptos_ambassador_token/processor.py
junkil-park: Done.

python/processors/aptos_ambassador_token/processor.py
bowenyang007: Yes agreed that stuff should def go into the general framework.

rtso: Yeah I've started abstracting some of the config stuff into its own class but this was a special case (no db writes) that isn't supported yet that I plan to look into. Same thing with connecting to grpc and returning a stream of txns; I want to turn this into a framework with least boilerplate as possible.

python/processors/aptos_ambassador_token/processor.py
```diff
@@ -0,0 +1,88 @@
...
+    stub = raw_data_pb2_grpc.RawDataStub(channel)
+    current_transaction_version = starting_version
+
+    for response in stub.GetTransactions(
+        raw_data_pb2.GetTransactionsRequest(starting_version=starting_version),
+        metadata=metadata,
+    ):
+        assert response.chain_id == config.chain_id
+        for transaction in response.transactions:
+            assert transaction.version == current_transaction_version
```
bowenyang007, line 42: Yea pretty much everything before this should be abstracted away.

PR #32, opened 2023-05-24
update the processor table creation.

larry-aptos:
* tested locally with fresh docker compose build, processor runs without problem.

image


larry-aptos:
@rtso going to merge it and we can do post-submission review... since i think docker image with one extra step to set up the table is annoying...

PR #36, opened 2023-06-20
rename config parameters and make indexer_name a processor-level constant

geekflyer:
This PR does a few things:
1. Fix crashing example processor (verified via docker compose) by adding Mapped[str] to event column names: https://github.com/aptos-labs/aptos-indexer-processors/pull/36/files#diff-a95726e6e182fd11c7bcaf79632ec6d3e8304f33b3f13efe5e3a845a3c299568L20-L21
2. rename config parameters to more clearly distinguish between processor-specific config and config to connect to remote indexer-grpc data stream service.
3. make indexer_name a processor level constant instead of runtime config.

### DIFF
--- a/python/processors/example_event_processor/models.py
+++ b/python/processors/example_event_processor/models.py
@@ -1,3 +1,4 @@
+from utils.models.annotated_types import StringType
 from utils.models.annotated_types import (
     StringPrimaryKeyType,
     BigIntegerType,
@@ -17,7 +18,7 @@ class Event(Base):
     transaction_version: BigIntegerType
     transaction_block_height: BigIntegerType
     transaction_timestamp: TimestampType
-    type: str
-    data: str
+    type: StringType
+    data: StringType
     inserted_at: InsertedAtType
     event_index: BigIntegerType
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Aptos Indexer Client Guide
 This guide will get you started with creating an Aptos indexer with custom parsing. We have several endpoints that provided a streaming RPC of transaction data. 
 
-## Indexer Endpoints(all endpoints are in GCP us-central1 unless specified)
-* devent: 35.225.218.95:50051
+## GRPC Data Stream Endpoints(all endpoints are in GCP us-central1 unless specified)
+* devnet: 35.225.218.95:50051
 
 * testnet: 35.223.137.149:50051
   * Asia(GCP asia-northeast3): 34.64.252.224:50051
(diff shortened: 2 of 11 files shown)


rtso:
The `processor_name` got moved to the config level to allow deploying the same processor with different configurations. Their `NextVersionToProcess` will be tracked differently. For example:
- configuration 1
  - grpc_data_stream_endpoint: mainnet
  - processor_name: mainnet-nft-activities
- configuration 2
  - grpc_data_stream_endpoint: devnet
  - processor_name: devnet-nft-activities
With moving `processor_name` to the processor level, you'd have to create multiple processor files to achieve the same effect.

geekflyer:
the original motivation of this PR was actually to fix https://github.com/aptos-labs/aptos-indexer-processors/pull/36/files#diff-a95726e6e182fd11c7bcaf79632ec6d3e8304f33b3f13efe5e3a845a3c299568L20-L21

> Can you help me understand why this change is necessary?
> 
> The `processor_name` got moved to the config level to allow deploying the same processor with different configurations. Their `NextVersionToProcess` will be tracked differently. For example:
> 
> * configuration 1
>   
>   * grpc_data_stream_endpoint: mainnet
>   * processor_name: mainnet-nft-activities
> * configuration 2
>   
>   * grpc_data_stream_endpoint: devnet
>   * processor_name: devnet-nft-activities
> 
> With moving `processor_name` to the processor level, you'd have to create multiple processor files to achieve the same effect.

Sure, just added a bit more description. Re the change of processor_name in particular:
I believe it doesn't really make sense to make this a runtime configuration and just adds to confusion to somebody who wants to deploy this.
The use-case you mentioned above of deploying the same processor with different configs using the same db won't work as both processors would attempt to write to the same nft/token tables and unless you make the processor name or chain name an extra column in each destination table it will be impossible to distinguish the data between different processor deployments then (furthermore you might run into conflicts).

The only scenario 2 processor deployments with different config make sense is if you let connect them to different database servers, logical dbs or different schemas, in which case they will all use distinct NextVersionToProcess table anyways, so having the same processor name for different deployments is not a problem.

python/processors/example_event_processor/models.py
```diff
@@ -17,7 +18,7 @@ class Event(Base):
     transaction_version: BigIntegerType
     transaction_block_height: BigIntegerType
     transaction_timestamp: TimestampType
-    type: str
-    data: str
+    type: Mapped[str]
```
rtso, line 21: Can you import and use `StringType` for this?

rtso (approved):
Thanks for the explanation, makes sense!

PR #39, opened 2023-06-21
Add souffle and bluemove parsing

rtso:
## Description
Add listing, token bid, and collection bid parsing for Souffle and Bluemove marketplaces.

## Testing
`poetry run python -m processors.nft_orderbooks.nft_marketplace_processor -c config.yaml`

## DQ Checks
Note: since the DQ check is only done on a subset of data, some values in total_liq_impact_vol are wrong because they rely on lookups on past data 
#### nft_orderbooks (Bigquery)
souffle listing dq comp
#### nft_marketplace_listings
souffle listing dq

#### nft_orderbooks (Bigquery)
bluemove dq comp
#### nft_marketplace tables 
nft_marketplace_listings
bluemove listing
nft_marketplace_bids
bluemove bids
nft_marketplace_collection_bids
bluemove collection bids


python/processors/nft_orderbooks/parsers/souffle_parser.py
```diff
@@ -1,108 +1,280 @@
...
+        transaction_version=transaction_metadata.transaction_version,
+        event_index=event_index,
+        event_type=display_event_type,
+        standard_event_type=standardize_marketplace_event_type(
+            display_event_type
+        ).value,
+        creator_address=standardize_address(creator),
+        collection=token_data_id_type.get_collection_trunc(),
+        token_name=token_data_id_type.get_name_trunc(),
+        token_data_id=token_data_id_type.to_hash(),
```
bowenyang007, line 90: actually you might want to add '0x' in front of this to enable join w/ the rest of the data. To confirm check current_token_ownerships_v2

rtso: The `to_hash` function calls `standardize_address` so the address will be `0x` prefixed. But good point, let me check that we're doing it correctly.

rtso: Checked a random NFT listing - the token_data_id and seller address in the listing matches with a row in current_token_ownership_v2.

| current_token_ownerships_v2 | current_nft_listings | 
| --- | --- |
| Screenshot 2023-06-23 at 10 59 33 AM | Screenshot 2023-06-23 at 10 59 51 AM |

bowenyang007 (approved):

PR #40, opened 2023-06-29
Add schema for nft_marketplace

rtso:
## Testing 
Screenshot 2023-06-29 at 3 38 39 PM

### DIFF
--- a/python/processors/nft_orderbooks/models/nft_marketplace_bid_models.py
+++ b/python/processors/nft_orderbooks/models/nft_marketplace_bid_models.py
@@ -15,6 +15,7 @@ from utils.models.general_models import Base
 
 class CurrentNFTMarketplaceBid(Base):
     __tablename__ = "current_nft_marketplace_bids"
+    __table_args__ = {"schema": "nft_marketplace"}
 
     token_data_id: StringPrimaryKeyType
     buyer: StringPrimaryKeyType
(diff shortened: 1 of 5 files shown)


bowenyang007 (approved):

python/processors/nft_orderbooks/models/nft_marketplace_bid_models.py
```diff
@@ -35,6 +36,7 @@ class CurrentNFTMarketplaceBid(Base):
 
 class NFTMarketplaceBid(Base):
     __tablename__ = "nft_marketplace_bids"
+    __table_args__ = {"schema": "nft_marketplace"}
```
geekflyer, line 39: can we move this to a shared constant or something to avoid repeating it? I think you can probably also add it to the `Base`

python/utils/transactions_processor.py
```diff
@@ -152,3 +153,23 @@ def insert_to_db(self, parsed_objs, txn_version) -> None:
...
+
+
+@event.listens_for(Base.metadata, "after_drop")
+def drop_schemas(target, connection, **kw):
+    schemas = set()
+    for table in target.tables.values():
+        if table.schema is not None:
+            schemas.add(table.schema)
+    for schema in schemas:
+        connection.execute(DDL("DROP SCHEMA IF EXISTS %s" % schema))
```
geekflyer, line 175: I think this will drop the entire schema if you delete a single table right now, which is not what we want.

geekflyer (changes requested):

python/processors/nft_orderbooks/models/nft_marketplace_bid_models.py
bowenyang007: @rtso fair optimization.

python/utils/transactions_processor.py
rtso: This event listener listens for drop on `Base.metadata`, so it will only drop the schema if the whole DB is deleted with `MetaData.drop_all`

PR #44, opened 2023-07-10
create monitoring and health check port.

larry-aptos:
Locally tested, looks good.

image

image

image

Keep alive check: 
image

### DIFF
--- a/python/utils/config.py
+++ b/python/utils/config.py
@@ -13,6 +13,14 @@ class Config(BaseSettings):
     grpc_data_stream_endpoint: str
     grpc_data_stream_api_key: str
     db_connection_uri: str
+    # Used for k8s liveness and readiness probes
+    health_port: int
+    # Used for Prometheus metrics collection at /metrics
+    monitoring_port: int
+    # HTTP2 ping interval in seconds to detect if the connection is still alive
+    indexer_grpc_http2_ping_interval_in_secs: int
+    # HTTP2 ping timeout in seconds to detect if the connection is still alive
+    indexer_grpc_http2_ping_timeout_in_secs: int
     starting_version_default: Optional[int] = None
     starting_version_override: Optional[int] = None
     ending_version: Optional[int] = None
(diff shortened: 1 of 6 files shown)


python/utils/transactions_processor.py
```diff
@@ -33,6 +38,25 @@ def __init__(
...
+        # Start the Prometheus client; this is a non-blocking call.
+        start_http_server(self.config.monitoring_port)
+
+        # Start the health server.
+        def start_health_server() -> None:
+            class Handler(http.server.SimpleHTTPRequestHandler):
+                def do_GET(self):
+                    self.send_response(http.HTTPStatus.OK)
+                    self.end_headers()
+                    self.wfile.write(b"OK")
```
rtso, line 50: Can you add more comment explaining what this does? How often do we send this response?

rtso (commented):
How did you test this?

python/utils/transactions_processor.py
larry-aptos: done, default is every 10 seconds from kubelet.

rtso: Thanks! Is this related to the 
```
indexer_grpc_http2_ping_interval_in_secs: 60
indexer_grpc_http2_ping_timeout_in_secs: 10
``` 
config vars that exist in the Rust config? If so we should add these to Python too.

python/utils/config.py
```diff
@@ -13,6 +13,8 @@ class Config(BaseSettings):
     grpc_data_stream_endpoint: str
     grpc_data_stream_api_key: str
     db_connection_uri: str
+    health_port: int = 8080
```
rtso, line 16: Can you also update `config.yaml.example`?

python/utils/transactions_processor.py
larry-aptos: oh this is a different config; i can update it here as well.

rtso: What are those vars for?

larry-aptos: There are to make sure connection is alive from client side. Without these configs, clients trust the connections blindly, which can be a problem if the underlying network is not stable.

larry-aptos: done.

python/utils/config.py
larry-aptos: done

rtso (approved):

PR #47, opened 2023-07-11
reduce one port from python processor.

larry-aptos:
* health is under `:health_port/`
* metrics is under `:health_port/metrics`

## Test plan

locally tested and looks good overall.


rtso:
We should move the health port stuff outside of `TransactionsProcessor` since we're planning to spawn multiple threads that run `TransactionsProcessor`. I'll approve for now!

rtso (approved):

PR #50, opened 2023-07-14
add the rust processor.

larry-aptos:
* Add the rust processor.
* Buf now generates the code for rust/python and typescript.
* Create a dummy Dockerfile for rust processor for image release. 

* Locally tested and looks good.
image


rtso (commented):
Can you run the Rust and Python processors locally to make sure everything works?

python/scripts/rust_lint.sh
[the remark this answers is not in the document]
larry-aptos: done

proto/aptos/indexer/v1/raw_data.proto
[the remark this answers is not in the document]
larry-aptos: good call... another solution is proto in a dedicated proto and submodule into these repos.

.gitignore
```diff
@@ -1,14 +1,44 @@
...
+.*\#
+\#*\#
+
+# VSCode settings
+.vscode/
+
+# Cached python files
+__pycache__
+
+# Protocol buffer generated files
```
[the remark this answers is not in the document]
larry-aptos: Added.

proto/aptos/internal/fullnode/v1/fullnode_data.proto
rtso, line 1: Is this proto needed in this repo? (I thought the grpc processors don't get their data from fullnode)

proto/aptos/indexer/v1/raw_data.proto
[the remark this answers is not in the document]
rtso: There's pros and cons with all the options. One advantage with having the protos defined in separate repos is they're highly customizable so we can have different version. We just need to know to update them if one changes.

rust/server-framework/src/lib.rs
rtso, line 1: We can probably remove the ServerConfig now because it will only run processors (don't need to run cache worker, data service, etc)

rust/aptos-protos/src/pb/aptos.bigquery_schema.transaction.v1.rs
rtso, line 1: Ignore these files?

rust/processor/src/processors/mod.rs
rtso, line 1: Could we split up each processor into separate folder? This is the proposed folder hierarchy - https://docs.google.com/document/d/1hj_D7p5lqXnS7qJwnQc7yyfBnbJLA9TzO5TK6xH05PY/edit

proto/aptos/indexer/v1/raw_data.proto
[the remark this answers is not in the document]
larry-aptos: true; i think how to handle the proto is a topic that needs some discussion... let me kick off the discussion.

rust/aptos-protos/src/pb/aptos.bigquery_schema.transaction.v1.rs
larry-aptos: done.

rust/processor/src/processors/mod.rs
larry-aptos: i'll do it with a follow-up. this PR is too large now...

proto/aptos/internal/fullnode/v1/fullnode_data.proto
larry-aptos: yea, going to leave to proto handling discussion.

.github/workflows/build-images.yaml
```diff
@@ -25,10 +24,34 @@ jobs:
...
+          cargo install protoc-gen-prost
+          cargo install protoc-gen-prost-serde
+          cargo install protoc-gen-prost-crate
+          cargo install protoc-gen-tonic
+      - name: Generate the protos
+        uses: bufbuild/buf-setup-action@v1.24.0
+      - run: buf generate
+      - run: > 
+          pip install grpcio-tools &&
+          python3 -m grpc_tools.protoc --proto_path=./proto --python_out=python --pyi_out=python --grpc_python_out=python
```
rtso, line 49: Can this just call build_protos.sh?

rtso (approved):
Can you test the Rust and Python build images works too?

.github/workflows/build-images.yaml
larry-aptos: i spent a lot of time on this... and the tldr is, we can't do it easily if we don't use buf remote registry..

the missing part is `protoc-gen-python_grpc`, which doesn't have a standalone binary(and ideally cross-platform). the only option is, build from source.

but buf has `grpc_python` built-in plugin, which generates `async` code instead. so to make it work and if we want, we can go that route...

PR #56, opened 2023-07-20
[Rust] Redo ans indexing

rtso:
## Description 
1. Add new tables, current_ans_primary_name and ans_primary_name, for indexing primary names
2. Add new table, ans_lookup, for tracking name record changes (
3. Add is_deleted column to existing current_ans_lookup table 

Indexing strategy: https://www.notion.so/aptoslabs/ANS-v1-Indexing-ca2fc68f454b497595dbab93b14109a7?pvs=4#b3234acc65f54c35a88eef71cdfb0692
TL;DR
For each transaction: 
1. Loop through name record table write set changes
    a. Parse Write table item
    b. Parse Delete table item
5. Loop through primary name table write set changes
    a. Parse Write table item
         i. Do a lookup to update the primary name record the address previously points to
         ii. Create address's new primary name mapping
    b. Parse Delete table item
         i. Do a lookup to update the primary name record the address previously points to

## Testing 
Screenshot 2023-08-15 at 10 30 13 PM
Screenshot 2023-08-15 at 10 14 13 PM
Screenshot 2023-08-15 at 10 13 52 PM
Screenshot 2023-08-15 at 10 13 35 PM


rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -24,6 +25,12 @@ type Subdomain = String;
 // PK of current_ans_lookup, i.e. domain and subdomain name
 pub type CurrentAnsLookupPK = (Domain, Subdomain);
 
+pub const APTOS_PRIMARY_NAMES_TABLE_HANDLE: String =
+    String::from("0x1d5f57aa505a2fa463b7a46341913b65757e3177c46a5e483a29d953627bee62");
+
+pub const APTOS_NAMES_RECORDS_TABLE_HANDLE: String =
+    String::from("0x21a0fd41330f3a0a38173c7c0e4ac59cd51505f0594f64d3d637c12425c3c155");
+
```
bowenyang007, line 33: &string. look at some of the other files.

rust/processor/parser.yaml
```diff
@@ -1,7 +1,7 @@
...
 health_check_port: 8084
 server_config:
-    processor_name: default_processor
-    postgres_connection_string: postgresql://postgres:@localhost:5432/postgres_v2
-    indexer_grpc_data_service_address: 127.0.0.1:50051:50051
-    auth_token: AUTH_TOKEN
\ No newline at end of file
+  processor_name: default_processor
+  postgres_connection_string: postgresql://postgres:@localhost:5432/postgres_v2
+  indexer_grpc_data_service_address: 127.0.0.1:50051:50051
```
bowenyang007, line 6: this is a typo lol

bowenyang007 (changes requested):
I see the general approach you're trying to go with, but it's kind of weird tbh. I'd recommend doing it differently: 
1. use a single map to track the overall mapping
2. when you see a new registration, add to the mapping
3. when you see a primary key, upsert to mapping. 

You should only need a single map for this.

rust/processor/parser.yaml
rtso: this is just reformatting

bowenyang007: No I mean the 50051:50051. Remove one of them (line 6)

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +77,225 @@ impl OptionalString {
...
+        // Extracts from user transactions. Other transactions won't have any ANS changes
+        if let (
+            TxnData::User(_),
+            Some(ans_primary_names_table_handle),
+            Some(ans_name_records_table_handle),
+        ) = (
+            txn_data,
+            maybe_ans_primary_names_table_handle,
+            maybe_ans_name_records_table_handle,
```
bowenyang007, line 116: oh this is interesting I didnt' realize you could do that lol.

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +77,225 @@ impl OptionalString {
...
+                    if let Some(name_record) = maybe_name_record {
+                        current_ans_lookups
+                            .entry((name_record.domain.clone(), name_record.subdomain.clone()))
+                            .and_modify(|e| {
+                                e.registered_address = name_record.registered_address.clone();
+                                e.expiration_timestamp = name_record.expiration_timestamp;
+                                e.is_primary = name_record.is_primary || e.is_primary;
+                                e.is_deleted = name_record.is_deleted;
+                            })
+                            .or_insert(name_record);
```
bowenyang007, line 152: huh I didn't realize this either...where r u getting all these cool syntax LOL

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +77,225 @@ impl OptionalString {
...
+pub struct NameRecordV1 {
+    #[serde(deserialize_with = "deserialize_from_string")]
+    expiration_time_sec: BigDecimal,
+    #[serde(deserialize_with = "deserialize_from_string")]
+    property_version: BigDecimal,
+    target_address: OptionalString,
+}
+
 impl CurrentAnsLookup {
     pub fn from_transaction(
```
bowenyang007, line 96: Can you add some comments on what the overall flow is? It's a bit difficult understanding how these different handles play with each other.

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +77,225 @@ impl OptionalString {
...
+                } else {
+                    continue;
                 }
             }
         }
+
         current_ans_lookups
     }
+
+    pub fn from_write_table_item(
```
bowenyang007, line 163: Add comment.

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +77,225 @@ impl OptionalString {
...
+                expiration_timestamp,
+                token_name,
+                is_primary: false,
+                is_deleted: false,
+            })
+        }
+        Ok(maybe_name_record)
+    }
+
+    pub fn from_delete_table_item(
```
bowenyang007, line 260: add comment

rust/processor/src/models/token_models/ans_lookup.rs
rtso: HashMap has an [Entry API](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry) that allows you to do upserts like this

rust/processor/src/models/token_models/ans_lookup.rs
rtso: This is combining a bunch of `if lets` into one, I didn't know we could do that either.

rust/processor/src/processors/token_processor.rs
```diff
@@ -889,9 +901,14 @@ impl ProcessorTrait for TokenTransactionProcessor {
...
-                CurrentAnsLookup::from_transaction(txn, self.ans_contract_address.clone());
-            all_current_ans_lookups.extend(current_ans_lookups);
+            let current_ans_lookups = CurrentAnsLookup::from_transaction(
+                txn,
+                &mut address_to_primary_name_records,
+                self.ans_primary_names_table_handle.clone(),
+                self.ans_name_records_table_handle.clone(),
+                &mut conn,
+            );
+            all_current_ans_lookups.extend(current_ans_lookups.clone());
```
bowenyang007, line 911: I think that you still need to pass `all_current_ans_lookups` back to the function to handle the case that same batch may have data you need.

rust/processor/src/models/token_models/ans_lookup.rs
bowenyang007: If possible can you elaborate a bit more about how the parsing actually works. For example, something like 
We have 2 tables, one table the key is nameRecord, the value is token, the other table the key is address and value is nameRecord. 

In case 1: this happens
case2: this happens.

rust/processor/src/processors/token_processor.rs
rtso: The current batch tracks address -> name mappings in `address_to_primary_name_records`

rust/processor/src/models/token_models/ans_lookup.rs
bowenyang007: ping

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +73,396 @@ impl OptionalString {
...
-                        event.r#type.as_ref().unwrap().content.as_ref().unwrap()
-                    {
-                        (
-                            inner.address.to_string(),
-                            format!("{}::{}", inner.module, inner.name),
+                .expect("Transaction info doesn't exist!");
+
+            // Extract primary names and name records
+            for wsc in &transaction_info.changes {
+                if let Ok(name_records) = match wsc.change.as_ref().unwrap() {
```
bowenyang007, line 123: Hmm change shouldn't be a Result. Isn't it an optional and .unwrap should already expose the struct?

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +73,396 @@ impl OptionalString {
...
+                    )
+                });
+            let subdomain = primary_name_record
+                .subdomain_name
+                .get_string()
+                .unwrap_or_default();
+            let mut token_name = format!("{}.apt", &primary_name_record.domain_name);
+            if !subdomain.is_empty() {
+                token_name = format!("{}.{}", &subdomain, token_name);
+            }
```
bowenyang007, line 233: Let's have NameRecordKeyV1 implement these methods.

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +73,396 @@ impl OptionalString {
...
+                    token_name,
+                    is_primary: false,
+                    is_deleted: true,
+                },
+            );
+        }
+        Ok(current_ans_lookup)
+    }
+
+    fn get_primary_record_from_address(
```
bowenyang007, line 430: Add comment.

rust/processor/src/models/token_models/ans_lookup.rs
```diff
@@ -74,107 +73,396 @@ impl OptionalString {
...
+        }
+        Ok(None)
+    }
+
+    fn get_by_registered_address(
+        conn: &mut PgPoolConnection,
+        registered_address: &str,
+    ) -> anyhow::Result<Self> {
+        let mut res: Vec<Option<Self>> =
+            sql_query("SELECT * FROM current_ans_lookup WHERE registered_address = $1")
```
bowenyang007, line 461: Should check is_primary = true?

rust/processor/src/models/token_models/ans_lookup.rs
rtso: It's unwrapping the Result returned by `Self::from_write_table_item`

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/down.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE current_ans_lookup
+DROP COLUMN is_primary,
+DROP COLUMN is_deleted;
```
bowenyang007, line 3: drop column if exists

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/up.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE current_ans_lookup
+ADD COLUMN is_primary BOOLEAN NOT NULL DEFAULT FALSE,
+ADD COLUMN is_deleted BOOLEAN NOT NULL DEFAULT FALSE;
```
bowenyang007, line 3: add column if not exists

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/up.sql
```diff
@@ -0,0 +1,19 @@
...
+
+CREATE TABLE ans_lookup (
```
bowenyang007, line 5: create table if not exists

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/up.sql
```diff
@@ -0,0 +1,19 @@
...
+    transaction_version BIGINT NOT NULL,
+    domain VARCHAR(64) NOT NULL,
+    -- if subdomain is null set to empty string
+    subdomain VARCHAR(64) NOT NULL,
+    registered_address VARCHAR(66),
+    expiration_timestamp TIMESTAMP,
+    token_name VARCHAR(128) NOT NULL,
```
bowenyang007, line 12: 140. 64 + 64 + some dots and APT. anyway 140 will work.

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/up.sql
```diff
@@ -0,0 +1,19 @@
...
+    is_primary BOOLEAN,
+    is_deleted BOOLEAN NOT NULL DEFAULT FALSE,
+    transaction_timestamp TIMESTAMP NOT NULL,
+    inserted_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    -- Constraints 
+    PRIMARY KEY (transaction_version, domain, subdomain)
```
bowenyang007, line 18: Why not just use transaction_version and write_set_change_index to conform to the pattern? It'll also be easier to find.

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+pub const QUERY_RETRY_DELAY_MS: u64 = 500;
+
+type Domain = String;
+type Subdomain = String;
+// Hashmap for looking up previous name records by target address
+pub type AddressToPrimaryNameRecord = HashMap<String, CurrentAnsLookup>;
+// PK of current_ans_lookup, i.e. domain and subdomain name
+pub type CurrentAnsLookupPK = (Domain, Subdomain);
+// PK of ans_lookup, i.e. transaction version, domain, and subdomain
+pub type AnsLookupPK = (i64, Domain, Subdomain);
```
bowenyang007, line 38: You don't need this if you follow the pattern of transaction_version and wsc_index

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/up.sql
```diff
@@ -0,0 +1,19 @@
...
+    transaction_version BIGINT NOT NULL,
+    domain VARCHAR(64) NOT NULL,
+    -- if subdomain is null set to empty string
+    subdomain VARCHAR(64) NOT NULL,
+    registered_address VARCHAR(66),
+    expiration_timestamp TIMESTAMP,
+    token_name VARCHAR(128) NOT NULL,
+    is_primary BOOLEAN,
+    is_deleted BOOLEAN NOT NULL DEFAULT FALSE,
+    transaction_timestamp TIMESTAMP NOT NULL,
```
bowenyang007, line 15: to be consistent, remove this or add it to the other table?

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/up.sql
```diff
@@ -0,0 +1,19 @@
...
+CREATE TABLE ans_lookup (
+    transaction_version BIGINT NOT NULL,
+    domain VARCHAR(64) NOT NULL,
+    -- if subdomain is null set to empty string
+    subdomain VARCHAR(64) NOT NULL,
+    registered_address VARCHAR(66),
+    expiration_timestamp TIMESTAMP,
+    token_name VARCHAR(128) NOT NULL,
+    is_primary BOOLEAN,
+    is_deleted BOOLEAN NOT NULL DEFAULT FALSE,
```
bowenyang007, line 14: why default null?

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+    Transaction as TransactionPB, WriteTableItem,
+};
+use bigdecimal::BigDecimal;
+use diesel::{prelude::*, ExpressionMethods};
+use field_count::FieldCount;
+use serde::{Deserialize, Serialize};
+use std::{cmp, collections::HashMap};
+
+pub const QUERY_RETRIES: u32 = 5;
+pub const QUERY_RETRY_DELAY_MS: u64 = 500;
```
bowenyang007, line 29: don't set these. There is already something defined somewhere else.

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+    pub fn from_transaction(
+        transaction: &TransactionPB,
+        current_ans_lookups: &HashMap<CurrentAnsLookupPK, Self>,
+        address_to_primary_name_record: &AddressToPrimaryNameRecord,
+        maybe_ans_primary_names_table_handle: Option<String>,
+        maybe_ans_name_records_table_handle: Option<String>,
+        conn: &mut PgPoolConnection,
+    ) -> (
+        HashMap<CurrentAnsLookupPK, Self>,
+        HashMap<AnsLookupPK, AnsLookup>,
```
bowenyang007, line 117: vec should be fine. Look at comments above.

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+pub struct NameRecordV1 {
+    #[serde(deserialize_with = "deserialize_from_string")]
+    expiration_time_sec: BigDecimal,
+    #[serde(deserialize_with = "deserialize_from_string")]
+    property_version: BigDecimal,
+    target_address: OptionalString,
+}
+
+impl CurrentAnsLookup {
+    pub fn from_transaction(
```
bowenyang007, line 108: If you're going to make a new processor, why not just move this logic to the processor? It'll be better b/c when you add v2 you don't need to make multiple loops again in the transaction

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+            )
+        });
+        let value_type: &str = table_item_data.value_type.as_str();
+        let value = &table_item_data.value;
+        let primary_name_record: NameRecordKeyV1 =
+            serde_json::from_str(value).unwrap_or_else(|e| {
+                panic!(
+                    "version {} failed! failed to parse type {}, data {:?}. Error: {:?}",
+                    txn_version, value_type, value, e
+                )
```
bowenyang007, line 299: we have a better pattern for this: 
tracing::error!(transaction_version=txn_version..., "failed to parse NameRecordKeyV1")
panic("failed to parse NameRecordKeyV1")

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+                                    e.is_deleted = name_record.is_deleted;
+                                })
+                                .or_insert(name_record);
+                        }
+                    },
+                    Err(e) => {
+                        tracing::error!(
+                            "Failed to parse primary name reverse lookup table item. Error: {:?}",
+                            e
+                        );
```
bowenyang007, line 259: structured error. 

Not sure why you don't panic here but panic in that other function? I'd recommend not panicking anywhere.

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+                    transaction_version: txn_version,
+                    domain: previous_primary_record.domain,
+                    subdomain: previous_primary_record.subdomain,
+                    registered_address: previous_primary_record.registered_address,
+                    expiration_timestamp: previous_primary_record.expiration_timestamp,
+                    token_name: previous_primary_record.token_name,
+                    is_primary: false,
+                    is_deleted: previous_primary_record.is_deleted,
+                    transaction_timestamp: txn_timestamp,
+                },
```
bowenyang007, line 333: why are we using data from previous primary record? Shouldn't this be fetched from the current transaction for the new item?

bowenyang007: oh I see, the variable name is kinda weird. Can we you name to ans_lookups? the new_ prefix is misleading.

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+                primary_name_record.get_subdomain().clone(),
+            ),
+            Self {
+                domain: primary_name_record.get_domain().clone(),
+                subdomain: primary_name_record.get_subdomain().clone(),
+                registered_address: Some(String::from(key)),
+                last_transaction_version: txn_version,
+                // Default to zero
+                expiration_timestamp: chrono::NaiveDateTime::from_timestamp_opt(0, 0)
+                    .unwrap_or_default(),
```
bowenyang007, line 351: how is this possible? I think that we should just unwrap? 

Also I'm confused why you're not using the actual expiration timestamp from the table item.

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+                    token_name: previous_primary_record.token_name,
+                    is_primary: false,
+                    is_deleted: previous_primary_record.is_deleted,
+                    transaction_timestamp: txn_timestamp,
+                },
+            );
+        }
+
+        // If the name record doesn't exist yet, create it and insert it to current_ans_lookups.
+        // This is possible if the primary name record wsc happens before the name record wsc.
```
bowenyang007, line 338: How? Aren't you handling the name record in a separate loop?

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+        let mut new_current_ans_lookups = HashMap::new();
+        let mut new_ans_lookups = HashMap::new();
+
+        let table_item_data = delete_table_item.data.as_ref().unwrap();
+        let key_type = table_item_data.key_type.as_str();
+        let key = serde_json::from_str(&table_item_data.key).unwrap_or_else(|e| {
+            panic!(
+                "version {} failed! failed to parse type {}, data {:?}. Error: {:?}",
+                txn_version, key_type, table_item_data.key, e
+            )
```
bowenyang007, line 400: same comment as above.

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+            ),
+            AnsLookup {
+                transaction_version: txn_version,
+                domain: previous_name_record.domain,
+                subdomain: previous_name_record.subdomain,
+                registered_address: previous_name_record.registered_address,
+                expiration_timestamp: previous_name_record.expiration_timestamp,
+                token_name: previous_name_record.token_name,
+                is_primary: false,
+                is_deleted: previous_name_record.is_deleted,
```
bowenyang007, line 444: why is this not just true?

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+                ),
+                AnsLookup {
+                    transaction_version: txn_version,
+                    domain: previous_primary_record.domain,
+                    subdomain: previous_primary_record.subdomain,
+                    registered_address: previous_primary_record.registered_address,
+                    expiration_timestamp: previous_primary_record.expiration_timestamp,
+                    token_name: previous_primary_record.token_name,
+                    is_primary: false,
+                    is_deleted: previous_primary_record.is_deleted,
```
bowenyang007, line 331: isn't this just true?

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+                            address_to_primary_name_record,
+                            conn,
+                        )
+                    },
+                    _ => continue,
+                };
+
+                match res {
+                    Ok((current_ans_lookups, ans_lookups)) => {
+                        for (record_key, name_record) in current_ans_lookups {
```
bowenyang007, line 219: This flow is hard to understand. I think that you may be able to do this differently if you first do a full loop to get the full mapping similar to how we deal with object parsing and then do a full loop to get both NR and PN. I roughly understand what you're doing here but it's very confusing.

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,776 @@
...
+}
+
+impl NameRecordKeyV1 {
+    pub fn get_domain(&self) -> String {
+        self.domain_name.clone()
+    }
+
+    pub fn get_subdomain(&self) -> String {
+        self.subdomain_name.get_string().unwrap_or_default()
+    }
```
bowenyang007, line 746: if you're just returning the objects without any modification, why the wrappers?

bowenyang007 (changes requested):
This is pretty close. Should just need more polishing. LMK if you want some help on the last touches.

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/up.sql
rtso: Because an ANS creation can exist in 2 wsc indices (the name record and the primary key reverse lookup). We could also split them up into 2 rows.

rust/processor/migrations/2023-07-19-190707_add_ans_is_primary/up.sql
rtso: I'll remove this

rust/processor/src/models/ans_models/ans_lookup.rs
rtso: I think we should panic here though. If it's unable to parse, then the assumption the indexer makes of the underlying data model is wrong and we should fix it.

rust/processor/src/models/ans_models/ans_lookup.rs
rtso: We can default to null

rust/processor/src/models/ans_models/ans_lookup.rs
bowenyang007: So this requires all the data to be backfilled right? We had a sev about 3 weeks ago where we made this assumnption and a partner didn't backfill so their node broke.

rtso: Hm ok, if it can't parse, then we can just skip this txn. Also we panic in a lot of other places so we should remove those too.

rust/processor/src/models/ans_models/ans_lookup.rs
rtso: Oop this got left over from when we had a combined loop for both. Will remove.

rust/processor/src/models/ans_models/ans_lookup.rs
rtso: is_deleted represents whether the whole record got deleted. 

Consider the case where NR WTI and PN DTI exist in the same transaction. is_deleted should persist from the current record.

rust/processor/src/models/ans_models/ans_lookup.rs
rtso: No is_deleted = true means the whole record got deleted. We should only set is_primary = false.

rust/processor/src/models/ans_models/ans_lookup.rs
rtso: Subdomain can be null. It's an OptionalString so we need to unwrap it.

rust/processor/src/worker.rs
```diff
@@ -351,7 +359,8 @@ impl Worker {
             let mut transactions_batches = vec![];
             let mut last_fetched_version = batch_start_version as i64 - 1;
             for task_index in 0..concurrent_tasks {
-                let receive_status = match task_index {
+                let receive_status: Result<(u64, Vec<Transaction>), TryRecvError> = match task_index
```
bowenyang007, line 362: I don't think that this is intentional? Can we remove?

rust/processor/src/models/ans_models/ans_lookup.rs
```diff
@@ -0,0 +1,356 @@
...
+        txn_version: i64,
+        write_set_change_index: i64,
+    ) -> anyhow::Result<Option<(Self, AnsPrimaryName)>> {
+        let table_handle = standardize_address(&delete_table_item.handle.to_string());
+        if table_handle
+            != standardize_address(
+                ans_v1_primary_names_table_handle
+                    .clone()
+                    .unwrap_or_default()
+                    .as_str(),
```
bowenyang007, line 321: Did I add this? hmm if I did then prolly got lazy. 

this is weird syntax. You should limit the use of unwrap_or_default to unexpected edge cases. Better to use .map or if let Some.

Also .clone....as_str usually means that you're copying data that you don't have to.

bowenyang007 (approved):

rust/processor/src/worker.rs
rtso: oops

PR #67, opened 2023-07-27
Rename to aptos-indexer-protos

banool:
We should rename aptos-protos to aptos-indexer-protos now that we’re not managing it in aptos-core, just in case we end up with more protos some day that aren’t related to the indexer stack. It wouldn’t matter if we just used the crate ourselves but if we wanna publish it to crates.io we should rename, esp since we’ve already used the aptos-protos name for a crate from when it was in aptos-core.


larry-aptos (approved):

larry-aptos:
Looks good overall!!!

PR #75, opened 2023-07-31
[Rust] hardcode apt aggregator table instead of fetching every version

bowenyang007:
We're querying postgres every version which is causing tps to be 1k instead of prolly 8-9k. 

It's actually the same for all networks so we can hardcode instead of using a config. 
```
  apt_supply_table_handle: "0x1b854694ae746cdbd8d44186ca4929b2b337df21d1c74633be19b2710552fdca"
  apt_supply_table_key: "0x619dc29a0aac8fa146714058e8dd6d2d0f3bdf5f6331907bf91f3acd81e6935"
```

### DIFF
--- a/rust/processor/src/models/coin_models/coin_supply.rs
+++ b/rust/processor/src/models/coin_models/coin_supply.rs
@@ -15,6 +15,11 @@ use bigdecimal::BigDecimal;
 use field_count::FieldCount;
 use serde::{Deserialize, Serialize};
 
+const APTOS_COIN_SUPPLY_TABLE_HANDLE: &str =
+    "0x1b854694ae746cdbd8d44186ca4929b2b337df21d1c74633be19b2710552fdca";
+const APTOS_COIN_SUPPLY_TABLE_KEY: &str =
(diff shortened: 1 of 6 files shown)


larry-aptos (approved):

rust/processor/src/models/coin_models/coin_supply.rs
```diff
@@ -5,14 +5,21 @@
...
+    processors::coin_processor::APTOS_COIN_TYPE_STR, schema::coin_supply, utils::util::hash_str,
+};
 use anyhow::Context;
 use aptos_indexer_protos::transaction::v1::WriteTableItem;
 use bigdecimal::BigDecimal;
 use field_count::FieldCount;
 use serde::{Deserialize, Serialize};
 
+const APTOS_COIN_SUPPLY_TABLE_HANDLE: &str =
+    "0x1b854694ae746cdbd8d44186ca4929b2b337df21d1c74633be19b2710552fdca";
```
rtso, line 19: Aren't these table handles going to be different based on the environment (mainnet vs devnet)?

PR #77, opened 2023-08-01
[Rust][TokenV2] Index fungible assets

bowenyang007:
## Context
We need to index fungible assets which is basically coin v2. The data mode is in https://dbdiagram.io/d/64b8c1d202bd1c4a5e61e486. The goal is to create unified tables that have both v1 and v2 data in a way that is not too disruptive to the existing ecosystem. 

## What does this PR do
I included a few things in this PR to move fast (sorry for the mess!)
### Fungible Assets
* We added 4 tables: 
  * fungible_asset_activities --> maps to coin_activities
  * fungible_asset_metadata --> maps to coin_infos
  * fungible_asset_balances --> maps to coin_balances
  * current_fungible_asset_balances --> maps to current_coin_balances
* Each model has get_v1_from_... which is the same as the old one
* V2 specific logic
  * fungible_asset_metadata
    * Indexes `0x1::fungible_asset::Metadata` resource. If it happens to be a token though, skip
  * fungible_asset_activities
    * Indexes `0x1::fungible_asset::DepositEvent`, `0x1::fungible_asset::WithdrawEvent`, `0x1::fungible_asset::FrozenEvent`. Skip if it's a token
  * fungible_asset_balances / current_...
    * Indexes `fungible_asset::FungibleStore`. skip if it's a token. 
    * Check for whether the store is primary. The way to do it is by checking if the hash of (owner address + metadata address + 0xfc) = fungible store address
  * Because of object and how everything is split into several parts, we do 4 loops to get all the data (first loop over write sets to get objects, loop over write sets again to get all the metadata, then loop events and final loop over write sets to index the actual thing)

### Changes to old coin
* Added fee payer support (extra column in coin_activities, also fungible_assets activities call fee_payer_address that tracks whether gas was paid by someone else).
* Removed coin_type_hash and applied a limit of 1k instead to coin_type. If the coin_type is above 1k, we don't index it.
* Created storage_id which is a hash of owner address::coin type to conform to how v2 works.

### NFT / FT
* Typo in token ownerships where storage id was mistakenly set to token data id instead of the container address for fungible tokens
* Fungible assets were included in token activities and token ownerships. Fixed that.

### Miscellaneous
* Made a better ownership by collection view query

## Backfill
This PR needs to backfill token processor from:
testnet: 553151941
mainnet: 199271122

coin_processor from genesis

## Test
Ensure that metadata works and fungible token metadata does not appear (e.g. squeezy)
image
Ensure that fungible tokens don't appear in balances
image
Ensure that FA activities show up correctly
image
Ensure that balances work as intended, including is_primary and is_frozen
image
Ensure that fee payer shows up correctly
image
Ensure that fungible assets don't show up in token tables
**Before**
image
After is blank. Same for token activities v2


rust/processor/src/schema.rs
```diff
@@ -48,6 +48,8 @@ diesel::table! {
         transaction_timestamp -> Timestamp,
         inserted_at -> Timestamp,
         event_index -> Nullable<Int8>,
+        #[max_length = 66]
+        gas_fee_payer_address -> Nullable<Varchar>,
```
movekevin, line 52: May be this can be in a separate commit to keep things surgical? :)

rust/processor/src/schema.rs
```diff
@@ -268,6 +270,25 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    current_fungible_asset_balances (storage_id) {
+        #[max_length = 66]
+        storage_id -> Varchar,
```
movekevin, line 276: This is meant to be of the format "0x1::aptos_coin::AptosCoin" for Coins and "0x123456" for FungibleAssets right? I'd prefer a more descriptive name such as asset_id, which is also more consistent with token_data_id or collection_id for token v2 tables

bowenyang007: It's not meant to be asset_id. For v1 this is a hash of owner address + asset type.

rust/processor/src/schema.rs
bowenyang007: Yes but we need to deploy and backfill so 2 PRs add a lot more overhead. We can work around this but if it's just optics I'd like to move fast here.

rust/processor/src/schema.rs
```diff
@@ -268,6 +270,25 @@ diesel::table! {
...
+        #[max_length = 66]
+        owner_address -> Varchar,
+        #[max_length = 1000]
+        asset_type -> Varchar,
```
movekevin, line 280: This is meant to be of the format "0x1::aptos_coin::AptosCoin" for Coins and "0x123456" for FungibleAssets right?

rust/processor/src/schema.rs
```diff
@@ -268,6 +270,25 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    current_fungible_asset_balances (storage_id) {
```
movekevin, line 274: One new unique attribute in fungible asset is whether the entire store can be transferred or not. May be we can add a quick attribute for this (is_store_transferrable)?

rust/processor/src/schema.rs
movekevin, line 276: Is this auto-generated or does it have anything to do with the account address in case of Coin vs the FungibleStore object in case of FungibleAsset?

rust/processor/src/schema.rs
bowenyang007: Wait really? This is a feature now? Interesting, can you point me to the PR lol.

rust/processor/src/schema.rs
bowenyang007: Yes.

rust/processor/src/schema.rs
bowenyang007: For FA: it's the MetadataStore address
For V1: it's the hash of owner and coin type (e.g. 0xbowen::0x1::aptos_coin::AptosCoin)

rust/processor/src/schema.rs
```diff
@@ -518,6 +539,80 @@ diesel::table! {
...
 }
 
+diesel::table! {
+    fungible_asset_activities (transaction_version, event_index) {
+        transaction_version -> Int8,
+        event_index -> Int8,
+        #[max_length = 66]
+        owner_address -> Varchar,
+        #[max_length = 66]
+        storage_id -> Varchar,
```
movekevin, line 549: What does this field store?

rust/processor/src/schema.rs
```diff
@@ -518,6 +539,80 @@ diesel::table! {
...
+        #[max_length = 1000]
+        asset_type -> Varchar,
+        is_frozen -> Nullable<Bool>,
```
movekevin, line 552: Does the activity cover freezing/unfreezing a fungible store or coin store as well?

rust/processor/src/schema.rs
```diff
@@ -518,6 +539,80 @@ diesel::table! {
...
+        amount -> Nullable<Numeric>,
+        #[sql_name = "type"]
+        type_ -> Varchar,
+        is_gas_fee -> Bool,
+        #[max_length = 66]
+        gas_fee_payer_address -> Nullable<Varchar>,
+        is_transaction_success -> Bool,
+        #[max_length = 1000]
+        entry_function_id_str -> Nullable<Varchar>,
```
movekevin, line 561: Is this going to empty if this activity is from a script tx? Also, in the case of multisig tx execution, what is this set to?

rust/processor/src/schema.rs
```diff
@@ -518,6 +539,80 @@ diesel::table! {
...
+        block_height -> Int8,
+        #[max_length = 10]
+        token_standard -> Varchar,
```
movekevin, line 564: Is it better to use enums here? "v1" vs "v2" or "coin" vs "fungible_asset" seem like arbitrary strings that we don't have documentation for. cc @gregnazario as well

rust/processor/src/schema.rs
bowenyang007: Same thing as storage_id in other tables.

rust/processor/src/schema.rs
```diff
@@ -518,6 +539,80 @@ diesel::table! {
...
+        asset_type -> Varchar,
+        #[max_length = 66]
+        creator_address -> Varchar,
+        #[max_length = 32]
+        name -> Varchar,
+        #[max_length = 10]
+        symbol -> Varchar,
+        decimals -> Int4,
+        #[max_length = 512]
+        icon_uri -> Nullable<Varchar>,
```
movekevin, line 602: Technically not optional but can be empty strings. What's the policy here on using Nullable vs just leaving it empty?

rust/processor/src/schema.rs
```diff
@@ -518,6 +539,80 @@ diesel::table! {
...
+        amount -> Nullable<Numeric>,
+        #[sql_name = "type"]
+        type_ -> Varchar,
+        is_gas_fee -> Bool,
+        #[max_length = 66]
+        gas_fee_payer_address -> Nullable<Varchar>,
+        is_transaction_success -> Bool,
+        #[max_length = 1000]
+        entry_function_id_str -> Nullable<Varchar>,
+        block_height -> Int8,
```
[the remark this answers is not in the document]
bowenyang007: block height lol. txn version is transaction_version

rust/processor/src/schema.rs
movekevin: 2 commits in the same PR?

rust/processor/src/schema.rs
gregnazario: I think it's fine to have `Varchar` so it's able to be extended if it's not supported on the schema level, as long as there is a matching enum.

bowenyang007: It's just v1 and v2 for now. We can always change this or do something else. We don't want enums in postgres btw it's hella complicated.

rust/processor/src/schema.rs
```diff
@@ -518,6 +539,80 @@ diesel::table! {
...
+        #[max_length = 1000]
+        asset_type -> Varchar,
+        is_frozen -> Nullable<Bool>,
+        amount -> Nullable<Numeric>,
+        #[sql_name = "type"]
+        type_ -> Varchar,
+        is_gas_fee -> Bool,
+        #[max_length = 66]
+        gas_fee_payer_address -> Nullable<Varchar>,
+        is_transaction_success -> Bool,
```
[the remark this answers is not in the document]
bowenyang007: This is for gas lol.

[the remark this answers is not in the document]
bowenyang007: We have a gas event in every user transaction so if it's failed then we just log this as failed. Other events this is set to true by default.

rust/processor/src/schema.rs
movekevin: It's always been the case. The fungible store is transferrable if it's non-primary (only primary store objects are frozen)

rust/processor/src/schema.rs
movekevin: What if the owner changes?

rust/processor/src/schema.rs
bowenyang007: yea open to suggestion on how these enums can be better. Maybe this should be handled in the graphql side cc @jjleng

rust/processor/src/schema.rs
movekevin: For FA: that's not necessarily true if it's a hash of owner addr + type like you mentioned in an earlier comment. That's only true for primary fungible stores. Secondary store addresses are arbitrary

rust/processor/src/schema.rs
movekevin: Is there a way to comment on the schema here? If not, we need documentation of these schemas somewhere for external consumption :)

rust/processor/src/schema.rs
```diff
@@ -518,6 +539,80 @@ diesel::table! {
...
+        amount -> Numeric,
+        transaction_timestamp -> Timestamp,
+        #[max_length = 10]
+        token_standard -> Varchar,
+        inserted_at -> Timestamp,
+    }
+}
+
+diesel::table! {
+    fungible_asset_metadata (asset_type) {
```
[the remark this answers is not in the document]
bowenyang007: Hmm I realized that but idk what the query pattern will be. I plan to leave it to a follow up PR.

rust/processor/src/schema.rs
movekevin: CoinStore has freezing. You can freeze CoinStores (but not other structs that contain Coin)

rust/processor/src/schema.rs
movekevin: Payload is not going to be an entry function payload (it's a multisig execution payload that wraps around entry or script payload)

rust/processor/src/schema.rs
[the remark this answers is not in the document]
movekevin: Oh I see. Does the wallet or explorer use this?

rust/processor/src/schema.rs
jjleng: using Postgres enum is a double edge sword. Upgrading the enum would require DB schema upgrades, which is painful. I prefer to keep it as a string.

movekevin: How about Int4? And then we can declare a matching enum somewhere

rust/processor/src/schema.rs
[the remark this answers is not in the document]
movekevin: Max supply is very useful for explorers to display to the users that a token has bounded supply (which much of crypto cares about). This was a feature that people really wanted in Coin that we missed.

rust/processor/src/models/default_models/signatures.rs
```diff
@@ -91,6 +91,22 @@ impl Signature {
...
+    ) -> Option<String> {
+        match t.signature.as_ref().unwrap_or_else(|| {
+            tracing::error!(
+                transaction_version = transaction_version,
+                "Transaction signature is missing"
+            );
+            panic!("Transaction signature is missing");
+        }) {
+            SignatureEnum::FeePayer(sig) => Some(standardize_address(&sig.fee_payer_address)),
+            _ => None,
```
movekevin, line 106: What does this return for transactions where a separate gas payer is not set?

bowenyang007: It returns None.

rust/processor/src/schema.rs
bowenyang007: Interesting. Idk if it'll reflect in the API tho? We can explore a bit more here.

rust/processor/src/schema.rs
bowenyang007: Hmm. Do you have an example of this? We can check how it shows up currently.

rust/processor/src/schema.rs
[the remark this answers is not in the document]
bowenyang007: Wallet should yea.

rust/processor/src/models/default_models/signatures.rs
movekevin: Shouldn't it return the sender? Or does it default to the sender before persisting somewhere else

rust/processor/src/schema.rs
bowenyang007: Hmm. I guess there is no event. Good point for balances though.

rust/processor/src/schema.rs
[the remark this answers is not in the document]
movekevin: I see

rust/processor/src/schema.rs
bowenyang007: Yea but it's the same pattern.

bowenyang007: Basically in v1 there is only primary stores.

rust/processor/src/schema.rs
bowenyang007: Oh then that's fine right? !is_primary = is_store_transferrable

rust/processor/src/schema.rs
movekevin: I don't have a tx that changes some coin balances but here's a multisig tx: https://explorer.aptoslabs.com/txn/207817466/payload?network=mainnet

rust/processor/src/schema.rs
movekevin: is_store_transferrable for non-primary depends on the actual object itself being transferrable and might always be true

rust/processor/src/schema.rs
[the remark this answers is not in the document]
movekevin: Okay sounds good

rust/processor/src/schema.rs
bowenyang007: So this is the ungated_transfer thign? Tbh I don't think that it's useful from a indexer standpoint. Move can get this field in some ways, and downstream shouldn't care about it.

rust/processor/src/schema.rs
bowenyang007: Exactly. We just update the owner address. If we had owner_address and asset type as PKs, I'd have to do a lookup for the previous owner every time and it's not pretty.

bowenyang007: Or maybe I can look at the transfer event? Anyway this should work too and is arguably cleaner.

rust/processor/src/schema.rs
bowenyang007: Yea we need to document the schema externally for sure. I plan to write docs. Idk if we can add comments here and have them show up in Hasura. Anyway schema documentation is a broader conversation we can have w/ @jjleng.

rust/processor/src/schema.rs
bowenyang007: Got it, transaction_payload shows up as null so we will show null as well b/c we get it from the API.

rust/processor/src/schema.rs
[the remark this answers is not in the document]
bowenyang007: Fair point, I'll add it as a fast follow. It won't be much backfill either b/c it's only a v2 thing.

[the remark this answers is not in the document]
movekevin: Sounds good

rust/processor/src/schema.rs
movekevin: Okay. That's fine for now

rust/processor/src/schema.rs
movekevin: It might be applicable here for activities though because I transfer ownership of a store of 1000 SomeAsset to you, it's the same as giving you 1000 SomeAsset?

rust/processor/src/schema.rs
movekevin: So storage_id is not a persistent id? i.e. it can change?

rust/processor/src/schema.rs
bowenyang007: Oh. I thought that we didn't allow this unless for "burning". We specifically talked about how the right way to transfer asset is a withdraw + deposit. Transferring the entire wallet is not ideal. Is this...not true anymore? It might mess things up for downstream, e.g. wallet cc @kent-white 

I Btw wallet did decide to not care about non primary wallet.

movekevin: Secondary stores can be created by anyone and still transferrable. I think for "burning", you're referring to a slightly different case of soul bound token v2 and soul bound fungible store (e.g. non-transferrable like primary stores). We decided last time to only do this for token v2 for now as it's messy to allow generate "burning" of non-transferrable objects, especially when it comes to fungible assets.

bowenyang007: Wait what? No we said that we would also "burn" fungible assets.

bowenyang007: Just transfer the primary store to 0xgarbage. I'm not crazy right that this is what we said last time right @lightmark @davidiw @kent-white @CapCap

movekevin: Nope we never talked about fungible assets. Transferring the primary store to 0xsomething is bad as it means youll never be able to receive that tokens (eg you realize later is not scammy). It also breaks the accounting in any defi modules that rely on locking down transfers so they can update their accounting if transfers do happen. Transferring primary store breaks everything there

rust/processor/src/schema.rs
bowenyang007: Oh for V1 it cannot change. For v2 it's not generated.

larry-aptos (approved):

PR #78, opened 2023-08-01
Support backfill and fix multithreaded inserts

rtso:
* Multithreaded DB writes means that inserting a row  may overwrite newer transaction data already inserted by a different thread. 
* This adds the fix by doing on_conflict_do_update and comparing the last_transaction_version. 
* Also fixes the multithreaded postgres DB lock by sorting by primary key when inserting.


bowenyang007 (approved):
This code doesn't seem dry but to unblock I guess it's fine for now.

PR #80, opened 2023-08-04
add checking label condition to build image workflow

abbyxion:


bowenyang007:
Is it more complicated to specify rust / python builds? I suspect that most PRs will need some tag, but only in 1 language. 

If it's a lot of work then NVM.

.github/workflows/build-images.yaml
```diff
@@ -27,6 +29,7 @@ jobs:
       matrix:
         example: [python, rust]
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'release image')
```
rustielin, line 32: can we run this on our self-hosted runners?

bowenyang007:
Does it work?

abbyxion:
> Does it work?

we have [cache](https://console.cloud.google.com/artifacts/docker/aptos-registry/us/docker?project=aptos-registry) for processor images, so it should not take too long to build the image. Instead of adding labels to the workflow, I'll switch to the self-hosted docker runner to expedite the build

PR #81, opened 2023-08-06
[Rust] Reconnect max once every minute (configurable in code)

bowenyang007:
The GRPC stream isn't expected to persist forever and it's annoying that things disconnect every 3-4 hours. This adds reconnect feature to the rust processor. We allow max 1 reconnect per minute because anything more would likely indicate that the grpc server itself is broken. 

Note, this will help with the backfill problem but does not fully solve it. (i.e. during backfill if we restart the processor we'll restart from the manually specified starting_version all over again). 

## Testing
Ran the processor and tested a few cases manually: 
* Kill the grpc data service - observed correctly that reconnection happened
* Kill the grpc data service again - observed correctly that the second time didn't reconnect
* Set starting version and ending version as both 0 and observed that reconnection didn't affect this behavior. When we get to the end version the processor ended as expected. 

No reconnects for 10 hours! Yay!
image

### DIFF
--- a/rust/processor/src/worker.rs
+++ b/rust/processor/src/worker.rs
@@ -24,7 +24,7 @@ use crate::{
 };
 use anyhow::Context;
 use aptos_indexer_protos::{
-    indexer::v1::{raw_data_client::RawDataClient, GetTransactionsRequest},
+    indexer::v1::{raw_data_client::RawDataClient, GetTransactionsRequest, TransactionsResponse},
     transaction::v1::Transaction,
 };
 use aptos_moving_average::MovingAverage;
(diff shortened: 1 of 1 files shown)


jjleng (commented):
1. I don't know how long operational down time will be, but have we considered exponential backoff? Would it be faster to recover from transient failures?
2. I don't see ending_version being enforced. Does the worker actually stop fetching at ending_version?

bowenyang007:
1. So the design principle is that we still want to have infra failures in order to not mask issues with upstream. This is meant to handle the normal case where from time to time the upstream grpc would reset. Agreed that exponential backoff could basically help us never restart the server, but I don't think that it's actually going to help us. 
2. It's enforced in the grpc connection, but yes we should add a check in there. I'll add this later.

just-in-chang (approved):

PR #89, opened 2023-08-13
[config] Adding generated python and typescript files by removing them from .gitignore

xbtmatt:
Also added __init__.py to `aptos` and `utils` folders


rtso (approved):
thanks!!

PR #93, opened 2023-08-17
wait the response channel to be empty before closure

larry-aptos:
* wait the response channel to be empty before closure
* also fix some linter.

## testing plan

tested locally multiple times with finite stream; problem no longer occurs.

image

image

### Logs
image


rust/processor/src/worker.rs
```diff
@@ -238,9 +238,27 @@ impl Worker {
...
+            loop {
+                let channel_capacity = stream_closure_receiver.capacity();
+                info!(
+                    processor_name = processor_name,
+                    channel_size = BUFFER_SIZE - channel_capacity,
+                    "[Parser] Waiting for channel to be empty"
+                );
+                if channel_capacity == BUFFER_SIZE {
+                    break;
+                }
```
bowenyang007, line 256: Hmm if we're at the head this loop will stop right?

bowenyang007: Why don't we kick off this loop in the stream thread when we're at the end? (end_version > ending_version)

larry-aptos: done.

just-in-chang (approved):

PR #94, opened 2023-08-21
Partial voting review

bowenyang007:
Comments about the changes: 
* primary key already adds an index
  * in this case it's `PRIMARY KEY (delegation_pool_address, delegator_address)`, which adds an index already on `delegation_poo_address` so we don't need to add it again (basically if pk or index is on `(a, b)`, it creates an index for `a`, then `(a,b)`, you do need to add one for `b` separately if you plan to only query `b`)
* upgrade diesel so that diesel migrations generates the proper schema.rs
* Shouldn't limit to user transactions because some changes might be coming from block metadata changes
* Adopt the PK from @rtso's ANS PR
* Persist mappings across transactions (whenever a look up is needed, we need to also persist mapping thru the full batch)

Tips
* Unindent by assigning variables
* Make sure that models encapsulates as much as possible, leaving the main flow clean
* Minimize number of mapping and logic in general
* Reuse code, e.g. get_active_share_from_write_table_item
* Add doc blocks to functions

Tested w/ testnet
image

### DIFF
--- a/rust/processor/src/models/stake_models/current_delegated_voter.rs
+++ b/rust/processor/src/models/stake_models/current_delegated_voter.rs
@@ -5,22 +5,15 @@
 #![allow(clippy::extra_unused_lifetimes)]
 
 use super::{
-    delegator_balances::ShareToStakingPoolMapping,
-    delegator_pools::DelegatorPool,
-    stake_utils::{DelgationVoteGovernanceRecordsResource, VoteDelegationTableItem},
+    delegator_balances::{CurrentDelegatorBalance, ShareToStakingPoolMapping},
+    stake_utils::VoteDelegationTableItem,
 };
 use crate::{
     models::token_models::collection_datas::{QUERY_RETRIES, QUERY_RETRY_DELAY_MS},
     schema::current_delegated_voter,
-    utils::{
-        database::PgPoolConnection,
-        util::{parse_timestamp, standardize_address},
-    },
+    utils::{database::PgPoolConnection, util::standardize_address},
 };
-use aptos_indexer_protos::transaction::v1::{
-    transaction::TxnData, write_set_change::Change as WriteSetChange, Transaction as TransactionPB,
-    WriteResource, WriteTableItem,
-};
+use aptos_indexer_protos::transaction::v1::WriteTableItem;
 use diesel::{prelude::*, ExpressionMethods};
 use field_count::FieldCount;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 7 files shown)


jillxuu (approved):

bowenyang007:
wait it's not done lmao.

rust/processor/src/processors/stake_processor.rs
```diff
@@ -379,22 +388,92 @@ impl ProcessorTrait for StakeTransactionProcessor {
...
+                    if let Some(DelegationVoteGovernanceRecordsResource::GovernanceRecords(inner)) =
+                        DelegationVoteGovernanceRecordsResource::from_write_resource(
+                            write_resource,
+                            txn_version,
+                        )?
+                    {
+                        let delegation_pool_address =
+                            standardize_address(&write_resource.address.to_string());
+                        let vote_delegation_handle =
+                            inner.vote_delegation.buckets.inner.get_handle();
```
rtso, line 416: Do we need to standardize_address here?

rust/processor/src/processors/stake_processor.rs
```diff
@@ -366,7 +371,11 @@ impl ProcessorTrait for StakeTransactionProcessor {
         let mut all_delegator_pools: DelegatorPoolMap = HashMap::new();
         let mut all_delegator_pool_balances = vec![];
         let mut all_current_delegator_pool_balances = HashMap::new();
-        let mut all_current_delegated_voter: CurrentDelegatedVoterMap = HashMap::new();
+
+        let mut active_pool_to_staking_pool = HashMap::new();
```
rtso, line 375: Can you add a comment about what this will be used for in subsequent parsing?

rust/processor/src/processors/stake_processor.rs
```diff
@@ -366,7 +371,11 @@ impl ProcessorTrait for StakeTransactionProcessor {
...
+        // structs needed to get delegated voters
+        let mut all_current_delegated_voter = HashMap::new();
+        let mut all_vote_delegation_handle_to_pool_address = HashMap::new();
```
rtso, line 378: Can you add a comment about what this will be used for in subsequent parsing?

rust/processor/src/models/stake_models/current_delegated_voter.rs
```diff
@@ -359,12 +216,27 @@ impl CurrentDelegatedVoterQuery {
...
     ) -> diesel::QueryResult<Self> {
         current_delegated_voter::table
-            .filter(current_delegated_voter::delegation_pool_address.eq(delegation_pool_address))
             .filter(current_delegated_voter::delegator_address.eq(delegator_address))
+            .filter(current_delegated_voter::delegation_pool_address.eq(delegation_pool_address))
             .first::<Self>(conn)
     }
 }
+
+impl Ord for CurrentDelegatedVoter {
```
rtso, line 229: Move the boilerplate stuff to the beginning of the file

rtso (approved):
Nice, lgtm!

rust/processor/src/processors/stake_processor.rs
bowenyang007: It's encapsulated.

PR #96, opened 2023-08-22
[Python] GRPC reconnection

rtso:
Add reconnection logic to Python processors because the GRPC stream isn't expected to persist forever.

## Testing 
Throw `grpc.RpcError` and see that the processor reconnects.
Transaction batches don't have any gaps.

### DIFF
--- a/python/utils/worker.py
+++ b/python/utils/worker.py
@@ -2,7 +2,6 @@ import argparse
 import grpc
 import json
 
-from utils.models.general_models import NextVersionToProcess
 from aptos.indexer.v1 import raw_data_pb2, raw_data_pb2_grpc
 from aptos.transaction.v1 import transaction_pb2
 from utils.config import Config
@@ -11,7 +10,7 @@ from utils.session import Session
 from utils.metrics import PROCESSED_TRANSACTIONS_COUNTER
 from sqlalchemy import DDL, create_engine
 from sqlalchemy import event
-from typing import List
+from typing import Iterator, List
 from prometheus_client.twisted import MetricsResource
 from twisted.web.server import Site
 from twisted.web.resource import Resource
@@ -27,6 +26,7 @@ from processors.nft_orderbooks.nft_marketplace_processor import NFTMarketplacePr
 from processors.nft_marketplace_v2.processor import NFTMarketplaceV2Processor
 
 INDEXER_GRPC_BLOB_STORAGE_SIZE = 1000
+INDEXER_GRPC_MIN_SEC_BETWEEN_GRPC_RECONNECTS = 60
 
 
 class IndexerProcessorServer:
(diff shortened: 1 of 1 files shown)


python/utils/worker.py
```diff
@@ -102,185 +104,214 @@ def run(self):
...
+        # Initialize the batch information for tracking
+        batch_start_version = starting_version
+        batch_end_version = None
+        prev_processed_start_version = None
+        prev_processed_end_version = None
+
+        print(
+            json.dumps(
+                {
+                    "message": "[Parser] Connecting and making requests to GRPC endpoint",
```
bowenyang007, line 116: can you include the processor name in these logs?

python/utils/worker.py
```diff
@@ -92,195 +92,258 @@ def run(self):
...
 
-                    # If ending version is in the current batch, truncate the transactions in this batcch
-                    if ending_version != None and batch_end_version >= ending_version:
-                        batch_end_version = ending_version
-                        transactions = transactions[
-                            : batch_end_version - batch_start_version + 1
-                        ]
-                        current_batch_size = len(transactions)
+                # If it is a partial batch, then skip polling and head to process it first.
+                if current_batch_size < INDEXER_GRPC_BLOB_STORAGE_SIZE:
```
bowenyang007, line 233: I don't think that we need to add special handling to partial batch anymore due to server side chunking.

bowenyang007: Or actually, can we switch to a similar model to rust where we pull in a separate thread?

python/utils/worker.py
```diff
@@ -92,195 +92,258 @@ def run(self):
...
-        options = [
-            ("grpc.max_receive_message_length", -1),
-            (
-                "grpc.keepalive_time_ms",
-                self.config.indexer_grpc_http2_ping_interval_in_secs * 1000,
+        # Initialize the batch information for tracking
+        batch_start_version = starting_version
+        batch_end_version = None
+        prev_processed_start_version = None
+        prev_processed_end_version = None
```
bowenyang007, line 125: why do you need these?

bowenyang007: OK I kinda get prev_processed_end_version but not sure why we need the prev start version.

bowenyang007: Although...not sure why batch end version isn't enough. Isn't the current logic pull based on a batch_start_version, then process? You could just update batch_end_version at the end of the process and that's effectively your prev batch end version. 

If you switch to a separate thread handling the pulling this might change ofc.

python/utils/worker.py
```diff
@@ -92,195 +92,258 @@ def run(self):
...
-                ),
-                flush=True,
-            )
+        # Add reconnection stuff here
+        response_stream = self.get_transaction_response_stream(starting_version)
+        print(
+            json.dumps(
+                {
+                    "processor_name": self.processor.name(),
+                    "message": f"[Parser] Successfully connected to GRPC endpoint: {self.config.grpc_data_stream_endpoint}",
```
bowenyang007, line 143: The endpoint can also be also part of the structured data?

python/utils/worker.py
```diff
@@ -92,195 +92,258 @@ def run(self):
...
+                        )
+                        if (
+                            sec_since_last_reconnection
+                            < INDEXER_GRPC_MIN_SEC_BETWEEN_GRPC_RECONNECTS
+                        ):
+                            print(
+                                json.dumps(
+                                    {
+                                        "processor_name": self.processor.name(),
+                                        "message": "[Parser] Recently reconnected to GRPC stream. Will not retry",
```
bowenyang007, line 175: Let's standardize these logs? In rust it's `[Parser] Recently reconnected. Will not retry`

I know that it's not easy keeping these always in sync, maybe we could look into some type of constants across languages. It's not great having different logging messages for different languages as it makes debugging down the line much more difficult.

PR #99, opened 2023-08-22
Clean up ANS

rtso:
Update non-`current_` ANS tables to use `do_nothing` on conflict 

## Testing 
Run the processor in testnet locally 
Screenshot 2023-08-22 at 4 28 46 PM


jillxuu (approved):

rust/processor/migrations/2023-08-22-232603_add_ans_view/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE OR REPLACE VIEW current_ans_lookup_v2 AS
```
bowenyang007, line 1: can you rename this to current_aptos_names or something more generic? Technically it's not v2.

PR #101, opened 2023-08-23
[hasura] update hasura api for ans and partial voting

bowenyang007:


rtso (approved):
this is exactly what i did

PR #104, opened 2023-08-25
[Python] Update processors to use API gateway

rtso:
Also update the ambassador token processor to the new way. For deployment, need to change the config.yaml's. 

## Testing
`poetry run python -m processors.main -c config.yaml`

### DIFF
--- a/python/utils/worker.py
+++ b/python/utils/worker.py
@@ -2,7 +2,6 @@ import argparse
 import grpc
 import json
 
-from utils.models.general_models import NextVersionToProcess
 from aptos.indexer.v1 import raw_data_pb2, raw_data_pb2_grpc
 from aptos.transaction.v1 import transaction_pb2
 from utils.config import Config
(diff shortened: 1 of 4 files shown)


bowenyang007 (approved):

python/utils/worker.py
```diff
@@ -122,14 +124,30 @@ def run(self):
...
         ]
 
+        # Get the API gateway endpoint
+        match (self.config.grpc_data_stream_network):
+            case "mainnet":
+                grpc_data_stream_endpoint = "grpc.mainnet.aptoslabs.com:443"
+            case "testnet":
+                grpc_data_stream_endpoint = "grpc.testnet.aptoslabs.com:443"
+            case "devnet":
+                grpc_data_stream_endpoint = "grpc.devnet.aptoslabs.com:443"
```
bowenyang007, line 134: why are these hardcoded in code instead of in the config?

rtso: oh i originally moved them over to code so that it's easier to just specify mainnet/testnet/devnet in config. but with the api gateway it's already pretty self explanatory. i'll move it back.

larry-aptos (approved):

xbtmatt (approved):

PR #108, opened 2023-08-28
[Rust] Improve reconnection logic

rtso:
There are 2 cases where we need to retry connection
1. When response stream returns an error
2. When response stream returns no more items

This PR moves the reconnection logic to account for both cases.

## Testing
* With ending_version
* Close the GRPC stream and check that it reconnects (TBD)

### DIFF
--- a/rust/processor/src/worker.rs
+++ b/rust/processor/src/worker.rs
@@ -262,95 +262,119 @@ impl Worker {
             let mut last_insertion_time = std::time::Instant::now();
             let mut next_version_to_fetch = batch_start_version;
             let mut last_reconnection_time: Option<std::time::Instant> = None;
-            while let Some(current_item) = resp_stream.next().await {
-                match current_item {
-                    Ok(r) => {
-                        let start_version = r.transactions.as_slice().first().unwrap().version;
-                        let end_version = r.transactions.as_slice().last().unwrap().version;
-                        next_version_to_fetch = end_version + 1;
+            loop {
+                while let Some(current_item) = resp_stream.next().await {
+                    match current_item {
+                        Ok(r) => {
+                            let start_version = r.transactions.as_slice().first().unwrap().version;
+                            let end_version = r.transactions.as_slice().last().unwrap().version;
+                            next_version_to_fetch = end_version + 1;
 
-                        TRANSMITTED_BYTES_COUNT
-                            .with_label_values(&[processor_name])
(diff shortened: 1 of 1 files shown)


rust/processor/src/worker.rs
```diff
@@ -262,95 +262,119 @@ impl Worker {
...
+                    if channel_capacity == BUFFER_SIZE {
+                        break;
+                    }
+                    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+                }
+                info!("[Parser] The stream is ended.");
+
+                // Don't try to reconnect if we've reached ending_version
+                if let Some(ending_version) = ending_version {
+                    if next_version_to_fetch >= ending_version {
```
larry-aptos, line 330: ending version is inclusive.

rust/processor/src/worker.rs
```diff
@@ -262,95 +262,119 @@ impl Worker {
...
+
+                // Handle reconnection logic
+                tracing::warn!(
+                    processor_name = processor_name,
+                    stream_address = indexer_grpc_data_service_address.clone(),
+                    "[Parser] Error receiving datastream response."
+                );
+                if let Some(lrt) = last_reconnection_time {
+                    let elapsed: u64 = lrt.elapsed().as_secs();
+                    if elapsed < MIN_SEC_BETWEEN_GRPC_RECONNECTS {
```
larry-aptos, line 349: panic results in another retry since pod restarts

larry-aptos, line 349: is `MIN_SEC_BETWEEN_GRPC_RECONNECTS` defined in this file

rtso: yep it's defined at 60 seconds at the top of this file

larry-aptos (approved):
overall looks good. ty!

rust/processor/src/worker.rs
larry-aptos: maybe sleep and reconnect?

larry-aptos: sounds good. 

some of my thoughts:
1. add backoff to client retry.
2. server side reconnection signal.

rtso: I think Bowen mentioned before that exponential backoff can help make the pod never restart, but it may also hide upstream infra failures. We should only need to reconnect once every 30 minutes. If the pod's restarting more than that, then it's likely more of an infra issue that we need to fix.

PR #112, opened 2023-08-30
Add aptos-indexer-protos package for TS

banool:
## Summary
This PR does the following:
- Creates the aptos-indexer-protos package, which contains all the code generated from the protos, basic instructions for using them, a guide explaining how to publish the package, etc.
- Switches all the code under `typescript/` to use this package. I modernize the packaging / tsc stuff as part of this.

I will update the proto generation stuff in the GH workflows in a later PR.

## Callouts
Some things worth calling out that we might want to discuss:
- I have configured the generated code to represent u64s using bigint. Using string isn't very clear to the user and using number is dangerous.
- You must refer to types in the generated code using the absolute "path" through the modules, e.g. `aptos.indexer.v1.GetTransactionsRequest`). Through a long discussion with some TS experts this seemed like the best option while maintaining compatibility with various downstream development environments while also maintaining our own sanity: https://aptos-org.slack.com/archives/C03JJF9DJ4V/p1693414593545719.
- The getter and setter methods are gone from the generated code, it is just properties now. Imo this is cleaner and more modern but one downside is you have to null check yourself now, like `transaction.user!`. This is how it works in Rust too though so it's more standard this way.

## Test Plan
I have tested that the library can be used from both commonjs and es code.

Testing our own example (commonjs):
```
$ cat /tmp/config.yaml
grpc_data_stream_endpoint: grpc.testnet.aptoslabs.com:443
grpc_data_stream_api_key: <token>
db_connection_uri: "postgresql://postgres:postgres@localhost:5432/example_event"
health_port: 8080
indexer_grpc_http2_ping_interval_in_secs: 30
indexer_grpc_http2_ping_timeout_in_secs: 5
starting_version: 0
chain_id: 2
```
```
cd typescript
pnpm build
node dist/processors/example-write-set-change-processor/processor.js process --config /tmp/config.yaml
```
```
{
  message: 'Successfully processed transaction',
  last_success_transaction_version: 39000n
}
{ message: 'Response received', starting_version: 40000n }
{
  message: 'Successfully processed transaction',
  last_success_transaction_version: 40000n
}
{ message: 'Response received', starting_version: 41000n }
{
  message: 'Successfully processed transaction',
  last_success_transaction_version: 41000n
}
```


typescript/patches/@grpc+grpc-js+1.8.14.patch
rtso, line 1: I think we need this patch to fix an issue with the grpc-js library. More disucussion in https://aptos-org.slack.com/archives/C04PRP1K1FZ/p1683760992469519?thread_ts=1683670305.272199&cid=C04PRP1K1FZ

typescript/processors/example-write-set-change-processor/processor.ts
```diff
@@ -1,62 +1,22 @@
...
 import { program } from "commander";
-import { createDataSource } from "@/processors/example-write-set-change-processor/models/data-source";
-import { Event } from "@/processors/example-write-set-change-processor/models/Event";
-import {
-  INDEXER_NAME,
-  parse,
-} from "@/processors/example-write-set-change-processor/event-parser";
-import { NextVersionToProcess } from "@/utils/common_models/NextVersionToProcess";
-
-// A hack to override the http2 settings
```
rtso, line 20: Why delete this?

typescript/patches/@grpc+grpc-js+1.8.14.patch
banool: Okay good to know, let me add it back.

rtso: Sweet, can we add a comment in this explaining what it's for? So we don't accidentally delete it later on

typescript/processors/example-write-set-change-processor/processor.ts
jjleng: no, this was not related to TLS. This is the settings we needed for the http2 flow control. Without these settings, the grpc link is very slow. With this, the link has similar performance as Python

banool: Okay got it, I'll try to make this work with TLS then.

banool: I'm having a bit of trouble getting this working, asking about it here: https://github.com/grpc/grpc-node/issues/2565.

rtso (approved):

buf.ts.gen.yaml
```diff
@@ -2,5 +2,18 @@ version: v1
 managed:
   enabled: true
 plugins:
-  - name: ts
-    out: typescript/
+  - plugin: buf.build/community/stephenh-ts-proto
```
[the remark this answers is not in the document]
banool: Yeah in the top level README. I clean that up a bit further in the PR after this.

PR #113, opened 2023-08-30
[Rust] clean up reconnect logic

bowenyang007:
* Remove the nested loops
* Add a retry field so that if the upstream server is rebuilding we won't panic. This will potentially create more short lived connections which is something we should watch out for.
* Reduce the reconnect time to 15 seconds from 60. 

Test: I still see restarts in kube. Hopefully no more. 

image

### DIFF
--- a/rust/processor/src/worker.rs
+++ b/rust/processor/src/worker.rs
@@ -53,7 +53,10 @@ const BUFFER_SIZE: usize = 50;
 // 20MB
 const MAX_RESPONSE_SIZE: usize = 1024 * 1024 * 20;
 // We will try to reconnect to GRPC once every X seconds if we get disconnected, before crashing
-const MIN_SEC_BETWEEN_GRPC_RECONNECTS: u64 = 60;
+// We define short connection issue as < 10 seconds so adding a bit of a buffer here
+const MIN_SEC_BETWEEN_GRPC_RECONNECTS: u64 = 15;
+// We will try to reconnect to GRPC 5 times in case upstream connection is being updated
+const RECONNECTION_MAX_RETRIES: u64 = 5;
 
 pub struct Worker {
     pub db_pool: PgDbPool,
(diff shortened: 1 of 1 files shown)


rust/processor/src/worker.rs
```diff
@@ -262,119 +265,102 @@ impl Worker {
...
-                    }
-                    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
-                }
-                info!("[Parser] The stream is ended.");
-
-                // Don't try to reconnect if we've reached ending_version
-                if let Some(ending_version) = ending_version {
-                    if next_version_to_fetch > ending_version {
+            let mut reconnection_retries = 0;
+            while let Some(current_item) = resp_stream.next().await {
```
rtso, line 269: The reason I added reconnection logic wrapping around this while loop is because I saw logs where `[Parser] The stream is ended.` every 5 minutes, which is when the GRPC restarts. We should handle these restarts because they're expected to happen.

rtso (changes requested):

rust/processor/src/worker.rs
bowenyang007: yea good point. changed the logic

rust/processor/src/worker.rs
```diff
@@ -262,118 +265,129 @@ impl Worker {
...
+                        info!(
+                            processor_name = processor_name,
+                            channel_size = BUFFER_SIZE - channel_capacity,
+                            "[Parser] Waiting for channel to be empty"
+                        );
+                        if channel_capacity == BUFFER_SIZE {
+                            break;
+                        }
+                        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+                    }
```
rtso, line 349: Can we move this outside of the `is_end` so it's clearer? This `if/else` is not necessary.

The general flow I propose is:
1. Read from the stream
2. When the stream errors or naturally restarts, finish processing whatever is left in the consumer channel.
3. If ending version is reached, do nothing. Otherwise, try to reconnect.

rtso (changes requested):

rust/processor/src/worker.rs
bowenyang007: I disagree actually. The only time we should have the loop to clear the channel is if we get to the end, otherwise we should just panic. This way it's clear what the purpose of that extra loop is. The whole thing is pretty confusing as is, I guarantee you a few months down the line we're not gonna remember why we need to loop even though we're not at the end of the list and we'll have a debate about it again.

rust/processor/src/worker.rs
```diff
@@ -706,3 +582,168 @@ pub async fn get_stream(
...
+                    );
+                    panic!("[Parser] Recently reconnected. Will not retry.")
+                }
+            }
+            reconnection_retries += 1;
+            last_reconnection_time = Some(std::time::Instant::now());
+            tracing::warn!(
+                processor_name = processor_name,
+                stream_address = indexer_grpc_data_service_address.clone(),
+                starting_version = next_version_to_fetch,
```
rtso, line 732: If we panic early (before the consumer channel finishes processing, using `next_version_to_fetch` as the new starting version would potentially skip some transactions that were lost in the consumer channel. 

Which is why I suggested waiting for the consumer channel to clear before attempting reconnect.

rust/processor/src/worker.rs
bowenyang007: also if we're going to restart, why do we want to do the loop? The loop is blocking so it'll slow down fetching.

rust/processor/src/worker.rs
bowenyang007: wait what how is this possible?

rtso (approved):

PR #114, opened 2023-08-31
Update GH workflows / READMEs to use proto packages

banool:
## Summary
There is now a proto package for all 3 languages and all the processors have been updated to use this package rather than rely on the protos locally. This means we can remove the proto generation from the build and lint steps.

To make sure we know when the protos and the generated code are out of sync, I have added a new non-blocking workflow that generates code based on the protos and ensures that the generated code matches what is checked in.

I fixed rust-setup and use it in the new check-protos.yaml workflow. I had to add a rust-toolchain file. Accordingly I made the Rust linter just use nightly across the board and fixed the one new lint it had accordingly (in rust/processor/src/utils/util.rs).

Building the image for Python is about 40% faster now.

## Test Plan
CI.


buf.ts.gen.yaml
```diff
@@ -2,5 +2,18 @@ version: v1
 managed:
```
larry-aptos, line 2: what's the relationship between `pnpm install -g protoc-gen-ts` and this file

banool: Yeah that's right.

larry-aptos (approved):

PR #115, opened 2023-08-31
Create TS SDK package

banool:
## Description
This PR adds the `@aptos-labs/aptos-processor-sdk` TS package. This package contains nothing fundamentally new, it just extracts all the generic stuff from what we already had and puts it in a package, separate from the stuff specific to the example processor.

The example processor has been moved into `examples/` and made to use the sdk package.

I've done a bunch of other necessary small stuff like setting up linting, fixing up the imports, updating our tsc configs, etc.

After all that I did make one other major change. Whereas before there was a function that users were expected to define that outputs entities to insert, I changed it to having the SDK just handle tracking the txn version and then the user is expected to do DB writes in their processor function. I then wrapped this all up in classes and cleaned up a bit. This matches how we do it in Python.

I will change the local path deps to refer to the published package before landing, as well as add a README for the SDK.

## Test Plan
First, prepare `/tmp/config.yaml` by copying `examples/event_processor/config.yaml.example`. Then:
```
cd examples/event_processor
pnpm install
pnpm start process --config ~/config.yaml
```
```
Requesting stream starting from version 0
{ message: 'Response received', starting_version: 0n }
{ message: 'Response received', starting_version: 1000n }
{ message: 'Response received', starting_version: 2000n }
{ message: 'Response received', starting_version: 3000n }
{ message: 'Response received', starting_version: 4000n }
{ message: 'Response received', starting_version: 5000n }
{ message: 'Response received', starting_version: 6000n }
{ message: 'Response received', starting_version: 7000n }
{ message: 'Response received', starting_version: 8000n }
{ message: 'Response received', starting_version: 9000n }
{ message: 'Response received', starting_version: 10000n }
{ message: 'Response received', starting_version: 11000n }
{ message: 'Response received', starting_version: 12000n }
...
```


typescript/sdk/src/models/next_version_to_process.ts
```diff
@@ -0,0 +1,24 @@
+import { Column, Entity, PrimaryColumn } from "typeorm";
+import { Base } from "./base";
+
+@Entity("next_version_to_process")
+export class NextVersionToProcess extends Base {
```
rtso, line 5: A note about `NextVersionToProcess` -

We should standardize how we track versions across all the languages. I was planning to update Python to track txn version using the same schema as Rust's `processor_status`. It tracks the latest processed version rather than the next version to process. Let's do the same for TS to standardize.

typescript/sdk/src/models/data_source.ts
```diff
@@ -0,0 +1,40 @@
+import "reflect-metadata";
+import { DataSource } from "typeorm";
+import { SnakeNamingStrategy } from "typeorm-naming-strategies";
+import { NextVersionToProcess } from "./next_version_to_process";
+import { Base } from "./base";
+
+// TODO: Do migrations stuff??, the below function takes migrations as an argument.
```
rtso, line 7: Yeah we need migrations, but good thing it seems like TypeORM supports this and you don't need a separate library.

typescript/sdk/src/worker.ts
```diff
@@ -0,0 +1,216 @@
...
+      async (response: aptos.indexer.v1.TransactionsResponse) => {
+        stream.pause();
+        const transactions = response.transactions;
+
+        if (transactions == null) {
+          return;
+        }
+
+        // Validate response chain ID matches expected chain ID
+        if (response.chainId != this.config.chain_id) {
```
rtso, line 125: Same comment as above, we need to change this to compare it to the `chain_id` stored in `ledger_info`.

typescript/sdk/src/worker.ts
```diff
@@ -0,0 +1,216 @@
...
+          throw new Error(
+            `Chain ID mismatch. Expected ${this.config.chain_id} but got ${response.chainId}`,
+          );
+        }
+
+        if (transactions == null) {
+          return;
+        }
```
rtso, line 133: This is duplicated from line 120-122.

typescript/sdk/src/worker.ts
```diff
@@ -0,0 +1,216 @@
...
+
+        const startVersion = transactions[0].version!;
+        const endVersion = transactions[transactions.length - 1].version!;
+
+        console.log({
+          message: "Response received",
```
rtso, line 139: Can we prefix all log messages with `[Parser]`? I'll also need to do a logging audit for all processors later on.

typescript/sdk/src/worker.ts
```diff
@@ -0,0 +1,216 @@
...
+  // Base configuration required to connect to the database and the data stream.
+  config: Config;
+
+  // The processor we'll run.
+  processor: TransactionsProcessor;
+
+  // The DB connection.
+  dataSource: DataSource;
+
+  constructor({
```
rtso, line 29: This is a pretty good start! Off the top of my head, it's missing a couple of key features: multithreading, GRPC reconnection, checking for gaps in txn version. We should add these before publishing the SDK.

typescript/examples/event_processor/config.yaml.example
```diff
@@ -0,0 +1,6 @@
+indexer_name: "event_processor"
```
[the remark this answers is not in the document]
banool: Ooop yeah my bad, I meant to remove this since I removed it from the config, the processors now report their own name.

As for multi processor deployment, yeah probably, though I wonder if putting them all in the same binary makes sense. I think having different build targets for each and they use shared code would maybe be better rather than having all part of one binary. Each has trade offs, we can discuss.

typescript/examples/event_processor/config.yaml.example
```diff
@@ -0,0 +1,6 @@
...
+chain_id: 2
```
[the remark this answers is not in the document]
banool: Sounds good! Let's do that in a later PR though, I've tried to keep this one as light as possible while still being functional.

typescript/examples/event_processor/config.yaml.example
```diff
@@ -0,0 +1,6 @@
...
+grpc_data_stream_endpoint: grpc.testnet.aptoslabs.com:443
+grpc_data_stream_api_key: <api_key>
+starting_version: 0
```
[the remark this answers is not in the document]
banool: As above, sounds good but let's do it in a later PR.

typescript/sdk/src/models/next_version_to_process.ts
banool: I think latest processed version makes more sense yeah. Can do in a later PR, I'll take notes for all these changes.

typescript/sdk/src/models/data_source.ts
banool: Without specifying anything it seems to just apply the migrations for us so I think we might be good here.

banool:
I addressed the smaller issues you called out above. We can do the rest in later PRs.

rtso (approved):

rtso:
LGTM! Should we wait to publish the package until we get the rest of the PRs?

banool:
Yeah for sure, no need to publish yet.

typescript/examples/event_processor/config.yaml.example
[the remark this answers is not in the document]
banool: Yeah probably simplest to have them all in one binary and worry about perf / size issues down the line. We could avoid the binary nightmare to some extent by having them all in the same image but let's not worry about any of that for now.

PR #117, opened 2023-09-01
Add index on account_address in account_transactions table

rtso:


jillxuu (approved):

bowenyang007:
discussed offline, we already have an index on the field.

PR #119, opened 2023-09-02
Install pnpm in GH workflow, fix tagging in release-processor-images.mjs

banool:
## Summary
I think installing the pnpm deps from inside the script is a bit irregular. I instead make the script something you just run via a script entry in package.json. I think everyone knows you should run `pnpm install` before running a JS script.

Beyond that, the script was broken, for example the way we were invoking `crane tag` was incorrect. I fix that in this PR.

## Test Plan
```
gh workflow run 'Release Processor Images' --ref banool/release-processor-images -f GIT_SHA=de766ae36a1c74212d370aae75bf234320612bbc -f version_tag=v1.0.4
```

### DIFF
--- a/.github/workflows/copy-processor-images-to-dockerhub.yaml
+++ b/.github/workflows/copy-processor-images-to-dockerhub.yaml
@@ -23,7 +23,7 @@ permissions:
 jobs:
   copy-processor-images:
     # Run on a machine with more local storage for large docker images
-    runs-on: medium-perf-docker-with-local-ssd
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


sherry-x (approved):
Looks great, thank you!

larry-aptos (approved):

.github/workflows/copy-processor-images-to-dockerhub.yaml
rustielin, line 1: do we really need to be running this with `medium-perf-docker-with-local-ssd` ? normal runner should be fine

rustielin (approved):

PR #122, opened 2023-09-05
[ECO-413] Add V2 market registration processor

elliottdehn:
This is the humble beginnings of our chain data processor, now on V2 and processing market registration data. The implementation is based on the `default_processor` given to us in the project.

I've not strived for particular elegance or grace here--I merely went with what was minimal and worked. Abstraction is low.


elliottdehn:
Ignore this, it was made by mistake.

PR #128, opened 2023-09-09
[Rust] Storage refund

bowenyang007:
Add new field to coin and fungible asset activities to indicate storage refund. 

## Backfill
It's not in testnet or mainnet just yet, so using ledger for now
Testnet: 667675894
Mainnet: 255894550

## Test
image
image

### DIFF
--- a/rust/processor/src/models/fungible_asset_models/v2_fungible_asset_activities.rs
+++ b/rust/processor/src/models/fungible_asset_models/v2_fungible_asset_activities.rs
@@ -6,7 +6,9 @@
 #![allow(clippy::unused_unit)]
 
 use super::{
-    v2_fungible_asset_utils::{FungibleAssetAggregatedDataMapping, FungibleAssetEvent},
+    v2_fungible_asset_utils::{
+        FeeStatement, FungibleAssetAggregatedDataMapping, FungibleAssetEvent,
+    },
     v2_fungible_metadata::FungibleAssetMetadataModel,
 };
 use crate::{
@@ -22,7 +24,7 @@ use crate::{
 };
 use anyhow::Context;
 use aptos_indexer_protos::transaction::v1::{Event, TransactionInfo, UserTransactionRequest};
-use bigdecimal::BigDecimal;
+use bigdecimal::{BigDecimal, Zero};
 use field_count::FieldCount;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
@@ -57,6 +59,7 @@ pub struct FungibleAssetActivity {
     pub block_height: i64,
     pub token_standard: String,
     pub transaction_timestamp: chrono::NaiveDateTime,
+    pub storage_refund_amount: BigDecimal,
 }
 
 impl FungibleAssetActivity {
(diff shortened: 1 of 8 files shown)


rust/processor/migrations/2023-09-01-231248_events_v2/up.sql
```diff
@@ -0,0 +1,5 @@
+-- Your SQL goes here
+ALTER TABLE events DROP CONSTRAINT events_pkey;
+ALTER TABLE events DROP CONSTRAINT IF EXISTS fk_transaction_versions;
+ALTER TABLE events
+ADD CONSTRAINT events_pkey PRIMARY KEY (transaction_version, event_index);
```
rtso, line 5: was this already done in your last PR? i think you need to rebase on main

rust/processor/src/models/fungible_asset_models/v2_fungible_asset_activities.rs
```diff
@@ -182,6 +187,7 @@ impl FungibleAssetActivity {
         transaction_version: i64,
         transaction_timestamp: chrono::NaiveDateTime,
         block_height: i64,
+        fee_statement: Option<FeeStatement>,
```
rtso, line 190: Does the comment mean that we only have storage refund for fungible asset v1 (not v2)?

rust/processor/migrations/2023-09-01-231248_events_v2/up.sql
bowenyang007: Yes

rust/processor/src/models/fungible_asset_models/v2_fungible_asset_activities.rs
bowenyang007: I added this to coin_activities as well.

rtso: I mean it seems like this function should only be indexing v1 activities, but I guess it's not anymore? So let's change the variables naming to clarify.

rust/processor/src/processors/fungible_asset_processor.rs
```diff
@@ -408,13 +408,19 @@ fn parse_v2_coin(
         for (index, event) in events.iter().enumerate() {
             // The artificial gas event, only need for v1
```
rtso, line 409: Also this comment.

rust/processor/src/models/fungible_asset_models/v2_fungible_asset_activities.rs
bowenyang007: sorry i don't understand your question. this is used for gas. not sure what function you're referring to.

rtso: I'm referring to `get_gas_event`

bowenyang007: ah i see. regardless of the naming, the plan is to deprecate coin_activities at some point down the line. Idk if we need to beat down on this variable naming. I do want to migrate all this code over to fungible_asset_activities instead of doing this.

rtso (approved):

rust/processor/src/processors/fungible_asset_processor.rs
bowenyang007: fair, some clean up needed.

PR #131, opened 2023-09-11
[NFT Metadata Crawler] Add trim to URIs

just-in-chang:
Trim URIs because some contain new lines at the end


GhostWalker562 (approved):
LGTM 🚀

PR #132, opened 2023-09-11
Revert "[NFT Metadata Crawler] Add trim to URIs"

just-in-chang:
Reverts aptos-labs/aptos-indexer-processors#131 to move trimming logic more down stream


GhostWalker562 (approved):
😞

PR #134, opened 2023-09-15
Move Rust processor config into lib

banool:
This lets you import it from other places, main should be thin.


just-in-chang (approved):
Had to do the same thing on aptos-core so that I wouldn't have to pass over all the configs as individual parameters into the worker. This should help reduce a lot of redundant code with the worker bc now we can directly pass in the config struct instead of copying over each parameter in the constructor 👍

larry-aptos (approved):

larry-aptos:
Neat!

PR #138, opened 2023-09-19
[Rust] Fix user transaction timestamp and split processor

bowenyang007:
This causes issues for downstream, e.g. inserting into bigquery. 
Need to backfill from genesis

## Test
Before
image

After
image


ying-w (approved):
is parse_timestamps called for all timestamps or just user_transactions

bowenyang007:
> is parse_timestamps called for all timestamps or just user_transactions

there are 2 timestamp parsing. The only timestamp that can go wrong is this one though b/c it's user input. We do try to make sure that the other one also has a max.

rtso:
When can the timestamp be greater than the max? Like where's this user input coming from?

rust/processor/migrations/2023-09-11-164718_ut_remove_constraint/down.sql
```diff
@@ -0,0 +1,7 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE user_transactions DROP CONSTRAINT IF EXISTS fk_versions;
+ALTER TABLE signatures DROP CONSTRAINT IF EXISTS fk_transaction_versions;
+ALTER TABLE user_transactions
+ADD CONSTRAINT fk_versions FOREIGN KEY (version) REFERENCES transactions (version);
```
rtso, line 5: When do we need to add a foreign key to a table?

bowenyang007: At this point never. This is only needed if you want to maintain consistency in a normal app (e.g. I want to make sure that when I'm writing to a post table that the user already exists).

bowenyang007: Given our work is really post processing...data cannot be wrong. That's a nice property of the blockchain.

PR #139, opened 2023-09-21
Improve Rust processor configuration, error handling

banool:
## Summary
You can most concretely see how this improves configuration by looking at the sample config. It'd be even more obvious for a processor that has its own additional config.

This PR does the following:
- Moves configuration specific to each processor into configs for that processor.
- Creates the ProcessorConfig struct, which we collect directly in the config. Previously we had a bunch of Options and we unwrapped them later on based on the processor name. This is no longer necessary, we collect all the required variables at config loading time.
- Creates the Processor enum using enum_dispatch. This is more efficient than using dynamic dispatch and a little more type safe.
- Creates the ProcessorName enum based on ProcessorConfig.
- Switches a few locations from panicking to using anyhow Result + Context.
- Adds support for non TLS connections to the data service.
- Uses appropriate types for fields, e.g. Url.

Once this is approved I will update the configs in internal-ops later (all 43 of them lol). We will land those config changes hand in hand with upgrading to the processor image that has this change in it.

## Test Plan
Run postgres locally and create a DB called default_processor.

Run a local testnet with a txn stream (from the banool/grpc-fullnode-local-testnet branch):
```
cargo run -p aptos -- node run-local-testnet --assume-yes --with-faucet --force-restart
```

Run the processor:
```
cargo run -- -c parser.yaml
```

See that data is being indexed:
```
> SELECT * FROM transactions LIMIT 1;
0	0	0x3897eb9f4ec29124a3076c40114294aebff84aa7ae7f46966d3070ef58f60083	TRANSACTION_TYPE_GENESIS		0x449a043ae839ce117710f5f40b9b17a36ad76631d1c593fabfb3771408a3f1b0	0xe7fb78a08c78cae785eef221dd13094c1c5bb265302c2e086e3a87dc7782f83c	0x3adc322fe22b69ebc529adc9d48f366d964f0da60ffee97bd33f806bc2d3edc5	0	TRUE	Executed successfully	0x56507fac5a074955e7609278e98936ac38698d50ab410ca2922400c4a15fb1bc	12	152	2023-09-21 18:07:48.100134	0
```


rtso (approved):
LGTM

larry-aptos (approved):

PR #143, opened 2023-09-28
Fix nested runtimes issue with Rust processors

banool:
## Description
Currently we create a runtime from within the context of a runtime. This is a big no no with tokio and can cause a variety of issues, particularly when using this code from another place: https://stackoverflow.com/q/62536566/3846032.

To address this, instead of creating a runtime in `run_server_with_config`, I require that the caller passes in a handle.

## Test Plan
I ran the formatter, linter, tests, and succesfully ran a processor against a txn stream.

### DIFF
--- a/rust/server-framework/src/lib.rs
+++ b/rust/server-framework/src/lib.rs
@@ -6,6 +6,7 @@ use clap::Parser;
 use prometheus::{Encoder, TextEncoder};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{fs::File, io::Read, panic::PanicInfo, path::PathBuf, process};
+use tokio::runtime::Handle;
 use tracing::error;
 use tracing_subscriber::EnvFilter;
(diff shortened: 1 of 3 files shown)


rust/server-framework/src/lib.rs
```diff
@@ -19,34 +20,31 @@ pub struct ServerArgs {
...
 }
 
-pub async fn run_server_with_config<C>(config: GenericConfig<C>) -> Result<()>
+/// Run a server and the necessary probes. For spawning these tasks, the user must
+/// provide a handle to a runtime they already have.
+pub async fn run_server_with_config<C>(config: GenericConfig<C>, handle: Handle) -> Result<()>
 where
     C: RunnableConfig,
 {
-    let runtime = tokio::runtime::Builder::new_multi_thread()
```
larry-aptos, line 38: yea the runtime is dumb... i also removed it in grpc side lol

larry-aptos (approved):

PR #146, opened 2023-10-01
Switch to diesel-async

banool:
## Summary
There are two main reasons to do this:
- Better perf since we don't block waiting for I/O to happen (esp if I can get pipelining to work).
- Breaking the dep on libpq.

As you can see, at least without pipelining, this PR has no real effect on perf, but in neither direction, it's just the same.

Update: I think I got pipelining to work (at least, it should be working as per the docs), but the impact was negligible. Probably our workload just doesn't benefit from it, not too sure: https://github.com/aptos-labs/aptos-indexer-processors/pull/147.

### Performance testing
Using this config:
```
health_check_port: 8084
server_config:
  processor_config:
    type: default_processor
  postgres_connection_string: postgresql://postgres:@localhost:5432/default_processor
  indexer_grpc_data_service_address: https://grpc.testnet.aptoslabs.com
  auth_token: <token>
```

Note: Prior to these results I ran this once all the way up to txn 2.5 million to warm up the cache upstream (just in case).

Experiment results. What txn we got up to after 3 minutes (last_success_version in the DB):

Built from main
1. Run 1: 2261999
1. Run 2: 2162999
1. Run 3: 2091999
1. Average: 2172332

Built from this branch
1. Run 1: 2250999
1. Run 2: 2231999
1. Run 3: 2030999
1. Average: 2171332
  
I alternated runs, so I did run 1 for each, then run 2 for each, then run 3 for each.

I ran tests with degraded networking: 100ms latency on downlink and uplink, 1% packet loss, and 50mbps bandwidth down / 10mbps up. There was no noticeable difference between the two builds (we weren't I/O bound in the end).

I used this command. I made sure the binary was built first:
```
timeout 180 target/release/processor -c processor/parser.yaml
```

### DIFF
--- a/rust/processor/src/processors/token_v2_processor.rs
+++ b/rust/processor/src/processors/token_v2_processor.rs
@@ -27,7 +27,8 @@ use crate::{
     schema,
     utils::{
         database::{
-            clean_data_for_db, execute_with_better_error, get_chunks, PgDbPool, PgPoolConnection,
+            clean_data_for_db, execute_with_better_error, get_chunks, MyDbConnection, PgDbPool,
+            PgPoolConnection,
         },
         util::{get_entry_function_from_user_request, parse_timestamp, standardize_address},
     },
@@ -37,7 +38,7 @@ use aptos_indexer_protos::transaction::v1::{
     transaction::TxnData, write_set_change::Change, Transaction,
 };
 use async_trait::async_trait;
-use diesel::{pg::upsert::excluded, result::Error, ExpressionMethods, PgConnection};
+use diesel::{pg::upsert::excluded, result::Error, ExpressionMethods};
 use field_count::FieldCount;
 use std::{
     collections::{HashMap, HashSet},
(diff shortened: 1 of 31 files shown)


banool:
The alternative would be to do this: https://github.com/sfackler/rust-postgres/issues/890.

banool:
See the benchmarking from the Diesel team themselves: https://github.com/diesel-rs/metrics.

rust/processor/src/processors/token_v2_processor.rs
```diff
@@ -77,19 +78,19 @@ fn insert_to_db_impl(
...
 ) -> Result<(), diesel::result::Error> {
-    insert_collections_v2(conn, collections_v2)?;
-    insert_token_datas_v2(conn, token_datas_v2)?;
-    insert_token_ownerships_v2(conn, token_ownerships_v2)?;
-    insert_current_collections_v2(conn, current_collections_v2)?;
-    insert_current_token_datas_v2(conn, current_token_datas_v2)?;
-    insert_current_token_ownerships_v2(conn, current_token_ownerships_v2)?;
-    insert_token_activities_v2(conn, token_activities_v2)?;
-    insert_current_token_v2_metadatas(conn, current_token_v2_metadata)?;
+    insert_collections_v2(conn, collections_v2).await?;
```
larry-aptos, line 81: qq, we're sequentially inserting data into multiple tables under one transaction, is it doable that we use `tokio::spawn` and `select_all` to parallelize the db operations?

larry-aptos: oh i see the pipelining part.

bowenyang007 (approved):

PR #170, opened 2023-10-24
[cockroach] write fullnode raw txn processors to cockroachdb

jillxuu:
relational diagram: https://dbdiagram.io/d/65330709ffbf5169f023ce51

tables created in cockroachdb: `transactions`, `events`, `write_set_changes_resource`, `write_set_changes_table`, `write_set_changes_module`


rust/processor/src/worker.rs
```diff
@@ -618,6 +647,9 @@ impl Worker {
             let batch_end = processed_versions_sorted.last().unwrap().end_version;
             batch_start_version = batch_end + 1;
 
+            LATEST_PROCESSED_VERSION
```
larry-aptos, line 650: current `LATEST_PROCESSED_VERSION` contains 3 values.

rust/processor/migration/src/m20220101_000001_create_table.rs
```diff
@@ -0,0 +1,155 @@
+use sea_orm::Statement;
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
```
grao1991, line 9: I noticed `transaction_version` is used as the first part of primary key for many tables, won't it cause the bottleneck on write qps since we are roughly writing our data in monotonic increasing order and most recent writes will hit the same machine and prevent us from horizontal scaling the write qps?

PR #178, opened 2023-11-03
[rust] Harden processors logging

rtso:
* Add more logging to processors code that include tps, kbps, duration, etc. [Logging spec](https://www.notion.so/aptoslabs/Indexer-Metrics-Logging-Spec-d60a5ed2c22e4d2a9910045e3e7283b7)
* Add a timeout to the consumer thread to panic if it doesn't receive any transactions in 15s. 

Open to suggestions to make the logging cleaner. 

## Testing
Consumer thread panicked 
Screenshot 2023-11-06 at 12 22 35 PM

`cargo run --release -- -c config.yaml --verbose true`
Screenshot 2023-11-06 at 12 24 39 PM

connection id string is logged
Screenshot 2023-11-06 at 12 22 47 PM

Prometheus metrics
Screenshot 2023-11-09 at 12 45 28 PM

### DIFF
--- a/rust/processor/src/config.rs
+++ b/rust/processor/src/config.rs
@@ -25,7 +25,7 @@ pub struct IndexerGrpcProcessorConfig {
 
 #[async_trait::async_trait]
 impl RunnableConfig for IndexerGrpcProcessorConfig {
-    async fn run(&self) -> Result<()> {
+    async fn run(&self, verbose: Option<bool>) -> Result<()> {
         let mut worker = Worker::new(
             self.processor_config.clone(),
             self.postgres_connection_string.clone(),
@@ -35,6 +35,7 @@ impl RunnableConfig for IndexerGrpcProcessorConfig {
             self.starting_version,
             self.ending_version,
             self.number_concurrent_processing_tasks,
+            verbose,
         )
         .await
         .context("Failed to build worker")?;
--- a/rust/processor/src/config.rs
+++ b/rust/processor/src/config.rs
@@ -21,11 +21,12 @@ pub struct IndexerGrpcProcessorConfig {
     pub starting_version: Option<u64>,
     pub ending_version: Option<u64>,
     pub number_concurrent_processing_tasks: Option<usize>,
+    pub enable_verbose_logging: Option<bool>,
 }
 
 #[async_trait::async_trait]
 impl RunnableConfig for IndexerGrpcProcessorConfig {
-    async fn run(&self, verbose: Option<bool>) -> Result<()> {
+    async fn run(&self) -> Result<()> {
         let mut worker = Worker::new(
             self.processor_config.clone(),
             self.postgres_connection_string.clone(),
@@ -35,7 +36,7 @@ impl RunnableConfig for IndexerGrpcProcessorConfig {
             self.starting_version,
             self.ending_version,
             self.number_concurrent_processing_tasks,
-            verbose,
+            self.enable_verbose_logging,
         )
         .await
         .context("Failed to build worker")?;
--- a/rust/processor/src/utils/counters.rs
+++ b/rust/processor/src/utils/counters.rs
@@ -100,16 +100,76 @@ pub static LATEST_PROCESSED_VERSION: Lazy<IntGaugeVec> = Lazy::new(|| {
     register_int_gauge_vec!(
         "indexer_processor_latest_version",
         "Latest version a processor has fully consumed",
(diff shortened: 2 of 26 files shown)


rust/processor/src/utils/util.rs
```diff
@@ -413,6 +414,41 @@ pub fn get_name_from_unnested_move_type(move_type: &str) -> &str {
...
+// }
+
+// pub fn calculate_size_in_kbs(vectors: Vec<Box<dyn std::fmt::Debug>>) {
+//     vectors
+//         .iter()
+//         .fold(0.0, |acc, v| acc + mem::size_of_val(&v[0]) * v.len())
+// }
+
+#[macro_export]
+macro_rules! calculate_size_in_kbs {
```
bowenyang007, line 436: provide some reference on where this comes from? Is this chatgpt or stack overflow?

rust/processor/src/utils/util.rs
```diff
@@ -413,6 +414,41 @@ pub fn get_name_from_unnested_move_type(move_type: &str) -> &str {
...
+//     let total_size_in_bytes = mem::size_of_val(&vector[0]) * vector.len();
+//     let total_size_in_kb = total_size_in_bytes as f64 / 1024.0;
+//     total_size_in_kb
+// }
+
+// pub fn calculate_size_in_kbs(vectors: Vec<Box<dyn std::fmt::Debug>>) {
+//     vectors
+//         .iter()
+//         .fold(0.0, |acc, v| acc + mem::size_of_val(&v[0]) * v.len())
+// }
```
bowenyang007, line 433: remove

rust/server-framework/src/lib.rs
```diff
@@ -17,6 +17,8 @@ use warp::{http::Response, Filter};
 pub struct ServerArgs {
     #[clap(short, long, value_parser)]
     pub config_path: PathBuf,
+    #[clap(short, long)]
+    pub verbose: bool,
```
rtso, line 21: How can I update server-framework to work with the flag?

rust/processor/src/config.rs
```diff
@@ -21,11 +21,12 @@ pub struct IndexerGrpcProcessorConfig {
     pub starting_version: Option<u64>,
     pub ending_version: Option<u64>,
     pub number_concurrent_processing_tasks: Option<usize>,
+    // pub enable_verbose_logging: Option<bool>,
```
bowenyang007, line 24: add verbose here lol

rtso: oh i guess we don't need to because it's passed as a param in the `run` function

rtso: is that the right way to do it @larry-aptos ? i'm not completely sure how server-framework works

bowenyang007: Don't pass it in the run function lmao. This config struct exists so that run doesn't have to change.

bowenyang007:
In the summary can you also add all the field names you're using? So that we can have the same exact field names upstream.

rust/processor/src/config.rs
rtso: hmm this is gonna be hard bc it's a command line argument rather than a config.yaml field.

larry-aptos:
1. you can do a follow-up to parse the `x-aptos-conneciton-id` in the response metadata and log it.
2. there are some fields in your screenshot but i didn't find them in the PR; anything i missed.

Overall looks good!

larry-aptos (approved):

rust/processor/src/utils/counters.rs
```diff
@@ -100,16 +100,76 @@ pub static LATEST_PROCESSED_VERSION: Lazy<IntGaugeVec> = Lazy::new(|| {
...
+
+/// Overall processing time for multiple (n = number_concurrent_processing_tasks) batch of transactions
+pub static MULTI_BATCH_PROCESSING_TIME_IN_SECS: Lazy<GaugeVec> = Lazy::new(|| {
+    register_gauge_vec!(
+        "indexer_processor_multi_batch_processing_time_in_secs",
+        "Time taken to process multiple batches of transactions",
+        &["processor_name"]
+    )
+    .unwrap()
+});
```
bowenyang007, line 146: can we follow the same format here? i.e. duration with processor, step, and message as keys?

bowenyang007: nvm this is fine.

PR #181, opened 2023-11-09
[rust][fix] Fix move module crash

bowenyang007:
## Summary
Basically we didn't realize that ABI could be empty when uploading a new module. But it is possible. 

## Testing
No longer crashes
image

### DIFF
--- a/rust/processor/src/models/default_models/move_modules.rs
+++ b/rust/processor/src/models/default_models/move_modules.rs
@@ -90,14 +90,9 @@ impl MoveModule {
     pub fn convert_move_module_bytecode(
         mmb: &MoveModuleBytecode,
     ) -> Option<MoveModuleByteCodeParsed> {
-        let abi = mmb.abi.as_ref();
-        if abi.is_none() {
-            return None;
-        }
-        Some(Self::convert_move_module(
-            abi.unwrap(),
-            mmb.bytecode.clone(),
-        ))
+        mmb.abi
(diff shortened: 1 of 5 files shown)


rust/processor/src/models/default_models/move_modules.rs
```diff
@@ -86,8 +90,12 @@ impl MoveModule {
...
     ) -> Option<MoveModuleByteCodeParsed> {
+        let abi = mmb.abi.as_ref();
+        if abi.is_none() {
+            return None;
+        }
         Some(Self::convert_move_module(
-            mmb.abi.as_ref().unwrap(),
+            abi.unwrap(),
             mmb.bytecode.clone(),
         ))
```
CapCap, line 100: nit; I think you can use a `map` or a match on `abi`

bowenyang007: true. done.

larry-aptos:
lint is broken.

bowenyang007:
@larry-aptos it looks like it's from a previous PR...I fixed some random xclippy issue but seems like that wasn't enough.

larry-aptos:
working on it.

bowenyang007:
It's OK I fixed it.

larry-aptos:
neat.

larry-aptos (approved):

NancySirois (approved):

PR #182, opened 2023-11-09
[Aptos Tournament] Processor Skeleton

just-in-chang:
Adds Postgres migrations and a skeleton for a processor to index Aptos Tournament. 

Also reorganizes how we are saving and structuring schemas. We may need to find a better solution for managing and organizing multiple schemas because Diesel doesn't do a good job supporting it.


jillxuu (commented):
What about we create more than one diesel toml as suggested by this https://stackoverflow.com/questions/70378560/is-it-possible-to-make-diesel-to-handle-multi-database? And you can add tables you wanna exclude in toml

rust/processor/migrations/2023-11-09-065952_tournament_indexer/up.sql
```diff
@@ -0,0 +1,28 @@
+CREATE SCHEMA IF NOT EXISTS aptos_tournament;
```
bowenyang007, line 1: I don't think htat we need aseparate schema given this'll be a separate DB?

bowenyang007 (commented):
Can do a review with the front end team? The schema might change depending on that.

PR #183, opened 2023-11-09
fix: wrong auth token header

blockchainlover2019:
* Authorization Header is not correct

https://github.com/aptos-labs/aptos-core/blob/1d47982c93c5b6e7525b0070c85856b8df8c3a3f/ecosystem/indexer-grpc/indexer-grpc-data-service/src/config.rs#L118C4-L135

In `indexer-grpc-data-service`, token verification is working as follows
```
let authentication_inceptor =
            move |req: Request<()>| -> std::result::Result<Request<()>, Status> {
                if disable_auth_check {
                    return std::result::Result::Ok(req);
                }
                let metadata = req.metadata();
                if let Some(token) =
                    metadata.get(aptos_indexer_grpc_utils::constants::GRPC_AUTH_TOKEN_HEADER)
                {
                    if token_set.contains(token) {
                        std::result::Result::Ok(req)
                    } else {
                        Err(Status::unauthenticated("Invalid token"))
                    }
                } else {
                    Err(Status::unauthenticated("Missing token"))
                }
            };
```

But in indexer api processor, that auth token headers is wrongly set.

### DIFF
--- a/rust/processor/src/worker.rs
+++ b/rust/processor/src/worker.rs
@@ -39,7 +39,7 @@ use tracing::{error, info};
 use url::Url;
 
 /// GRPC request metadata key for the token ID.
-const GRPC_API_GATEWAY_API_KEY_HEADER: &str = "authorization";
+const GRPC_AUTH_TOKEN_HEADER: &str = "x-aptos-data-authorization";
 /// GRPC request metadata key for the request name. This is used to identify the
 /// data destination.
 const GRPC_REQUEST_NAME_HEADER: &str = "x-aptos-request-name";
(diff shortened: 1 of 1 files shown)


larry-aptos:
Thank you for your contribution! 

Good catch; there is a gap between current data service implementation and the processor, which makes the self-hosted end-2-end stack unusable. 

In the long term, data service will be no longer responsible for auth token verification; in other words, auth token will not be required to connect to data service. If auth is required, we recommend api gateway to handle this.

blockchainlover2019:
> Thank you for your contribution!
> 
> Good catch; there is a gap between current data service implementation and the processor, which makes the self-hosted end-2-end stack unusable.
> 
> In the long term, data service will be no longer responsible for auth token verification; in other words, auth token will not be required to connect to data service. If auth is required, we recommend api gateway to handle this.

Okay, noted.
But it is necessary to merge this PR at the moment, isn't it?

larry-aptos:
The authentication is going to be deprecated at data service side in the very short term. 

You can use `disable_auth_check` flag at data service side to disable it. The current header is required to connect to our developer portal; other than this, it's optional.

PR #184, opened 2023-11-09
[metrcs] add the latest processed version for python processor.

larry-aptos:


python/utils/metrics.py
```diff
@@ -1,5 +1,10 @@
-from prometheus_client import Counter
+from prometheus_client import Counter, Gauge
 
 PROCESSED_TRANSACTIONS_COUNTER = Counter(
-    "processed_transactions", "Number of transactions processed"
+    "indexer_py_processor_processed_transactions", "Number of transactions processed"
 )
+
+LATEST_PROCESSED_VERSION = Gauge(
+    "indexer_py_processor_latest_processed_version", "Latest processed version"
```
rtso, line 8: Let's add processor_name as a label to the metric

rtso (approved):

python/utils/metrics.py
larry-aptos: good call; added.

PR #190, opened 2023-11-13
migrate all logging to use the new util.

larry-aptos:
image

### DIFF
--- a/python/utils/config.py
+++ b/python/utils/config.py
@@ -66,7 +66,9 @@ class Config(BaseSettings):
                             next_version_to_process_from_db.next_version
                         )
             except:
-                logging.warn("Config] Database error when getting NextVersionToProcess. Skipping...")
+                logging.warn(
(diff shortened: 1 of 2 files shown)


python/utils/config.py
```diff
@@ -68,30 +67,20 @@ def get_starting_version(self, processor_name: str) -> int:
                         )
             except:
                 logging.warn(
-                    json.dumps(
-                        {
-                            "message": "Config] Database error when getting NextVersionToProcess. Skipping..."
-                        }
-                    )
+                    "Config] Database error when getting NextVersionToProcess. Skipping..."
```
jillxuu, line 70:
```suggestion
                    "[Config] Database error when getting NextVersionToProcess. Skipping..."
```

jillxuu (approved):

python/utils/config.py
larry-aptos: good catch; done.

rtso (approved):

PR #195, opened 2023-11-19
[Rust] Create metric helper service based on post processor

bowenyang007:
Post processor has a few issues as is. It doesn't capture enough context and it is too unreliable as is. This new implementation is much simpler and therefore more reliable. It queries transaction version and transaction timestamp from several sources. Currently it's just the PFN. 

## Next
* Need to add hasura APIs but currently those APIs don't have transaction timestamps so we'll need to add that.
* Use it in new dashboards
* Deprecate post_processor


rust/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apt-get update && apt-get install -y cmake curl clang git pkg-config libssl-
 RUN cargo build --locked --release -p processor
 RUN cp target/release/processor /usr/local/bin
 RUN cargo build --locked --release -p post-processor
+RUN cargo build --locked --release -p metrics-helper
```
rustielin, line 15: you will also need to copy this into the final image like on line 31.

`COPY --from=builder /usr/local/bin/metrics-helper /usr/local/bin`

Also nit maybe `metrics-helper` --> ` indexer-metrics` or something more descriptive?

rust/metrics-helper/src/metrics.rs
```diff
@@ -0,0 +1,46 @@
...
+use once_cell::sync::Lazy;
+use prometheus::{
+    register_gauge, register_int_counter_vec, register_int_gauge, register_int_gauge_vec, Gauge,
+    IntCounterVec, IntGauge, IntGaugeVec,
+};
+
+/// Task failure count.
+pub static TASK_FAILURE_COUNT: Lazy<IntCounterVec> = Lazy::new(|| {
+    register_int_counter_vec!(
+        "indexer_processors_post_processing_task_failure_count",
```
rustielin, line 13: nit: `indexer_metrics_post_processing_task_failure_count` ?

rust/metrics-helper/src/metrics.rs
```diff
@@ -0,0 +1,46 @@
...
+        "Task failure count.",
+        &["task_name"],
+    )
+    .unwrap()
+});
+
+pub static HASURA_API_LATEST_VERSION: Lazy<IntGaugeVec> = Lazy::new(|| {
+    register_int_gauge_vec!("hasura_api_latest_version", "Processor latest version", &[
```
rustielin, line 21: nit: I'd give most of these metrics a relevant prefix letting the user know that these are indeed coming from the indexer_metrics / post-processors. It will be difficult to discern on the TSDB side whether this latency is coming from hasura itself or something else like this service. see https://prometheus.io/docs/practices/naming/

rust/metrics-helper/src/metrics.rs
```diff
@@ -0,0 +1,46 @@
...
+        "processor_name"
+    ],)
+    .unwrap()
+});
+
+pub static HASURA_API_LATEST_VERSION_TIMESTAMP: Lazy<IntGaugeVec> = Lazy::new(|| {
+    register_int_gauge_vec!(
+        "hasura_api_latest_version_timestamp",
```
rustielin, line 29: nit: note that this is in seconds

rust/metrics-helper/src/metrics.rs
```diff
@@ -0,0 +1,46 @@
...
+        "Processor latest timestamp (unix timestamp)",
+        &["processor_name"],
+    )
+    .unwrap()
+});
+
+pub static PFN_LEDGER_VERSION: Lazy<IntGauge> = Lazy::new(|| {
+    register_int_gauge!("pfn_ledger_version_from_indexer", "Ledger latest version",).unwrap()
```
rustielin, line 37: nit: same as above. prefix this with indexer to make it easier to find and figure out where it's measured from

rustielin (changes requested):

rust/Dockerfile
bowenyang007: lol i see i see good point

PR #198, opened 2023-12-01
Increase reconnection times

rtso:
## Description 
1. Increase max reconnection retries
2. Increase time to wait for transactions from GRPC before panic
3. Add a wait in between reconnections

### DIFF
--- a/rust/processor/src/worker.rs
+++ b/rust/processor/src/worker.rs
@@ -55,13 +55,10 @@ const GRPC_CONNECTION_ID: &str = "x-aptos-connection-id";
 const BUFFER_SIZE: usize = 50;
 // 20MB
 const MAX_RESPONSE_SIZE: usize = 1024 * 1024 * 20;
-// We will try to reconnect to GRPC once every X seconds if we get disconnected, before crashing
-// We define short connection issue as < 10 seconds so adding a bit of a buffer here
-const MIN_SEC_BETWEEN_GRPC_RECONNECTS: u64 = 15;
 // We will try to reconnect to GRPC 5 times in case upstream connection is being updated
-const RECONNECTION_MAX_RETRIES: u64 = 5;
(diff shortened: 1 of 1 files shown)


rust/processor/src/worker.rs
```diff
@@ -1126,23 +1122,20 @@ pub async fn create_fetcher_loop(
...
-                        service_type = PROCESSOR_SERVICE_TYPE,
-                        stream_address = indexer_grpc_data_service_address.to_string(),
-                        seconds_since_last_retry = elapsed,
-                        "[Parser] Recently reconnected. Will not retry.",
-                    );
-                    panic!("[Parser] Recently reconnected. Will not retry.")
-                }
+
+            // Sleep for 100ms between reconnect tries
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
CapCap, line 1127: I would recommend an exponential retry strategy - ala https://docs.rs/tokio-retry/latest/tokio_retry/ - so that we try reconnecting the first time quickly, to minimize latency, and fall off if we realize the end service might be down, restarting, etc

rtso: Currently the exponential backoff is managed in the pods and it's obscuring some of the issues we're trying to test/reproduce with indexer GRPC. This PR moves the reconnect logic to be managed within code. I'll add a todo to add exponential backoff.

CapCap: ah, pods didn't come back up immediately?

rtso: No I've seen it taking > 15s for a pod to restart after crash loop backoff. Removing the backoff fixed the problem we were seeing, so I'll keep this and revisit backoff later.

bowenyang007: On the infra level there are also resource contention problems. E.g. there are too many pods and we hit a quota or something.

PR #212, opened 2023-12-07
[rust] Optimize look up fungible token query

bowenyang007:


rust/processor/src/models/token_v2_models/v2_token_datas.rs
```diff
@@ -260,6 +263,13 @@ impl TokenDataV2 {
             match Self::get_by_token_data_id(conn, address).await {
                 Ok(_) => return true,
                 Err(_) => {
+                    // TODO: this'll unblock very short term but we should clean this up
+                    if FungibleAssetMetadataModel::get_by_asset_type(conn, address)
```
rtso, line 267: making sure i understand this: 

1. if it can't find it in the token table, check fungible asset metadata table
2. if it can find it in fungible asset metadata table, then it's not a token and return false
3. if it can't find it in fungible asset metadata table, loop again

rtso (approved):

rust/processor/src/models/token_v2_models/v2_token_datas.rs
rtso: This is a temporary fix to unblock. The fast follow and better way to do it is to create a table mapping of address -> type (coin, token, fungible asset) but that will require backfill which takes too long.

PR #215, opened 2023-12-08
Add objects processor

rtso:
## Description 
* Move objects and current_objects parsing into its own objects processor
* Add 2 new columns (is_token, is_fungible_asset) which will be used in token_v2_processor and fungible_asset_processor to check if an object is token or fungible asset

## Backfill
Given this splits out the objects processor we need to start the new processor from when objects were created. No need to backfill default_processor though. 
* Testnet: 469526272
* Mainnet: 144201952

## Testing 
object is token 
[Testnet version 528464455](https://explorer.aptoslabs.com/txn/528464455/changes?network=testnet)
Screenshot 2023-12-08 at 11 52 35 AM
Screenshot 2023-12-08 at 11 52 28 AM

object is fungible asset 
[Testnet version 802980551](https://explorer.aptoslabs.com/txn/802980551/changes?network=testnet) 
Screenshot 2023-12-08 at 11 54 07 AM
Screenshot 2023-12-08 at 11 53 56 AM

object is both token and fungible asset
[Testnet version 617679498](https://explorer.aptoslabs.com/txn/617679498/changes?network=testnet)
Screenshot 2023-12-08 at 11 55 29 AM
Screenshot 2023-12-08 at 11 55 18 AM

PR #216, opened 2023-12-08
Update fungible asset and token processor queries to use objects table

rtso:
## Description 
* Update the fungible asset and token v2 processors to query for whether an object is token and/or fungible asset using the current_objects table 
* Also use ObjectAggregatedDataMapping everywhere instead of TokenV2DataMapping or FungibleAssetMetadataMapping

## Breaking change
* Dependent on this PR: https://github.com/aptos-labs/aptos-indexer-processors/pull/215
* If you're running fungible asset processor or token v2 processor, you must also run objects processor to lookup object data
* Testnet: Backfill objects processor
```
server_config:
  processor_config:
    type: "objects_processor"
  starting_version: 469526272
```
* Mainnet: Backfill objects processor
```
server_config:
  processor_config:
    type: "objects_processor"
  starting_version: 144201952
```

### Before
Run token v2 processor starting_version = 803795000 (when the polling ended)
Processor is slow/stuck > 10 min. 

### After 
Run objects processor starting_version = 802980551 (when the question was created) 
Run token v2 processor starting_version = 803795000 (when the polling ended)

Duration to process version 803795008: 0.061555459s
```
{"timestamp":"2023-12-08T17:33:19.184016Z","level":"INFO","fields":{"message":"[Parser] Parsing time of one batch of transactions","processor_name":"token_v2_processor","service_type":"processor","start_version":803795008,"end_version":803795008,"start_txn_timestamp_iso":"2023-12-07T05:26:38.527385000Z","end_txn_timestamp_iso":"2023-12-07T05:26:38.527385000Z","size_in_bytes":281546,"duration_in_secs":0.061555459,"tps":0.0,"bytes_per_sec":4251470.2700600065},"filename":"processor/src/worker.rs","line_number":501,"threadName":"tokio-runtime-worker","threadId":"ThreadId(11)"}
```

### DIFF
--- a/rust/processor/src/models/fungible_asset_models/v2_fungible_metadata.rs
+++ b/rust/processor/src/models/fungible_asset_models/v2_fungible_metadata.rs
@@ -9,7 +9,8 @@ use super::v2_fungible_asset_utils::FungibleAssetMetadata;
 use crate::{
     models::{
         coin_models::coin_utils::{CoinInfoType, CoinResource},
-        object_models::v2_object_utils::{Object, ObjectAggregatedDataMapping},
+        object_models::v2_object_utils::ObjectAggregatedDataMapping,
+        object_models::v2_objects::Object,
         token_v2_models::v2_token_utils::TokenStandard,
     },
     schema::fungible_asset_metadata,
--- a/rust/processor/src/models/fungible_asset_models/v2_fungible_metadata.rs
+++ b/rust/processor/src/models/fungible_asset_models/v2_fungible_metadata.rs
@@ -9,8 +9,7 @@ use super::v2_fungible_asset_utils::FungibleAssetMetadata;
 use crate::{
     models::{
         coin_models::coin_utils::{CoinInfoType, CoinResource},
-        object_models::v2_object_utils::ObjectAggregatedDataMapping,
-        object_models::v2_objects::Object,
+        object_models::{v2_object_utils::ObjectAggregatedDataMapping, v2_objects::Object},
         token_v2_models::v2_token_utils::TokenStandard,
     },
     schema::fungible_asset_metadata,
@@ -58,17 +57,17 @@ impl FungibleAssetMetadataModel {
         write_resource: &WriteResource,
         txn_version: i64,
         txn_timestamp: chrono::NaiveDateTime,
-        fungible_asset_metadata: &ObjectAggregatedDataMapping,
+        object_metadatas: &ObjectAggregatedDataMapping,
     ) -> anyhow::Result<Option<Self>> {
         if let Some(inner) =
             &FungibleAssetMetadata::from_write_resource(write_resource, txn_version)?
         {
             // the new coin type
             let asset_type = standardize_address(&write_resource.address.to_string());
-            if let Some(metadata) = fungible_asset_metadata.get(&asset_type) {
-                let object = &metadata.object.object_core;
+            if let Some(object_metadata) = object_metadatas.get(&asset_type) {
+                let object = &object_metadata.object.object_core;
                 // Do not write here if asset is fungible token
-                if metadata.token.is_some() {
+                if object_metadata.token.is_some() {
                     return Ok(None);
                 }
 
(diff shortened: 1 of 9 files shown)


bowenyang007 (changes requested):
So conceptually now that this is in a different processor altogether you can't just assume that it's not something just b/c you don't find it in the table. You likely have to wait forever to ensure that the data is not wrong. If object processor falls behind for example b/c of a bug, token v2 processor should not be able to advance either.

rust/processor/src/models/object_models/v2_objects.rs
```diff
@@ -278,14 +269,20 @@ impl Object {
...
-                    })
+                    };
                 },
-                Err(_) => {
+                Err(e) => {
                     std::thread::sleep(std::time::Duration::from_millis(QUERY_RETRY_DELAY_MS));
+                    warn!(
+                        transaction_version,
+                        error = ?e,
+                        "Failed to get object from current_objects table for object_address: {}, retrying in {} ms. If the object doesn't exist you should probably backfill db.",
```
larry-aptos, line 279: if db doesn't setup correctly, retry forever may not help; consider adding a upper bound for retrying.

larry-aptos (approved):

rust/processor/src/models/object_models/v2_objects.rs
rtso: what do you mean by not setup correctly?

larry-aptos: db requires backfilling

rtso: that's a good point. the new columns require backfill or else the token v2 and fungible asset processors will not advance. i guess the question is whether we should 
1. retry forever. this is a silent error and the processor will be stuck. 
2. return an error after a couple retries to signal users they need to backfill 

i think option 2 is better so users know they need to take an action to advance the processors.

rust/processor/src/models/fungible_asset_models/v2_fungible_metadata.rs
```diff
@@ -135,55 +132,19 @@ impl FungibleAssetMetadataModel {
...
-            retried += 1;
-            match Self::get_by_asset_type(conn, address).await {
-                Ok(_) => return true,
-                Err(_) => {
-                    std::thread::sleep(std::time::Duration::from_millis(QUERY_RETRY_DELAY_MS));
-                },
-            }
+            // Look up in the db. The object must exist in current_objects table for this processor to proceed
+            let object = Object::get_object(conn, address, txn_version).await;
+            object.is_fungible_asset && !object.is_token
```
bowenyang007, line 147: These should be nullable so...my guess is that you want to keep the old code for now in case these are nullable, but add an warning that they should backfill the object processor.

rust/processor/src/models/object_models/v2_objects.rs
```diff
@@ -258,17 +249,18 @@ impl Object {
         }
     }
 
-    /// This is actually not great because object owner can change. The best we can do now though
-    async fn get_object_owner(
+    /// This is actually not great because object owner can change. The best we can do now though.
+    /// This will loop forever until we get the object from the db
+    pub async fn get_object(
```
bowenyang007, line 254: get_current_object

rust/processor/src/models/object_models/v2_objects.rs
```diff
@@ -278,14 +270,21 @@ impl Object {
...
                 },
-                Err(_) => {
+                Err(e) => {
                     std::thread::sleep(std::time::Duration::from_millis(QUERY_RETRY_DELAY_MS));
+                    warn!(
+                        transaction_version,
+                        error = ?e,
+                        "Failed to get object from current_objects table for object_address: {}, retrying in {} ms. ",
+                        object_address,
+                        QUERY_RETRY_DELAY_MS,
```
bowenyang007, line 282: Structure object_address and the delay ms as well.

bowenyang007 (approved):

PR #228, opened 2024-01-05
Fix issue where burn is missing from token activities

bowenyang007:


rust/processor/src/models/token_v2_models/v2_token_activities.rs
```diff
@@ -208,6 +211,43 @@ impl TokenActivityV2 {
                     is_fungible_v2: Some(false),
                     transaction_timestamp: txn_timestamp,
                 }));
+            } else {
+                // This should only happen in the case where we have a burn event where the token is gone
+                // and the previous instance of the token wasn't in the batch. We need to look up in the db
```
rtso, line 216: This isn't definitively the last owner, right? Token could have changed ownership in a transaction processed in a different batch.  Can we add a comment that this lookup in db is a best effort guess at the previous owner?

rtso (approved):
LGTM!

rust/processor/src/models/token_v2_models/v2_token_activities.rs
bowenyang007: Ah shit...we need lookups. This limitation applies to every query then.

rtso: Yeah this is the limitation with every lookup we do for past data. If we want the definitive owner, we need to build a dependency graph between all the processors, ie this processor can't resume until X processor has processed Y version.

PR #239, opened 2024-01-17
[DRAFT][DO_NOT_LAND] create all NFT marketplace tables and index nft_marketplace_activities 

0xjinn:
"DO NOT LAND"

### DIFF
--- a/rust/processor/src/models/nft_marketplace_models/nft_marketplace_utils.rs
+++ b/rust/processor/src/models/nft_marketplace_models/nft_marketplace_utils.rs
@@ -1,0 +1,186 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+// This is required because a diesel macro makes clippy sad
+#![allow(clippy::extra_unused_lifetimes)]
+
+use std::str::FromStr;
+
+use bigdecimal::BigDecimal;
+use serde::{Deserialize, Serialize};
+
+use crate::{utils::util::standardize_address, models::token_v2_models::v2_token_utils::TokenStandard};
+
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MarketplaceTokenMetadata {
+    pub collection_id: String,
+    pub token_data_id: String,
+    pub creator_address: String,
+    pub collection_name: String,
+    pub token_name: String,
+    pub property_version: Option<BigDecimal>,
+    pub token_standard: String,
+}
+
+impl MarketplaceTokenMetadata {
+    pub fn from_event_data(
+        event_data: serde_json::Value,
+    ) -> anyhow::Result<Option<Self>> {
+        let token_metadata = event_data["token_metadata"].clone();
+        if token_metadata.is_null() {
+            return Ok(None);
+        }
+
+        // V2 token will 
+        let is_token_v2 = token_metadata["token"]["vec"].as_array().unwrap().len() > 0;
+        let creator_address = standardize_address(token_metadata["creator_address"].as_str().unwrap());
+        let property_version = if token_metadata["property_version"]["vec"].as_array().unwrap().len() > 0 {
+            Some(BigDecimal::from_str(token_metadata["property_version"]["vec"][0].as_str().unwrap()).unwrap())
+        } else {
+            None
+        };
+
+        // V2 Token
+        if (is_token_v2) {
+            let collection_id = standardize_address(token_metadata["collection"]["vec"][0]["inner"].as_str().unwrap());
+            let token_data_id = standardize_address(token_metadata["token"]["vec"][0]["inner"].as_str().unwrap());
+            let collection_name = token_metadata["collection_name"].as_str().unwrap().to_string();
+            let token_name = token_metadata["token_name"].as_str().unwrap().to_string();
+            let token_standard = TokenStandard::V2.to_string();
+
+            Ok(Some(Self {
+                collection_id,
+                token_data_id,
+                creator_address,
+                collection_name,
+                token_name,
+                property_version,
+                token_standard,
+            }))
+        } else {
+            // V1 Token
+            let token_data_id_type = TokenDataIdType::new(
+                creator_address.clone(),
+                token_metadata["collection_name"].as_str().unwrap().to_string(),
+                token_metadata["token_name"].as_str().unwrap().to_string(),
+            );
+
+            let collection_id = token_data_id_type.to_collection_id();
+            let token_data_id = token_data_id_type.to_hash();
+            let collection_name = token_metadata["collection_name"].as_str().unwrap().to_string();
+            let token_name = token_metadata["token_name"].as_str().unwrap().to_string();
+            let token_standard = TokenStandard::V1.to_string();
+
+            Ok(Some(Self {
+                collection_id,
+                token_data_id,
+                creator_address,
+                collection_name,
+                token_name,
+                property_version,
+                token_standard,
+            }))
+        }
+    }
+}
+
+pub struct MarketplaceCollectionMetadata {
+    pub collection_id: String,
+    pub creator_address: String,
+    pub collection_name: String,
+    pub token_standard: String,
+}
+
+impl MarketplaceCollectionMetadata {
+    pub fn from_event_data(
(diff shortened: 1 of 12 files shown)


rust/processor/src/models/nft_marketplace_models/nft_marketplace_utils.rs
```diff
@@ -0,0 +1,251 @@
...
+    pub fn from_event_data(
+        event_data: &serde_json::Value,
+    ) -> anyhow::Result<Option<Self>> {
+        let token_metadata = event_data["token_metadata"].clone();
+        if token_metadata.is_null() {
+            return Ok(None);
+        }
+
+        // V2 token will 
+        let is_token_v2 = token_metadata["token"]["vec"].as_array().unwrap().len() > 0;
```
bowenyang007, line 36: This doesn't seem right.

rust/processor/src/models/ans_models/ans_utils.rs
```diff
@@ -19,12 +19,12 @@ use serde::{Deserialize, Serialize};
 pub const DOMAIN_LENGTH: usize = 64;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-struct OptionalString {
+pub struct OptionalString {
```
rtso, line 22: this looks like it's generic type. let's move it to some generic utils file

rust/processor/src/processors/marketplaces_processor.rs
```diff
@@ -0,0 +1,1048 @@
...
+                    token_metadatas.insert(token_metadata.get_token_address().unwrap(), token_metadata.clone());
+                }
+                let collection_metadata =
+                    CollectionMetadata::from_event(event, txn_version).unwrap();
+                if let Some(collection_metadata) = collection_metadata {
+                    collection_metadatas.insert(
+                        collection_metadata.get_collection_address().unwrap(),
+                        collection_metadata,
+                    );
+                }
```
rtso, line 435: I realize the way the Python processor is written is not the "correct" way to write processors.. The Python processor tries to read the individual data fields from the event. It's less code but can be error-prone (eg. if the smart contract changes, it will break assumptions made in this indexer and we won't know about it). 

The right way would be to create a Rust representation of the events data model that matches exactly the format of the events we expect. The ANS processor is a good example of this. ([link](https://github.com/aptos-labs/aptos-indexer-processors/blob/a9e42dc9d80a69e9d02cd5f19c39e0f3f5cfbeb0/rust/processor/src/models/ans_models/ans_lookup_v2.rs#L273))
For each event type, we have a struct for it. It's more verbose but less error-prone.

rtso (changes requested):

rust/processor/src/processors/marketplaces_processor.rs
0xjinn: Yep, I have them on line 438 - 506. Each data models (activities, offer, listing, etc) has a `from_event` method which extra and build the corresponding model.

Though I could get the `token_metadata` and `collection_metadata` from line 438 - 506 as well, but I think having it separately (line 424 - 435) is more cleaner. Any thoughts?

rtso: It's less error-prone to have `token_metadata` and `collection_metadata` be read from the `from_event`.

Same thing with `price` and `activity_offer_or_listing_id`

rust/processor/src/processors/marketplaces_processor.rs
```diff
@@ -0,0 +1,1048 @@
...
+
+                // Get the acitivty, listing, token_offer, and auction from event.
```
rtso, line 437: Do something like [this](https://github.com/aptos-labs/aptos-indexer-processors/blob/b2325e585d68c5e2089527504943445906b4711d/rust/processor/src/models/token_v2_models/v2_token_utils.rs#L370), so that if it matches a condition, it won't run the next statements.

rust/processor/src/processors/marketplaces_processor.rs
```diff
@@ -0,0 +1,1048 @@
...
+            for (wsc_index, wsc) in txn.info.as_ref().unwrap().changes.iter().enumerate() {
+                match wsc.change.as_ref().unwrap() {
+                    Change::WriteResource(write_resource) => {
+                        let move_resource_address = standardize_address(&write_resource.address);
+                        let move_resource_type = write_resource.type_str.clone();
+                        let move_resource_type_address =
+                            write_resource.r#type.as_ref().unwrap().address.clone();
+                        if !move_resource_type_address.eq(marketplace_contract_address) {
+                            continue;
+                        }
```
rtso, line 536: We can also wrap this in a helper function. [Example](https://github.com/aptos-labs/aptos-indexer-processors/blob/b2325e585d68c5e2089527504943445906b4711d/rust/processor/src/models/token_v2_models/v2_token_utils.rs#L357)

rust/processor/src/models/ans_models/ans_utils.rs
bowenyang007: sounds like a good idea. let's do this separately.

rust/processor/src/models/nft_marketplace_models/marketplace_utils.rs
```diff
@@ -0,0 +1,795 @@
...
+        events.insert(Self::TOKEN_OFFER_PLACED_EVENT);
+        events.insert(Self::TOKEN_OFFER_CANCELED_EVENT);
+        events.insert(Self::TOKEN_OFFER_FILLED_EVENT);
+        events.insert(Self::AUCTION_BID_EVENT);
+        events
+    }
+
+    pub fn from_event(data_type: &str, data: &str, txn_version: i64) -> Result<Option<Self>> {
+        match data_type {
+            x if x.ends_with(Self::LISTING_FILLED_EVENT) => {
```
bowenyang007, line 540: you can't do this b/c what if it's a different address.

rust/processor/src/models/nft_marketplace_models/marketplace_utils.rs
```diff
@@ -0,0 +1,795 @@
...
+        .context(format!(
+            "MarketplaceEvent with version {} failed! failed to parse type {}, data {:?}",
+            txn_version, data_type, data
+        ))
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ListingFilledEvent {
+    pub r#type: String,
```
bowenyang007, line 578: actually if you don't need type then you don't need this.

rust/processor/src/processors/marketplaces_processor.rs
```diff
@@ -0,0 +1,1022 @@
...
+                    collection_offer_filled_metadatas.insert(
+                        collection_offer_filled_metadata.collection_offer_id.clone(),
+                        collection_offer_filled_metadata,
+                    );
+
+                    nft_collection_offers.push(collection_offer);
+                }
+            }
+
+            let mut object_metadatas: HashMap<String, ObjectCore> = HashMap::new();
```
bowenyang007, line 495: are you removing code below? this seems like it should be in the model files.

rust/processor/src/models/nft_marketplace_models/marketplace_utils.rs
0xjinn: Yeah, so the idea i have is to gate the whole marketplace processor at the beginning. Only the contract_address we have on file will pass the gating. Though this is still TODO part - I still have to all all other marketplaces.

rust/processor/src/processors/marketplaces_processor.rs
0xjinn: Not sure what is the best way to handle the code below, these are HashMap and being used to store data we aggregate from wsc Loop 2 below, then we restruct them in Loop 3. 

One idea I have, is to define a model that has all `Option` field, and we construct the model/object as we go on Loop 2?

cc: Wonder if @rtso has any good idea to simplify this?

rust/processor/src/models/nft_marketplace_models/marketplace_utils.rs
rtso: We can early return and gate it here too

PR #242, opened 2024-01-18
CRDB compatible migrations

CapCap:


rust/processor/src/schema.rs
```diff
@@ -13,14 +13,14 @@ diesel::table! {
     ans_lookup (transaction_version, write_set_change_index) {
         transaction_version -> Int8,
         write_set_change_index -> Int8,
-        #[max_length = 64]
+        #[max_length = 0]
```
jillxuu, line 16: oh you might need to upgrade your local diesel

PR #245, opened 2024-01-19
Remove panic for unknown signature types

rtso:
## Testing 
Run user_transaction_processor
Screenshot 2024-01-19 at 6 46 35 PM


bowenyang007 (approved):
Can we add a prom metric. Also this should fix ZKID as well

CapCap (approved):

PR #255, opened 2024-01-31
[processor][DO NOT MERGE] Randomnet special cut. 

larry-aptos:
Experimental build for randomnet

DO NOT MERGE

### DIFF
--- a/rust/processor/src/models/user_transactions_models/signatures.rs
+++ b/rust/processor/src/models/user_transactions_models/signatures.rs
@@ -10,7 +10,8 @@ use crate::{
 use anyhow::{Context, Result};
 use aptos_protos::transaction::v1::{
     account_signature::Signature as AccountSignatureEnum,
-    any_signature::Type as AnySignatureTypeEnumPb, signature::Signature as SignatureEnum,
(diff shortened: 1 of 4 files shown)


rust/processor/src/models/user_transactions_models/signatures.rs
```diff
@@ -353,6 +354,12 @@ impl Signature {
                 "".to_string()
             },
         };
+        let signature_bytes = match signature.signature.as_ref().unwrap() {
```
bowenyang007, line 357: This isn't mergeable on main right? b/c we don't want to unwrap here.

larry-aptos: nope this is a hack; good call out and thanks Rustie marking this as draft

bowenyang007: Also why are we doing this?

larry-aptos: existing processor is broken.

context: randomnet aptos-core cut is from latest main with a lot of cherrypicks. To use that fullnode, we are required to bump protobuf versions; otherwise, processor cannot recognize the new transaction type.

PR #256, opened 2024-02-01
CDB Compatible migrations

CapCap:


rust/processor/migrations/2023-09-01-231248_events_v2/down.sql
```diff
@@ -1,8 +1 @@
 -- This file should undo anything in `up.sql`
```
just-in-chang, line 1: `down.sql` not needed?

rust/processor/migrations/2023-11-09-234725_delegator_balances_2/down.sql
```diff
@@ -0,0 +1,6 @@
+-- This file should undo anything in `up.sql`
```
just-in-chang, line 1: Not dropping `delegator_balances` table here?

just-in-chang (approved):
Things mostly line up, could use another set of eyes here tho 🙂

rust/processor/migrations/2023-09-01-231248_events_v2/down.sql
CapCap: nah- many of these shouldn't have had a `down` to begin with

rust/processor/migrations/2023-11-09-234725_delegator_balances_2/down.sql
CapCap: yeah nah, no worries; we wont

rust/processor/migrations/2023-09-01-231248_events_v2/down.sql
rtso: when should it not have a `down`?

rust/processor/migrations/2023-09-01-231247_events_v2_1/up.sql
```diff
@@ -0,0 +1,6 @@
+-- Your SQL goes here
+
+ALTER TABLE events DROP CONSTRAINT IF EXISTS fk_transaction_versions;
+
+ALTER TABLE events
+ALTER COLUMN "event_index" SET NOT NULL;
```
rtso, line 6: where did this line come from?

CapCap: it was necessary to get the migrations to run as  CDB doesn't allow index on full fields

rtso (approved):

PR #263, opened 2024-02-08
Gap detector

rtso:
## Description 
1. Create a gap detector in a separate tokio task
2. Enables parallel parsing and DB inserts while ensuring data completeness

## Testing
Run locally

### DIFF
--- a/rust/processor/src/gap_detector.rs
+++ b/rust/processor/src/gap_detector.rs
@@ -1,96 +1,72 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{
-    models::processor_status::ProcessorStatus,
-    processors::ProcessedVersions,
-    schema::processor_status,
-    utils::{
-        database::{execute_with_better_error, PgDbPool},
-        util::parse_timestamp,
-    },
-};
+use crate::processors::ProcessingResult;
 use ahash::AHashMap;
-use diesel::{upsert::excluded, ExpressionMethods};
 
 // Number of batches processed before gap detected
 pub const GAP_DETECTION_BATCH_COUNT: u64 = 50;
-// Number of batches to process before updating processor status
-const PROCESSOR_STATUS_UPDATE_BATCH_COUNT: u64 = 10;
 
 pub struct GapDetector {
-    processor_name: String,
-    db_pool: PgDbPool,
     starting_version: u64,
-    seen_versions: AHashMap<u64, ProcessedVersions>,
-    maybe_prev_batch: Option<ProcessedVersions>,
-    num_batches_processed_without_gap: u64,
-    num_batches_processed_with_gap: u64,
+    seen_versions: AHashMap<u64, ProcessingResult>,
+    maybe_prev_batch: Option<ProcessingResult>,
 }
 
 #[derive(Debug, PartialEq)]
 pub enum GapDetectorResult {
-    GapDetected { gap_start_version: u64 },
-    NoGapDetected(),
+    GapDetected {
+        gap_start_version: u64,
+    },
+    NoGapDetected {
+        last_success_batch: Option<ProcessingResult>,
+        num_batches_processed_without_gap: u64,
+    },
 }
 
 impl GapDetector {
-    pub fn new(processor_name: String, db_pool: PgDbPool, starting_version: u64) -> Self {
+    pub fn new(starting_version: u64) -> Self {
         Self {
-            processor_name,
-            db_pool,
             starting_version,
             seen_versions: AHashMap::new(),
             maybe_prev_batch: None,
-            num_batches_processed_without_gap: 0,
-            num_batches_processed_with_gap: 0,
         }
     }
 
-    pub async fn process_versions(
+    pub fn process_versions(
         &mut self,
-        result: ProcessedVersions,
+        result: ProcessingResult,
     ) -> anyhow::Result<GapDetectorResult> {
         // Check for gaps
         let gap_detected = self.detect_gap(result.clone());
 
+        let mut num_batches_processed_without_gap = 0;
         if gap_detected {
             self.seen_versions.insert(result.start_version, result);
-            self.num_batches_processed_with_gap += 1;
             tracing::debug!("Gap detected");
+
+            // It detects a gap if it processed GAP_DETECTION_BATCH_COUNT batches without finding (maybe_prev_batch.end_version + 1)
+            if self.seen_versions.len() as u64 >= GAP_DETECTION_BATCH_COUNT {
+                let gap_start_version = if let Some(prev_batch) = &self.maybe_prev_batch {
+                    prev_batch.end_version + 1
+                } else {
+                    self.starting_version
+                };
+                return Ok(GapDetectorResult::GapDetected { gap_start_version });
+            }
         } else {
             // If no gap is detected, find the latest processed batch without gaps
-            self.update_prev_batch(result);
-            self.num_batches_processed_with_gap = 0;
+            num_batches_processed_without_gap = self.update_prev_batch(result);
             tracing::debug!("No gap detected");
         }
 
-        // If there's a gap detected, panic.
-        // A gap is detected if we've processed GAP_DETECTION_BATCH_COUNT batches without seeing the batch that comes after prev_batch
-        if self.num_batches_processed_with_gap >= GAP_DETECTION_BATCH_COUNT {
-            let gap_start_version = if let Some(prev_batch) = &self.maybe_prev_batch {
-                prev_batch.end_version + 1
-            } else {
-                self.starting_version
-            };
-            return Ok(GapDetectorResult::GapDetected { gap_start_version });
-        }
-
-        // Check if need to update processor status
-        if self.num_batches_processed_without_gap >= PROCESSOR_STATUS_UPDATE_BATCH_COUNT {
-            let prev_batch = self.maybe_prev_batch.clone().unwrap();
-            self.update_processor_status(
-                prev_batch.end_version,
-                prev_batch.last_transaction_timstamp,
-            )
-            .await;
-            self.num_batches_processed_without_gap = 0;
-        }
-
-        Ok(GapDetectorResult::NoGapDetected())
+        Ok(GapDetectorResult::NoGapDetected {
+            last_success_batch: self.maybe_prev_batch.clone(),
+            num_batches_processed_without_gap,
+        })
     }
 
-    fn detect_gap(&self, result: ProcessedVersions) -> bool {
+    fn detect_gap(&self, result: ProcessingResult) -> bool {
         if let Some(prev_batch) = &self.maybe_prev_batch {
             prev_batch.end_version + 1 != result.start_version
         } else {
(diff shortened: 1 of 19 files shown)


rust/processor/src/worker.rs
```diff
@@ -35,9 +43,10 @@ use aptos_protos::{
     indexer::v1::{raw_data_client::RawDataClient, GetTransactionsRequest, TransactionsResponse},
     transaction::v1::Transaction,
 };
+use diesel::{upsert::excluded, ExpressionMethods};
 use futures::StreamExt;
 use prost::Message;
-use std::{sync::Arc, time::Duration};
+use std::{collections::HashMap, sync::Arc, time::Duration};
```
CapCap, line 49: can use `ahash::AHashMap` now

rust/processor/src/worker.rs
```diff
@@ -60,6 +69,11 @@ const MAX_RESPONSE_SIZE: usize = 1024 * 1024 * 20;
 const RECONNECTION_MAX_RETRIES: u64 = 100;
 // Consumer thread will wait X seconds before panicking if it doesn't receive any data
 const CONSUMER_THREAD_TIMEOUT_IN_SECS: u64 = 60 * 5;
+// Number of batches processed before gap detected
+const GAP_DETECTION_BATCH_COUNT: u64 = 50;
```
CapCap, line 73: this is potentially 50,000 transactions, right? Maybe it'd make sense to have a time based component to the check too

rust/processor/src/worker.rs
```diff
@@ -1182,3 +1179,152 @@ pub async fn create_fetcher_loop(
...
+        for i in 0..GAP_DETECTION_BATCH_COUNT {
+            let result = ProcessingResult {
+                start_version: i * 100,
+                end_version: i * 100 + 99,
+                processing_duration_in_secs: 0.0,
+                db_insertion_duration_in_secs: 0.0,
+            };
+            tx.send(result).await.unwrap();
+        }
+        gap_detector_task.await;
```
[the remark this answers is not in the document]
CapCap: you could try closing the channel?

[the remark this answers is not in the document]
CapCap: If you make a `gap_detector.rs` and put a class in there that handles gap detection, testing it in a non-async context becomes much easier

rust/processor/src/worker.rs
```diff
@@ -1182,3 +1188,184 @@ pub async fn create_fetcher_loop(
...
+    // Keep track of the latest batch processed without gaps
+    let mut maybe_prev_batch: Option<ProcessedVersions> = None;
+    // Counter of how many batches have been processed with no gaps
+    let mut num_batches_processed_without_gap = 0;
+    // Counter of how many batches have been processed with a gap from prev_end
+    let mut num_batches_processed_with_gap = 0;
+
+    loop {
+        match receiver.recv().await {
+            Some(result) => {
```
CapCap, line 1212: I think you could flatten this logic, instead of being so indented- i.e
```
  let result = match receiver.recv().await {
            Some(result) => result,
            None => continue,
}
```

rust/processor/src/worker.rs
```diff
@@ -1182,3 +1188,184 @@ pub async fn create_fetcher_loop(
...
+                        processor_status::last_success_version
+                            .eq(excluded(processor_status::last_success_version)),
+                        processor_status::last_updated.eq(excluded(processor_status::last_updated)),
+                        processor_status::last_transaction_timestamp
+                            .eq(excluded(processor_status::last_transaction_timestamp)),
+                    )),
+                Some(" WHERE processor_status.last_success_version <= EXCLUDED.last_success_version "),
+            )
+            .await
+            .expect("[Parser] Error updating processor status");
```
CapCap, line 1319: can put all of this into it's own stand alone method for clarity

rust/processor/src/worker.rs
rtso: what if it's processing large transactions that takes long time to insert to DB? it might panic and crash loop forever. having the detection based on # of transactions means if there's a gap it will eventually detect it and panic.

CapCap: we could do both- we just want progress, but 50k transactions is a _very_ long time; and it shouldn't take more than 30s for any gap, right? I think it's a question of how we can guarantee we panic in the lowest time possible

rtso: We could test with panicking after 30s without any advancement of the gap counter

rtso: Actually if we had a 30s counter, this would panic in the case of processing Graffio txn's. For example, we see it happening whenever we backfill a mainnet processor. When it hits the Graffio txn's, it becomes slower than 30s, but the processor is still working. We don't want it to panic or it'd be in a crash loop forever.

rust/processor/src/processors/mod.rs
```diff
@@ -62,6 +62,14 @@ pub struct ProcessingResult {
     pub db_insertion_duration_in_secs: f64,
 }
 
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+
+pub struct ProcessedVersions {
+    pub start_version: StartVersion,
+    pub end_version: EndVersion,
+    pub last_transaction_timstamp: Option<aptos_protos::util::timestamp::Timestamp>,
+}
```
bowenyang007, line 71: seems like you can reuse https://github.com/aptos-labs/aptos-indexer-processors/blob/main/rust/processor/src/processors/mod.rs#L58-L64.

rust/processor/src/utils/database.rs
```diff
@@ -146,6 +147,29 @@ pub async fn get_connection(pool: &PgPool) -> Result<AsyncConnectionWrapper<Asyn
 }
 */
 
+/// Gets the connection.
+/// If it was unable to do so (default timeout: 30s), it will keep retrying until it can.
+pub async fn get_conn(db_pool: &PgDbPool) -> PgPoolConnection {
```
bowenyang007, line 152: what about the other get_conn?

rust/processor/src/gap_detector.rs
```diff
@@ -0,0 +1,183 @@
...
+// Number of batches processed before gap detected
+pub const GAP_DETECTION_BATCH_COUNT: u64 = 50;
+// Number of batches to process before updating processor status
+const PROCESSOR_STATUS_UPDATE_BATCH_COUNT: u64 = 10;
+
+pub struct GapDetector {
+    processor_name: String,
+    db_pool: PgDbPool,
+    starting_version: u64,
+    seen_versions: AHashMap<u64, ProcessedVersions>,
```
bowenyang007, line 25: why is this so complicated? Just use a queue sorted by time and pop items that don't have a gap. Once you can't pop anymore record the latest version.

rust/processor/src/gap_detector.rs
```diff
@@ -0,0 +1,183 @@
...
+        self.num_batches_processed_without_gap += 1;
+        while let Some(next_version) = self.seen_versions.remove(&(new_prev_batch.end_version + 1))
+        {
+            new_prev_batch = next_version;
+            self.num_batches_processed_without_gap += 1;
+        }
+        self.maybe_prev_batch = Some(new_prev_batch);
+    }
+
+    async fn update_processor_status(
```
bowenyang007, line 112: Isn't this copy pasting a bunch of code from existing files? Why not just reuse or at least clean up the old stuff.

rust/processor/src/gap_detector.rs
```diff
@@ -0,0 +1,183 @@
...
+            db_pool,
+            starting_version,
+            seen_versions: AHashMap::new(),
+            maybe_prev_batch: None,
+            num_batches_processed_without_gap: 0,
+            num_batches_processed_with_gap: 0,
+        }
+    }
+
+    pub async fn process_versions(
```
CapCap, line 50: this no longer needs to be async :-)

rtso: this needs to be async so that it can update processor status in DB. haven't added the write to file part yet but once that's built, we can make this non-async!

rust/processor/src/gap_detector.rs
rtso: i'll clean up the old stuff

rust/processor/src/worker.rs
```diff
@@ -205,6 +209,74 @@ impl Worker {
...
+                            processor_name,
+                            service_type = PROCESSOR_SERVICE_TYPE,
+                            error = ?e,
+                            "[Parser] Gap detector task has panicked"
+                        );
+                        panic!();
+                    },
+                }
+            }
+        });
```
CapCap, line 278: can you move this to it's own function? will make this section much easier to reason through

rust/processor/src/gap_detector.rs
```diff
@@ -0,0 +1,150 @@
...
+                GapDetectorResult::NoGapDetected {
+                    last_success_batch: _,
+                    num_batches_processed_without_gap: _,
+                } => {},
+                _ => panic!("No gap should be detected"),
+            }
+        }
+        let gap_detectgor_result = gap_detector.process_versions(ProcessingResult {
+            start_version: 100 + (GAP_DETECTION_BATCH_COUNT - 1) * 100,
+            end_version: 199 + (GAP_DETECTION_BATCH_COUNT - 1) * 100,
```
bowenyang007, line 118: wait why is this hardcoded to every 100? I don't think that we can make this assumption.

rust/processor/src/gap_detector.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    fn detect_gap(&self, result: ProcessingResult) -> bool {
+        if let Some(prev_batch) = &self.maybe_prev_batch {
+            prev_batch.end_version + 1 != result.start_version
+        } else {
+            result.start_version != self.starting_version
+        }
+    }
+
+    fn update_prev_batch(&mut self, result: ProcessingResult) -> u64 {
```
bowenyang007, line 77: I don't think that you need this given you have starting_version. Why not change starting_version to last_processed_version?

rust/processor/src/gap_detector.rs
```diff
@@ -0,0 +1,198 @@
...
+            last_success_batch: None,
+        }
+    }
+
+    pub fn process_versions(
+        &mut self,
+        result: ProcessingResult,
+    ) -> anyhow::Result<GapDetectorResult> {
+        // Check for gaps
+        let gap_detected = self.has_gap(result.clone());
```
CapCap, line 45: `&result` :-)

rust/processor/src/gap_detector.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    fn update_prev_batch(&mut self, result: ProcessingResult) {
+        let mut new_prev_batch = result;
+        while let Some(next_version) = self.seen_versions.remove(&(new_prev_batch.end_version + 1))
+        {
+            new_prev_batch = next_version;
+        }
+        self.last_success_batch = Some(new_prev_batch.clone());
+        self.next_version_to_process = new_prev_batch.end_version + 1;
```
CapCap, line 73: if you swap the order of these, you won't need the `new_prev_batch.clone()` 

in general, try to minimize any `.clone()` if you can- it's much more expensive to copy all the data than to pass a reference to it

rust/processor/src/gap_detector.rs
```diff
@@ -0,0 +1,193 @@
...
+                .unwrap()
+                .start_version,
+            100 + (GAP_DETECTION_BATCH_COUNT - 1) * 100
+        );
+        assert_eq!(
+            gap_detector_result.last_success_batch.unwrap().end_version,
+            199 + (GAP_DETECTION_BATCH_COUNT - 1) * 100
+        );
+    }
+}
```
CapCap, line 124: can we keep the tests at the end of the file?

CapCap (approved):

larry-aptos (approved):

rust/processor/src/utils/database.rs
bowenyang007: still same question. I thought we had a get_conn in the trait. Are we not using it anymore in favor of this one?

rtso: I delete this now

PR #264, opened 2024-02-08
Process with parallelism

CapCap:
Changes the logic to fetch in batches and throw them into a channel, as before, but no longer worrying about ordering

There are a number of tasks spawned equal to `number_concurrent_processing_tasks` that are launched, each with their own processor instance, which pull directly from the fetched pb channel

Each block of fetched transactions is processed in parallel. Local testing saw 1-2.5k TPS per task 

Next is parallel insertion on the DB side, in much smaller batches

It may make sense to move the pb fetcher to it's own OS thread given how critical it is, and how easy it is to starve otherwise

Will be tested via devnet load test

### DIFF
--- a/rust/processor/src/config.rs
+++ b/rust/processor/src/config.rs
@@ -5,6 +5,7 @@ use crate::{
     gap_detector::DEFAULT_GAP_DETECTION_BATCH_SIZE, processors::ProcessorConfig,
     transaction_filter::TransactionFilter, worker::Worker,
 };
+use ahash::AHashMap;
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 use server_framework::RunnableConfig;
@@ -21,12 +22,23 @@ pub struct IndexerGrpcProcessorConfig {
     #[serde(flatten)]
     pub grpc_http2_config: IndexerGrpcHttp2Config,
     pub auth_token: String,
+    // Version to start indexing from
     pub starting_version: Option<u64>,
+    // Version to end indexing at
     pub ending_version: Option<u64>,
+    // Number of tasks waiting to pull transaction batches from the channel and process them
     pub number_concurrent_processing_tasks: Option<usize>,
+    // Size of the pool for writes/reads to the DB. Limits maximum number of queries in flight
     pub db_pool_size: Option<u32>,
+    // Maximum number of batches "missing" before we assume we have an issue with gaps and abort
     #[serde(default = "IndexerGrpcProcessorConfig::default_gap_detection_batch_size")]
     pub gap_detection_batch_size: u64,
+    // Number of protobuff transactions to send per chunk to the processor tasks
+    #[serde(default = "IndexerGrpcProcessorConfig::default_pb_channel_txn_chunk_size")]
+    pub pb_channel_txn_chunk_size: usize,
+    // Number of rows to insert, per chunk, for each DB table. Default per table is ~32,768 (2**16/2)
+    #[serde(default = "AHashMap::new")]
+    pub per_table_chunk_sizes: AHashMap<String, usize>,
     pub enable_verbose_logging: Option<bool>,
     #[serde(default)]
     pub transaction_filter: TransactionFilter,
@@ -35,6 +47,12 @@ pub struct IndexerGrpcProcessorConfig {
 impl IndexerGrpcProcessorConfig {
     pub const fn default_gap_detection_batch_size() -> u64 {
         DEFAULT_GAP_DETECTION_BATCH_SIZE
+    }
+
+    /// Make the default very large on purpose so that by default it's not chunked
+    /// This prevents any unexpected changes in behavior
+    pub const fn default_pb_channel_txn_chunk_size() -> usize {
+        100_000
     }
 }
 
@@ -52,6 +70,8 @@ impl RunnableConfig for IndexerGrpcProcessorConfig {
             self.number_concurrent_processing_tasks,
             self.db_pool_size,
             self.gap_detection_batch_size,
+            self.pb_channel_txn_chunk_size,
+            self.per_table_chunk_sizes.clone(),
             self.enable_verbose_logging,
             self.transaction_filter.clone(),
         )
--- a/rust/processor/src/gap_detector.rs
+++ b/rust/processor/src/gap_detector.rs
@@ -3,15 +3,15 @@
 
 use crate::{
     processors::{ProcessingResult, Processor, ProcessorTrait},
+    utils::counters::PROCESSOR_DATA_GAP_COUNT,
     worker::PROCESSOR_SERVICE_TYPE,
 };
 use ahash::AHashMap;
 use kanal::AsyncReceiver;
-use std::sync::Arc;
 use tracing::{error, info};
 
-// Number of batches processed before gap detected
-pub const DEFAULT_GAP_DETECTION_BATCH_SIZE: u64 = 50;
+// Size of a gap (in txn version) before gap detected
+pub const DEFAULT_GAP_DETECTION_BATCH_SIZE: u64 = 500;
 // Number of seconds between each processor status update
 const UPDATE_PROCESSOR_STATUS_SECS: u64 = 1;
 
@@ -70,7 +70,7 @@ impl GapDetector {
 
 pub async fn create_gap_detector_status_tracker_loop(
     gap_detector_receiver: AsyncReceiver<ProcessingResult>,
-    processor: Arc<Processor>,
+    processor: Processor,
     starting_version: u64,
     gap_detection_batch_size: u64,
 ) {
@@ -100,16 +100,17 @@ pub async fn create_gap_detector_status_tracker_loop(
 
         match gap_detector.process_versions(result) {
             Ok(res) => {
+                PROCESSOR_DATA_GAP_COUNT
+                    .with_label_values(&[processor_name])
+                    .set(res.num_gaps as i64);
                 if res.num_gaps >= gap_detection_batch_size {
-                    error!(
+                    tracing::debug!(
                         processor_name,
                         gap_start_version = res.next_version_to_process,
                         num_gaps = res.num_gaps,
-                        "[Parser] Processed {gap_detection_batch_size} batches with a gap. Panicking."
+                        "[Parser] Processed {gap_detection_batch_size} batches with a gap",
                     );
-                    panic!(
-                        "[Parser] Processed {gap_detection_batch_size} batches with a gap. Panicking."
-                    );
+                    // We don't panic as everything downstream will panic if it doesn't work/receive
                 }
 
                 if let Some(res_last_success_batch) = res.last_success_batch {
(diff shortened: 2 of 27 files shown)


rust/processor/src/processors/account_transactions_processor.rs
```diff
@@ -42,7 +42,7 @@ async fn insert_to_db(
     name: &'static str,
     start_version: u64,
     end_version: u64,
-    account_transactions: Vec<AccountTransaction>,
+    account_transactions: &[AccountTransaction],
```
rtso, line 45: when should we use `&Vec<T>` vs `&[T]`?

CapCap: always use &[T] when we can; [this](https://github.com/rust-lang/rust-clippy/issues/10269#:~:text=Advantage,dereferences%20instead%20of%20one%20(godbolt)) is a great explanation

rust/processor/src/grpc_stream.rs
```diff
@@ -99,7 +100,13 @@ pub async fn get_stream(
         end_version = ending_version,
         "[Parser] Setting up GRPC client"
     );
-    let mut rpc_client = match RawDataClient::connect(channel).await {
+
+    // Use tokio::time::timeout to set a timeout for the request
+    let connect_res = timeout(Duration::from_secs(1), RawDataClient::connect(channel))
```
rtso, line 105: If it times out, can we add some retry logic?

CapCap: Yeee I can add three retries

rust/processor/src/utils/database.rs
```diff
@@ -206,26 +189,79 @@ where
...
         )
     })?;
+    let res = final_query.execute(conn).await;
+    if let Err(ref e) = res {
+        tracing::warn!("Error running query: {:?}\n{:?}", e, debug_string);
+    }
+    res
+}
 
+pub async fn execute_with_better_error_conn<U>(
```
rtso, line 206: Why do we need both execute_with_better_error_conn and execute_with_better_error?

rust/processor/src/utils/counters.rs
```diff
@@ -202,7 +202,17 @@ pub static SINGLE_BATCH_DB_INSERTION_TIME_IN_SECS: Lazy<GaugeVec> = Lazy::new(||
...
         "indexer_processor_single_batch_db_insertion_time_in_secs",
         "Time taken to insert to DB for a single batch of transactions",
-        &["processor_name"]
+        &["processor_name", "task_index"]
+    )
+    .unwrap()
+});
+
+/// Count of transactions processed per batch
+pub static SINGLE_BATCH_NUM_TRANSACTIONS_PROCESSED_COUNT: Lazy<IntCounterVec> = Lazy::new(|| {
```
rtso, line 211: this seems like a redundant metric of `NUM_TRANSACTIONS_PROCESSED_COUNT`?

rust/processor/src/worker.rs
```diff
@@ -739,6 +514,92 @@ impl Worker {
...
+
+    let start_version = first_txn.version;
+    let end_version = last_txn.version;
+    let txn_time = first_txn.timestamp.clone();
+
+    if let Some(ref t) = txn_time {
+        PROCESSOR_DATA_RECEIVED_LATENCY_IN_SECS
+            .with_label_values(&[auth_token, processor_name])
+            .set(time_diff_since_pb_timestamp_in_secs(t));
+    }
```
rtso, line 569: We can probably remove this metric since we already have `PB_CHANNEL_FETCH_WAIT_TIME_SECS`

rust/processor/src/utils/database.rs
CapCap: one uses the regular diesel, one uses diesel async; diesel async doesn't support `run_without_transaction` or whatever for migrations 😢

rust/processor/src/utils/counters.rs
CapCap: good call!

rust/processor/src/gap_detector.rs
```diff
@@ -88,13 +87,13 @@ pub async fn create_gap_detector_status_tracker_loop(
...
             Err(e) => {
-                info!(
+                error!(
                     processor_name,
                     service_type = PROCESSOR_SERVICE_TYPE,
                     error = ?e,
                     "[Parser] Gap detector channel has been closed",
                 );
-                return;
+                panic!("[Parser] Gap detector channel has been closed");
```
CapCap, line 96: @rtso am I wrong in thinking this should be a panic if it ever happens?

rtso: I think panic should be fine here. The gap detector channel will only be closed if the main thread is finished. There are other places, like in the fetcher channel, where we want to wait for the processor tasks to finish processing before ending the process.

CapCap: I undid this change- reasonable

larry-aptos (approved):
Nice!

also the only usage of regular diesel is for chain id verification/status tracking; should be simple to migrate to async diesel.

rust/processor/src/config.rs
```diff
@@ -26,13 +26,21 @@ pub struct IndexerGrpcProcessorConfig {
     pub db_pool_size: Option<u32>,
     #[serde(default = "IndexerGrpcProcessorConfig::default_gap_detection_batch_size")]
     pub gap_detection_batch_size: u64,
+    #[serde(default = "IndexerGrpcProcessorConfig::default_pb_channel_txn_chunk_size")]
+    pub pb_channel_txn_chunk_size: usize,
```
rtso, line 30: Do we also wanna make a config var for 
* DB max pool size (does that even make sense to have diff max pool size for different processors?)
* # of concurrent write to DB tasks

rust/processor/src/grpc_stream.rs
```diff
@@ -251,8 +288,11 @@ pub async fn create_fetcher_loop(
         "[Parser] Successfully connected to GRPC stream",
     );
 
-    let mut last_fetched_version = batch_start_version as i64 - 1;
-    let mut batch_start_version = batch_start_version;
+    let mut last_fetched_version = starting_version as i64 - 1;
```
rtso, line 291: We have both `next_version_to_fetch` and `last_fetched_version`, could we get rid of one of them

rust/processor/src/utils/database.rs
```diff
@@ -46,14 +45,14 @@ pub const MAX_DIESEL_PARAM_SIZE: u16 = u16::MAX / 2;
 /// Given diesel has a limit of how many parameters can be inserted in a single operation (u16::MAX)
 /// we may need to chunk an array of items based on how many columns are in the table.
 /// This function returns boundaries of chunks in the form of (start_index, end_index)
-pub fn get_chunks(num_items_to_insert: usize, column_count: usize) -> Vec<(usize, usize)> {
-    let max_item_size = MAX_DIESEL_PARAM_SIZE as usize / column_count;
-    let mut chunk: (usize, usize) = (0, min(num_items_to_insert, max_item_size));
+pub fn get_chunks(num_items_to_insert: usize, chunk_size: usize) -> Vec<(usize, usize)> {
+    let chunk_size = min(chunk_size, MAX_DIESEL_PARAM_SIZE as usize);
```
rtso, line 49: We probably still want to keep the default as MAX_DIESEL_PARAM_SIZE / column_count so the default size doesn't go out of bounds.

rust/processor/src/processors/ans_processor.rs
```diff
@@ -76,77 +76,86 @@ async fn insert_to_db(
...
         start_version = start_version,
         end_version = end_version,
         "Inserting to db",
     );
-    execute_in_chunks(
+    let cal = execute_in_chunks(
         conn.clone(),
         insert_current_ans_lookups_query,
         current_ans_lookups,
         CurrentAnsLookup::field_count(),
```
rtso, line 98: We probably need to pass down the DB chunk size through config to here.

rtso (approved):

rust/processor/src/grpc_stream.rs
```diff
@@ -431,12 +430,13 @@ pub async fn create_fetcher_loop(
...
                         let size_in_bytes = average_size_in_bytes * txns.len() as u64;
                         let txn_pb = TransactionsPBResponse {
-                            transactions: r.transactions,
+                            transactions: txns,
                             chain_id,
                             start_version,
                             end_version,
-                            start_txn_timestamp,
-                            end_txn_timestamp,
+                            // TODO: this is only for gap checker + filtered txns, but this is wrong
```
rtso, line 437: thoughts on sending the ProcessingResult straight to the gap detector without having to wrap it up in the struct and send to the processor tasks? that way we don't have to have hacks for the filtered txn's in the processor tasks.

rust/processor/src/utils/counters.rs
```diff
@@ -228,9 +228,9 @@ pub static TRANSACTION_UNIX_TIMESTAMP: Lazy<GaugeVec> = Lazy::new(|| {
 });
 
 /// Data gap warnings
-pub static PROCESSOR_DATA_GAP_COUNT: Lazy<IntCounterVec> = Lazy::new(|| {
-    register_int_counter_vec!("indexer_processor_data_gap_count", "Data gap count", &[
-        "type"
+pub static PROCESSOR_DATA_GAP_COUNT: Lazy<IntGaugeVec> = Lazy::new(|| {
```
rtso, line 231: we can actually delete this metric, just checked it's not used anywhere

rtso (approved):

rust/processor/src/utils/counters.rs
CapCap: I added a use for it in this PR 😬

rust/processor/src/config.rs
```diff
@@ -40,12 +40,13 @@ pub struct IndexerGrpcProcessorConfig {
...
     pub per_table_chunk_sizes: AHashMap<String, usize>,
     pub enable_verbose_logging: Option<bool>,
+    #[serde(default)]
     pub transaction_filter: TransactionFilter,
 }
 
 impl IndexerGrpcProcessorConfig {
     pub const fn default_gap_detection_batch_size() -> u64 {
-        DEFAULT_GAP_DETECTION_BATCH_SIZE
+        DEFAULT_GAP_DETECTION_NUM_TXNS
```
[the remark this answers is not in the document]
CapCap: oi yeah sorry Iw as going to rewrite 🤦 will undo

rust/processor/src/grpc_stream.rs
CapCap: yeah, I think the overall approach of writing from multiple places + counter is good; could we save that for a further PR? I think we could have an enum that's like `completed(start_version, end_version)` `multi_version(start_version, end_version, num_queries)` etc and use that for communication instead of `ProcessingResult`

rust/processor/src/processors/default_processor.rs
```diff
@@ -293,47 +312,72 @@ impl ProcessorTrait for DefaultProcessor {
...
-                    }
-                },
+        let last_transaction_timestamp = transactions.last().unwrap().timestamp.clone();
+        let (
+            txns,
+            block_metadata_transactions,
+            write_set_changes,
+            (move_modules, move_resources, table_items, current_table_items, table_metadata),
+        ) = tokio::task::spawn_blocking(move || {
+            let (txns, block_metadata_txns, write_set_changes, wsc_details) =
```
grao1991, line 322: nit: can we move this block to a separate function?

CapCap: ah... this rebase is going to suck 😭

PR #269, opened 2024-02-08
Pass around DB pools instead of holding connections for long periods

CapCap:
small theoretical performance gain, but mainly in preparation for a world of parallelism, so each sub-task is able to pull from the DB pool whenever it needs a connection, instead of re-using the same one


bowenyang007:
Can we add some explanation in the summary? Is this a performance gain?

bowenyang007 (approved):

PR #273, opened 2024-02-09
exit -> exist

just-in-chang:


rust/processor/src/models/token_models/nft_points.rs
```diff
@@ -38,7 +38,7 @@ impl NftPoints {
         let txn_data = transaction
             .txn_data
             .as_ref()
-            .expect("Txn Data doesn't exit!");
+            .expect("Txn Data doesn't exist!");
```
jillxuu, line 41: lets all add txn versions here as well? apply to all others

jillxuu (approved):

PR #276, opened 2024-02-10
Fix token burns

rtso:
## Description 
Handle token burn when the mint happened in the same transaction 

## Testing
Tested locally on this transaction: https://explorer.aptoslabs.com/txn/445585423/events?network=mainnet

### DIFF
--- a/rust/processor/src/models/token_v2_models/v2_token_activities.rs
+++ b/rust/processor/src/models/token_v2_models/v2_token_activities.rs
@@ -8,7 +8,7 @@
 use super::{
     v2_token_datas::TokenDataV2,
     v2_token_ownerships::CurrentTokenOwnershipV2Query,
-    v2_token_utils::{TokenStandard, V2TokenEvent},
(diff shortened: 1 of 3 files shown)


rust/processor/src/models/token_v2_models/v2_token_activities.rs
```diff
@@ -238,6 +239,28 @@ impl TokenActivityV2 {
...
+                    return Ok(Some(Self {
+                        transaction_version: txn_version,
+                        event_index,
+                        event_account_address,
+                        token_data_id,
+                        property_version_v1: BigDecimal::zero(),
+                        type_: event_type,
+                        from_address: None,
+                        to_address: None,
+                        token_amount: BigDecimal::one(),
```
bowenyang007, line 254: shouldn't this be 0?

rtso: just following what we do for burns on line 292. we set the token_amount to 1

jillxuu (approved):

PR #278, opened 2024-02-12
Rtso/generic types

rtso:


rust/processor/src/processors/default_processor.rs
```diff
@@ -152,6 +156,36 @@ fn insert_transactions_query(
     )
 }
 
+#[derive(Clone, Debug)]
+pub struct TransactionModelBatch {
+    start_version: StartVersion,
+    end_version: EndVersion,
+    items_to_insert: Vec<TransactionModel>,
```
CapCap, line 163: can these be `&[TransactionModel]`?

rust/processor/src/processors/default_processor.rs
```diff
@@ -152,6 +156,36 @@ fn insert_transactions_query(
...
+        &self,
+    ) -> (
+        impl QueryFragment<Pg> + diesel::query_builder::QueryId + Send,
+        Option<&'static str>,
+    ) {
+        use schema::transactions::dsl::*;
+
+        (
+            diesel::insert_into(schema::transactions::table)
+                .values(self.items_to_insert.clone())
```
CapCap, line 177: do we need to clone here, if it's a ref?

rust/processor/src/processors/default_processor.rs
rtso: i asked also in your PR. when do we use `&[TransactionModel]` vs `&Vec<TransactionModel>`?

rust/processor/src/processors/default_processor.rs
```diff
@@ -380,4 +414,67 @@ impl ProcessorTrait for DefaultProcessor {
...
         &self.connection_pool
     }
+
+    fn parse_transactions(
+        &self,
+        transactions: Vec<Transaction>,
+        start_version: u64,
+        end_version: u64,
+        _: Option<u64>,
+    ) -> anyhow::Result<Option<ProcessingParseResult>> {
```
CapCap, line 424: is the idea this would return something that implements `QueryModelBatchTrait` ?

rust/processor/src/processors/default_processor.rs
```diff
@@ -380,4 +414,67 @@ impl ProcessorTrait for DefaultProcessor {
...
+        let (txns, block_metadata_txns, write_set_changes, wsc_details) =
+            TransactionModel::from_transactions(&transactions);
+
+        let mut block_metadata_transactions = vec![];
+        for block_metadata_txn in block_metadata_txns {
+            block_metadata_transactions.push(block_metadata_txn.clone());
+        }
```
CapCap, line 431: this seems like an un-needed clone

rust/processor/src/query_models.rs
```diff
@@ -0,0 +1,29 @@
...
+        Option<&'static str>,
+    );
+}
+
+#[enum_dispatch(QueryModelBatchTrait)]
+#[derive(Clone, Debug)]
+pub enum QueryModelBatch {
+    // BlockMetadataTransactionModel,
+    TransactionModelBatch,
+}
```
CapCap, line 29: instead of using an enum, just have it be a trait- otherwise this is impossible for anyone in our ecosystem to re-use without some hoops

rust/processor/src/processors/default_processor.rs
rtso: after processing we should return `ProcessingParseResult` which has a field for an array of `QueryModelBatchTrait`. the reason i did it this way instead of returning a vec of `QueryModelBatchTrait`, each with its own start and end versions is because it makes gap detection really difficult.

rust/processor/src/query_models.rs
rtso: i ran into compiler issues using trait objects with dynamic dispatch. 

if they want to create their own, they can make make a new enum with `#[enum_dispatch(QueryModelBatchTrait)]`

rtso: actually i take that back. this makes it impossible for other folks to add a new enum...

PR #285, opened 2024-02-16
upgrade processor to grpc 1.4 (node 1.10)

bowenyang007:
https://github.com/aptos-labs/aptos-core/releases/tag/aptos-indexer-grpc-v1.4.0

Add new validator transaction handling and change signature parsing

also remove ModuleBundlePayload b/c protobuf doesn't support anymore. 

## Backfill
None

## Testing
in case of an unknown transaction
image

yay! it works for new types
image

### DIFF
--- a/rust/processor/src/utils/util.rs
+++ b/rust/processor/src/utils/util.rs
@@ -1,7 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::models::property_map::{PropertyMap, TokenObjectPropertyMap};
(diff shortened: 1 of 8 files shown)


rust/processor/src/utils/util.rs
```diff
@@ -129,12 +139,6 @@ pub fn get_clean_payload(payload: &TransactionPayload, version: i64) -> Option<V
                 panic!()
             }))
         },
-        PayloadType::ModuleBundlePayload(inner) => {
```
rtso, line 132: why delete this?

bowenyang007: grpc doesn't have this anymore.

bowenyang007: it was part of the protobuf change.

rtso (approved):

PR #288, opened 2024-02-21
Move DB upserts to it's own batch of tasks via channel

CapCap:
# Worker Flow
- The application starts by initializing a `Worker` instance with the necessary configurations such as the processor
    configuration, database connection string, GRPC data service address, and other parameters.

- The `Worker` instance then runs migrations on the database to ensure the schema is up-to-date.

- The `Worker` fetches the chain ID from the GRPC service and verifies it against the database.

- The `Worker` then starts a fetcher task that continuously fetches transactions from the GRPC stream and writes them into a channel.
    The number of transactions fetched in each batch is determined by the `pb_channel_txn_chunk_size` parameter.

- Concurrently, the `Worker` also starts multiple processor tasks that consume the transactions from the channel.
    These tasks process the transactions in parallel.
    The number of processor tasks is determined by the `number_concurrent_processing_tasks` parameter.
    TODO: is this right? The size of the channel is determined by the `PB_FETCH_QUEUE_SIZE` parameter.

- Each processor task uses a specific `Processor` instance to process the transactions.
    The type of `Processor` used depends on the configuration provided when initializing the `Worker`.
    Each `Processor` type corresponds to a different way of processing transactions.

- After processing the transactions, the processor tasks send the results to a gap detector.
    The gap detector checks for any gaps in the processed transactions and panics if it finds any.
    The maximum batch size for gap detection is determined by the `gap_detection_batch_size` parameter.

- The processed transactions are also sent to the `DbWriter` instance associated with the `Processor`, via a channel.
    The `DbWriter` is responsible for writing the processed transactions to the database ("executing" them).
    The size of the channel is determined by the `query_executor_channel_size` parameter.
    The number of concurrent DB writer tasks is determined by the `number_concurrent_db_writer_tasks` parameter.

- The `DbWriter` sends the transactions to be written to the database in chunks.
    It uses an `AsyncSender` to send `QueryGenerator` instances to a DB writer task.
    Each `QueryGenerator` contains a table name and a `DbExecutable` instance, which represents the transactions to be written to the database.
    The chunk size for sending queries to the database is determined by the `per_table_chunk_sizes` parameter;
    this parameter specifies the maximum number of rows to be inserted in a single query. It is a map from table name to chunk size.

- The DB writer task executes the queries represented by the `DbExecutable` instances.
    If an error occurs during execution, it logs the error and continues with the next query.

- This process continues in a loop, with the fetcher task fetching transactions,
    the processor tasks processing them, and the DB writer task writing them to the database,
    and the gap detector ensuring that if there is a large gap in the transactions, it panics.

```
# Architecture Diagram

 ┌──────────────┐
 │ GRPC Service │
 └─────┬────────┘
       │Stream
 ┌─────▼────────┐ Transaction ┌───────────┐
 │ Fetcher Task ├────Chunk ───► Processor │
 └─────┬────────┘   Channel   │ Tasks     │
       │ Channel              └────┬──────┘
 ┌─────▼────────┐                  │ DbWriter
 │ Gap Detector │                  │ Channel
 └──────────────┘             ┌────▼────────┐
                              │ DB Executor │
                              └────┬────────┘
                                   │
                              ┌────▼─────┐
                              │ Database │
                              └──────────┘
```

Also adds the ability to tune insert sizes per table, and skip writes to tables entirely 😄

### DIFF
--- a/rust/processor/src/main.rs
+++ b/rust/processor/src/main.rs
@@ -1,7 +1,6 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-#![feature(type_alias_impl_trait)]
 use anyhow::Result;
 use clap::Parser;
 use processor::IndexerGrpcProcessorConfig;
@@ -11,7 +10,7 @@ const RUNTIME_WORKER_MULTIPLIER: usize = 2;
 
 fn main() -> Result<()> {
     let num_cpus = num_cpus::get();
-    let worker_threads = (num_cpus * RUNTIME_WORKER_MULTIPLIER).max(16);
+    let worker_threads = num_cpus * RUNTIME_WORKER_MULTIPLIER;
     println!(
         "[Processor] Starting processor tokio runtime: num_cpus={}, worker_threads={}",
         num_cpus, worker_threads
--- a/rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
+++ b/rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
@@ -80,47 +80,6 @@ pub struct CurrentTokenOwnershipV2 {
     pub non_transferrable_by_owner: Option<bool>,
 }
 
-impl From<CurrentTokenOwnershipV2> for CurrentDeletedTokenOwnershipV2 {
-    fn from(val: CurrentTokenOwnershipV2) -> Self {
-        CurrentDeletedTokenOwnershipV2 {
-            token_data_id: val.token_data_id,
-            property_version_v1: val.property_version_v1,
-            owner_address: val.owner_address,
-            storage_id: val.storage_id,
-            amount: val.amount,
-            table_type_v1: val.table_type_v1,
-            token_properties_mutated_v1: val.token_properties_mutated_v1,
-            is_soulbound_v2: val.is_soulbound_v2,
-            token_standard: val.token_standard,
-            is_fungible_v2: val.is_fungible_v2,
-            last_transaction_version: val.last_transaction_version,
-            last_transaction_timestamp: val.last_transaction_timestamp,
-            non_transferrable_by_owner: val.non_transferrable_by_owner,
-        }
-    }
-}
-
-// because we have the impl on Vec<T> we need to create a wrapper struct to implement the trait if
-// we want to use the same underlying type between different queries
-#[derive(Clone, Debug, Deserialize, FieldCount, Identifiable, Insertable, Serialize)]
-#[diesel(primary_key(token_data_id, property_version_v1, owner_address, storage_id))]
-#[diesel(table_name = current_token_ownerships_v2)]
-pub struct CurrentDeletedTokenOwnershipV2 {
-    pub token_data_id: String,
-    pub property_version_v1: BigDecimal,
-    pub owner_address: String,
-    pub storage_id: String,
-    pub amount: BigDecimal,
-    pub table_type_v1: Option<String>,
-    pub token_properties_mutated_v1: Option<serde_json::Value>,
-    pub is_soulbound_v2: Option<bool>,
-    pub token_standard: String,
-    pub is_fungible_v2: Option<bool>,
-    pub last_transaction_version: i64,
-    pub last_transaction_timestamp: chrono::NaiveDateTime,
-    pub non_transferrable_by_owner: Option<bool>,
-}
-
 // Facilitate tracking when a token is burned
 #[derive(Clone, Debug)]
 pub struct NFTOwnershipV2 {
--- a/rust/processor/src/processors/account_transactions_processor.rs
+++ b/rust/processor/src/processors/account_transactions_processor.rs
@@ -4,10 +4,9 @@
 use super::{ProcessingResult, ProcessorName, ProcessorTrait};
 use crate::{models::account_transaction_models::account_transactions::AccountTransaction, schema};
 use ahash::AHashMap;
-use anyhow::bail;
 use aptos_protos::transaction::v1::Transaction;
 use async_trait::async_trait;
-use tracing::error;
+use diesel::query_builder::QueryFragment;
 
 pub struct AccountTransactionsProcessor {
     db_writer: crate::db_writer::DbWriter,
@@ -32,39 +31,15 @@ impl AccountTransactionsProcessor {
     }
 }
 
-async fn insert_to_db(
-    db_writer: &crate::db_writer::DbWriter,
-    name: &'static str,
-    start_version: u64,
-    end_version: u64,
-    account_transactions: Vec<AccountTransaction>,
-) -> Result<(), diesel::result::Error> {
-    tracing::trace!(
-        name = name,
-        start_version = start_version,
-        end_version = end_version,
-        "Inserting to db",
-    );
-    db_writer
-        .send_in_chunks("account_transactions", account_transactions)
-        .await;
-    Ok(())
-}
+pub fn insert_account_transactions_query(
+    items_to_insert: &[AccountTransaction],
+) -> impl QueryFragment<diesel::pg::Pg> + diesel::query_builder::QueryId + Sync + Send + '_ {
+    use crate::schema::account_transactions::dsl::*;
 
-#[async_trait::async_trait]
-impl crate::db_writer::DbExecutable for Vec<AccountTransaction> {
-    async fn execute_query(
-        &'_ self,
-        conn: crate::utils::database::PgDbPool,
-    ) -> diesel::QueryResult<usize> {
-        use crate::schema::account_transactions::dsl::*;
-
-        let query = diesel::insert_into(schema::account_transactions::table)
-            .values(self)
-            .on_conflict((transaction_version, account_address))
-            .do_nothing();
-        crate::db_writer::execute_with_better_error(conn, query).await
-    }
+    diesel::insert_into(schema::account_transactions::table)
+        .values(items_to_insert)
+        .on_conflict((transaction_version, account_address))
+        .do_nothing()
 }
 
 #[async_trait]
@@ -99,36 +74,30 @@ impl ProcessorTrait for AccountTransactionsProcessor {
         });
 
         let processing_duration_in_secs = processing_start.elapsed().as_secs_f64();
+
         let db_insertion_start = std::time::Instant::now();
-        let tx_result = insert_to_db(
-            self.db_writer(),
-            self.name(),
-            start_version,
-            end_version,
-            account_transactions,
-        )
-        .await;
+        tracing::trace!(
+            name = self.name(),
+            start_version = start_version,
+            end_version = end_version,
+            "Finished parsing, sending to DB",
+        );
+        self.db_writer()
+            .send_in_chunks(
+                "account_transactions",
+                account_transactions,
+                insert_account_transactions_query,
+            )
+            .await;
 
         let db_channel_insertion_duration_in_secs = db_insertion_start.elapsed().as_secs_f64();
-        match tx_result {
-            Ok(_) => Ok(ProcessingResult {
-                start_version,
-                end_version,
-                processing_duration_in_secs,
-                db_channel_insertion_duration_in_secs,
-                last_transaction_timestamp,
-            }),
-            Err(err) => {
-                error!(
-                    start_version = start_version,
-                    end_version = end_version,
-                    processor_name = self.name(),
-                    "[Parser] Error inserting transactions to db: {:?}",
-                    err
-                );
-                bail!(format!("Error inserting transactions to db. Processor {}. Start {}. End {}. Error {:?}", self.name(), start_version, end_version, err))
-            },
-        }
+        Ok(ProcessingResult {
+            start_version,
+            end_version,
+            processing_duration_in_secs,
+            db_channel_insertion_duration_in_secs,
+            last_transaction_timestamp,
+        })
     }
 
     fn db_writer(&self) -> &crate::db_writer::DbWriter {
--- a/rust/processor/src/processors/stake_processor.rs
+++ b/rust/processor/src/processors/stake_processor.rs
@@ -21,11 +21,11 @@ use crate::{
     utils::util::{parse_timestamp, standardize_address},
 };
 use ahash::AHashMap;
-use anyhow::bail;
 use aptos_protos::transaction::v1::{write_set_change::Change, Transaction};
 use async_trait::async_trait;
-use diesel::{pg::upsert::excluded, query_dsl::filter_dsl::FilterDsl};
-use tracing::error;
+use diesel::{
+    pg::upsert::excluded, query_builder::QueryFragment, query_dsl::filter_dsl::FilterDsl,
+};
 
 pub struct StakeProcessor {
     db_writer: crate::db_writer::DbWriter,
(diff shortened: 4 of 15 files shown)


rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,98 @@
...
+use tokio::task::JoinHandle;
+use tracing::info;
+
+pub async fn launch_db_writer_tasks<
+    Query: diesel::query_builder::QueryFragment<diesel::pg::Pg>
+    + diesel::query_builder::QueryId
+    + Send
+    + 'static,
+>(
+    query_receiver: AsyncReceiver<Query>,
```
rtso, line 20: So the query needs to be built before it can be sent to the `query_sender`. We can do that as part of the processor tasks since it's CPU-bound. Later we can think about splitting out query building as a separate task. 

So then the idea is that `process_transactions` will return the built + chunked queries. Then the queries get send to `query_sender` in worker.rs. Am I understanding that correctly?

CapCap: yep :-) exactly

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,98 @@
...
+    })
+}
+
+async fn execute_or_retry<Query>(conn: PgDbPool, query: Query) -> Result<(), diesel::result::Error>
+    where
+        Query: diesel::query_builder::QueryFragment<diesel::pg::Pg>
+        + diesel::query_builder::QueryId
+        + Send,
+{
+    // TODO: retry
```
rtso, line 71: The hard part (which I couldn't figure out lol) is that if we do the retry with cleaned query here, it will need a reference to the items to be cleaned and the function to build the query. 

So I thought maybe we could have a struct, QueryWriterItem, that includes the built query (uncleaned), the items, the build_query function. But doing it this way I ran into a lot of compiler issues.

just-in-chang (approved):

rust/processor/src/processors/stake_processor.rs
```diff
@@ -17,63 +18,52 @@ use crate::{
...
-    proposal_votes: &[ProposalVote],
-    delegator_actvities: &[DelegatedStakingActivity],
-    delegator_balances: &[DelegatorBalance],
-    current_delegator_balances: &[CurrentDelegatorBalance],
-    delegator_pools: &[DelegatorPool],
-    delegator_pool_balances: &[DelegatorPoolBalance],
-    current_delegator_pool_balances: &[CurrentDelegatorPoolBalance],
-    current_delegated_voter: &[CurrentDelegatedVoter],
-    per_table_chunk_sizes: &AHashMap<String, usize>,
+    current_stake_pool_voters: Vec<CurrentStakingPoolVoter>,
```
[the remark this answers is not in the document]
CapCap: If you don't need the receiving function to own it, then `&[T]` is perfect. If the receiving function has to `.clone()` and then parent function doesn't need its own copy, then `Vec<T>`

rust/processor/src/models/user_transactions_models/user_transactions.rs
```diff
@@ -103,3 +106,23 @@ impl UserTransaction {
 
 // Prevent conflicts with other things named `Transaction`
 pub type UserTransactionModel = UserTransaction;
+
+#[async_trait::async_trait]
+impl crate::db_writer::DbExecutable for Vec<UserTransactionModel> {
```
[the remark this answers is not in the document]
CapCap: I was hoping to move all of the DB related stuff to models, and then all processing stuff to processors

rust/processor/src/processors/stake_processor.rs
```diff
@@ -535,45 +441,44 @@ impl ProcessorTrait for StakeProcessor {
...
                 error!(
-                    start_version = start_version,
-                    end_version = end_version,
-                    processor_name = self.name(),
-                    error = ?e,
-                    "[Parser] Error inserting transactions to db",
+                start_version = start_version,
+                end_version = end_version,
+                processor_name = self.name(),
+                error = ? e,
```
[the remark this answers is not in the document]
CapCap: idk why the linter doesnt lint it... weird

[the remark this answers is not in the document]
just-in-chang: my linter also sussy sometimes
and CI passes anyway so... prob issue with linter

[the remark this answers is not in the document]
CapCap: nah you right this is weird; I tried indenting one of the lines and the linter didn't fix that either

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub async fn send_in_chunks<Item>(&self, table_name: &'static str, items_to_insert: Vec<Item>)
+    where
+        Item: field_count::FieldCount
+            + serde::Serialize
+            + for<'de> serde::Deserialize<'de>
+            + Send
+            + Sync
+            + Clone
+            + 'static,
+        Vec<Item>: DbExecutable,
```
rtso, line 200: how does contraints on a Vec<T> work? i thought you could only constraint generics.

rust/processor/src/db_writer.rs
rtso, line 1: this whole `db_writer.rs` file is chef's kiss

rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
```diff
@@ -80,6 +80,47 @@ pub struct CurrentTokenOwnershipV2 {
...
+        }
+    }
+}
+
+// because we have the impl on Vec<T> we need to create a wrapper struct to implement the trait if
+// we want to use the same underlying type between different queries
+#[derive(Clone, Debug, Deserialize, FieldCount, Identifiable, Insertable, Serialize)]
+#[diesel(primary_key(token_data_id, property_version_v1, owner_address, storage_id))]
+#[diesel(table_name = current_token_ownerships_v2)]
+pub struct CurrentDeletedTokenOwnershipV2 {
```
rtso, line 108: haha yep i did this originally but removed it because it's not DRY :)

rust/processor/src/processors/account_transactions_processor.rs
```diff
@@ -2,86 +2,69 @@
...
-        insert_account_transactions_query,
-        account_transactions,
-        get_config_table_chunk_size::<AccountTransaction>(
-            "account_transactions",
-            per_table_chunk_sizes,
-        ),
-    )
-    .await?;
+    db_writer
+        .send_in_chunks("account_transactions", account_transactions)
```
rtso, line 49: would it be cleaner to have the `table_name` as a function in the `DBExecutable` struct? we don't have to worry about mismatching table_name and structs. 

or... even better if there's a way to get table_name from the `AccountTransaction` itself (i remember you were looking into that).

rust/processor/src/db_writer.rs
CapCap: you can `impl` on a `Vec<Item>` :-)

rust/processor/src/processors/account_transactions_processor.rs
CapCap: yeah... I think this is where writing a macro or forking diesel might come into play

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,300 @@
...
+                // If res is a success, we can terminate this loop
+                Ok(_) => break,
+                Err(e) => {
+                    num_tries += 1;
+                    let error_type = diesel_error_to_metric_str(e);
+                    DB_EXECUTION_RETRIES_COUNT
+                        .with_label_values(&[processor_name, table_name, error_type])
+                        .inc();
+
+                    // TODO: handle CDB retries differently here (higher retry count?)
```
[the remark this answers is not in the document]
CapCap: CBD has 'expected' failures where the DB will tell the client to re-submit the whole query

rust/processor/src/processors/token_v2_processor.rs
```diff
@@ -94,172 +87,83 @@ async fn insert_to_db(
...
+    let td_v2 = db_writer.send_in_chunks("token_datas_v2", token_datas_v2);
+    let to_v2 = db_writer.send_in_chunks("token_ownerships_v2", token_ownerships_v2);
+    let cc_v2 = db_writer.send_in_chunks("current_collections_v2", current_collections_v2);
+    let ctd_v2 = db_writer.send_in_chunks("current_token_datas_v2", current_token_datas_v2);
+    let cto_v2 =
+        db_writer.send_in_chunks("current_token_ownerships_v2", current_token_ownerships_v2);
+    let cbto_v2 = db_writer.send_in_chunks(
+        "current_token_ownerships_v2",
+        current_burned_token_ownerships_v2,
     );
```
rtso, line 100: currently the model associates a `Vec<Item>` with a single query. i feel like this is the wrong model because there can be multiple queries that can be run on a Vec of the same items. I think better UX would be at `db_writer.send` time, decide which vec + query combo to send to db writer. maybe have the execute_query fn as part of `QueryGenerator`? not sure how feasible this would be though.

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,309 @@
...
+        Error::RollbackErrorOnCommit { .. } => "RollbackErrorOnCommit",
+        Error::NotInTransaction => "NotInTransaction",
+        Error::BrokenTransactionManager => "BrokenTransactionManager",
+        _ => "Other",
+    }
+}
+
+// A holder struct for processors db writing so we don't need to keep adding new params
+#[derive(Clone)]
+pub struct DbWriter {
```
grao1991, line 162: I feel this should be called as DbQueryGenerator, QueryGenerator should be called as Query, and launch_db_writer_task should be DbWriter.

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,309 @@
...
+
+// This is a wrapper struct that allows us to pass additional data alongside the DbExecutable
+pub struct QueryGenerator {
+    pub table_name: &'static str,
+    pub db_executable: Box<dyn DbExecutable>,
+}
+
+pub fn launch_db_writer_task(
+    query_receiver: AsyncReceiver<QueryGenerator>,
+    processor_name: &'static str,
```
grao1991, line 257: would it be better to pass this per query?

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,309 @@
...
+) {
+    info!(
+        processor_name = processor_name,
+        service_type = PROCESSOR_SERVICE_TYPE,
+        "[Parser] Starting db writer tasks",
+    );
+    let tasks = (0..num_tasks)
+        .map(|_| launch_db_writer_task(query_receiver.clone(), processor_name, conn.clone()))
+        .collect::<Vec<_>>();
+    futures::future::try_join_all(tasks)
```
grao1991, line 244: you probably want a separate runtime for this

rust/processor/src/processors/token_v2_processor.rs
CapCap: Fixed :-)

rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
CapCap: now I removed too! nice :-)

rust/processor/src/db_writer.rs
CapCap: agreed, I probably do- will do those changes in a follow up PR

rust/processor/src/db_writer.rs
CapCap: I added `DataWithQuery`- let me know if you think the current is more or less clear?

rust/processor/src/db_writer.rs
CapCap: I want to move this to a trait or OOP or something instead of all this passing

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+        .expect("Error joining db writer tasks");
+}
+
+// This is a wrapper struct that allows us to pass additional data alongside the DbExecutable
+pub struct QueryGenerator {
+    pub table_name: &'static str,
+    pub db_executable: Box<dyn DbExecutable>,
+}
+
+pub fn launch_db_writer_task(
```
grao1991, line 271: nit: don't need pub?

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+            .set(insertion_timer.elapsed().as_secs_f64());
+
+        DB_INSERTION_ROWS_COUNT
+            .with_label_values(&[processor_name, table_name])
+            .inc_by(self.len() as u64);
+        res
+    }
+}
+
+pub fn diesel_error_to_metric_str(error: &Error) -> &'static str {
```
grao1991, line 133: nit: don't need pub?

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+pub struct FakeTable {}
+impl field_count::FieldCount for FakeTable {
+    fn field_count() -> usize {
+        100
+    }
+}
+
+// A holder struct for processors db writing so we don't need to keep adding new params
+#[derive(Clone)]
+pub struct DbWriter {
```
grao1991, line 171: nit: can we use pub(crate) when possible?

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+// This is a fake table for default fallback chunk size
+pub struct FakeTable {}
+impl field_count::FieldCount for FakeTable {
+    fn field_count() -> usize {
+        100
+    }
+}
+
+// A holder struct for processors db writing so we don't need to keep adding new params
+#[derive(Clone)]
```
grao1991, line 170: We probably don't need Clone here?

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+pub struct DbWriter {
+    pub processor_name: &'static str,
+    pub db_pool: PgDbPool,
+    pub query_sender: AsyncSender<QueryGenerator>,
+    pub per_table_chunk_sizes: AHashMap<String, usize>,
```
grao1991, line 175: nit: most of fields don't need to be pub?

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+pub trait DbExecutable: DbCleanable + DbRowCountable + Send {
+    async fn execute_query(&self, db_pool: PgDbPool) -> QueryResult<usize>;
+
+    // TODO: make this config/constant?
+    fn max_retries(&self) -> usize {
+        2
+    }
+
+    async fn execute_or_retry_cleaned(
+        &mut self,
```
grao1991, line 72: do we need &mut here?

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+
+pub fn launch_db_writer_task(
+    query_receiver: AsyncReceiver<QueryGenerator>,
+    processor_name: &'static str,
+    db_pool: PgDbPool,
+) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        loop {
+            match query_receiver.recv().await {
+                Ok(mut query_generator) => {
```
grao1991, line 279: do we need mut here?

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+                    let query_res = query_generator
+                        .db_executable
+                        .execute_or_retry_cleaned(
+                            processor_name,
+                            query_generator.table_name,
+                            db_pool.clone(),
+                        )
+                        .await;
+                    query_res.expect("Error executing query");
+                    drop(query_generator);
```
grao1991, line 289: do we need to explicitly drop here?
should we drop it in a background thread?

rust/processor/src/db_writer.rs
grao1991: Still feel the `QueryGenerator` is a weird name, especially when it doesn't have a method called `generate`.

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+        Self {
+            processor_name,
+            db_pool,
+            query_sender,
+            per_table_chunk_sizes,
+            skip_tables,
+        }
+    }
+
+    pub fn chunk_size<Item: field_count::FieldCount>(&self, table_name: &str) -> usize {
```
grao1991, line 196: not related to this PR, but one day we probably also need to have limit on bytes size in addition to # of rows.

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        DB_EXECUTOR_CHANNEL_CHUNK_INSERT_COUNT
+            .with_label_values(&[self.processor_name, table_name])
+            .inc();
+
+        let chunk_size = self.chunk_size::<Item>(table_name);
+        let chunks = get_chunks(items_to_insert.len(), chunk_size);
+        for (start_ind, end_ind) in chunks {
+            let items = items_to_insert[start_ind..end_ind].to_vec();
```
grao1991, line 228: is it a copy here?

rust/processor/src/main.rs
```diff
@@ -10,7 +10,7 @@ const RUNTIME_WORKER_MULTIPLIER: usize = 2;
 
 fn main() -> Result<()> {
     let num_cpus = num_cpus::get();
-    let worker_threads = (num_cpus * RUNTIME_WORKER_MULTIPLIER).max(16);
+    let worker_threads = num_cpus * RUNTIME_WORKER_MULTIPLIER;
```
grao1991, line 13: not related to this PR, but why do we need a multiplier here?

rust/processor/src/db_writer.rs
CapCap: I will always pub everything because the number of times something being pub that shouldn't be has hurt me across my career is 0, whereas the number of times i've had to fork things because things arent public is somewhere around 1000 :-P

rust/processor/src/db_writer.rs
CapCap: why?

rust/processor/src/db_writer.rs
CapCap: we clone a bunch when passing to different processors

rust/processor/src/db_writer.rs
CapCap: I explicitly drop so that I can be guaranteed it holds across this unsafe block

rust/processor/src/db_writer.rs
grao1991: my experience was the other way around 🤔 I heavily rely on visibility to reason about the design when reading other people's code.

rust/processor/src/db_writer.rs
```diff
@@ -0,0 +1,325 @@
...
+
+#[async_trait::async_trait]
+/// The `DbExecutable` trait is for types that can be executed in the database
+pub trait DbExecutable: DbCleanable + DbRowCountable + Send {
+    async fn execute_query(&self, db_pool: PgDbPool) -> QueryResult<usize>;
+
+    // TODO: make this config/constant?
+    fn max_retries(&self) -> usize {
+        2
+    }
```
just-in-chang, line 69: We have #321 to make num retries and timeout into config params for the 1.10 release, could follow smtg similar?

CapCap: yes- I want to wait until we can pipe configs through to processors more easily

rust/processor/src/db_writer.rs
CapCap: ex: diesel doesnt expose table names; we'd need a proc macro, or to fork, or PR it, etc. Easily avoidable mess 🙃

rust/processor/src/db_writer.rs
just-in-chang: Not related to this PR, but in the future we should prob have a system that allows each processor to have a "context" that can be accessed/passed around from where ever... The context would just contain  processor specific stuff like Arcs of channel txs/rxs, in-memory cache, config, etc

Been thinking of ways to do this but think the best is just to have a trait in SDK 🤣

rust/processor/src/db_writer.rs
CapCap: it is :-( I think there's likely a way to avoid it (and I didn't want `pop` or whatever) but I did not do it yet

rust/processor/src/main.rs
CapCap: hyperthreading? IDK aptos-core had this I think

rust/processor/src/db_writer.rs
CapCap: 100000%%%%

rust/processor/src/main.rs
grao1991: `num_cpus::get()` actually considers hyperthreading.

PR #289, opened 2024-02-21
[rust] making signatures backward compatible

bowenyang007:
## Summary
We changed parsing for signature in 1.10 so this makes sure that we handle both not just the new one. 

## Backfill
None

## Testing

Before
image

After
image

### DIFF
--- a/rust/processor/src/models/user_transactions_models/signatures.rs
+++ b/rust/processor/src/models/user_transactions_models/signatures.rs
@@ -356,6 +356,7 @@ impl Signature {
                 },
                 wildcard => {
                     tracing::warn!(
+                        transaction_version = transaction_version,
                         "Unspecified signature type or un-recognized type is not supported: {:?}",
                         wildcard
                     );
@@ -433,6 +434,7 @@ impl Signature {
                     },
                     wildcard => {
                         tracing::warn!(
+                            transaction_version = transaction_version,
                             "Unspecified signature type or un-recognized type is not supported: {:?}",
                             wildcard
                         );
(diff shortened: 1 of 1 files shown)


rust/processor/src/models/user_transactions_models/signatures.rs
```diff
@@ -338,9 +340,32 @@ impl Signature {
...
+            // old way of getting signature type prior to node 1.10
+            match AnySignatureTypeEnumPb::try_from(signature.r#type) {
+                Ok(AnySignatureTypeEnumPb::Ed25519) => String::from("single_key_ed25519_signature"),
+                Ok(AnySignatureTypeEnumPb::Secp256k1Ecdsa) => {
+                    String::from("single_key_secp256k1_ecdsa_signature")
+                },
+                wildcard => {
+                    tracing::warn!(
+                        "Unspecified signature type or un-recognized type is not supported: {:?}",
+                        wildcard
```
rtso, line 360: nit: log transaction_version

rust/processor/src/models/user_transactions_models/signatures.rs
```diff
@@ -338,9 +340,32 @@ impl Signature {
...
+                    );
+                    PROCESSOR_UNKNOWN_TYPE_COUNT
+                        .with_label_values(&["unspecified_signature_type"])
+                        .inc();
+                    "".to_string()
+                },
+            }
```
rtso, line 367: same comment. we can wrap it within the `get_any_signature_type`

rtso (approved):
left some comments, otherwise lgtm.

PR #290, opened 2024-02-22
Integrating the latest token v2 changes

vusirikala:
## Description 
This PR incorporates the latest Token V2 changes in https://github.com/aptos-labs/aptos-core/pull/12145/.
This PR mainly updates ConcurrentMintEvent --> Mint and ConcurrentBurnEvent -->  Burn

## Backfill
* Testnet: TBD not in testnet yet
* Mainnet: TBD not in mainnet yet 

### Test 1
Index burn with object core still there
Screenshot 2024-02-28 at 9 39 00 PM

### Test 2
Index mint
Screenshot 2024-02-28 at 9 40 47 PM
I manually modified is_soulbound_v2 to be false. After indexing the burn, is_soulbound_v2 is unchanged
Screenshot 2024-02-28 at 9 42 13 PM

### Test 3
https://explorer.aptoslabs.com/txn/929569453/events?network=testnet
Mint, transfer, old burn in 1 txn
Screenshot 2024-02-29 at 6 36 33 PM

Screenshot 2024-02-29 at 6 37 18 PM

### Test 4
Mint, transfer x2, old burn in 1 txn 
https://explorer.aptoslabs.com/txn/929571277?network=testnet

Screenshot 2024-02-29 at 6 51 35 PM
Screenshot 2024-02-29 at 6 51 58 PM

### Test 5
Mint, transfer, new burn in 1 txn
Screenshot 2024-03-01 at 12 21 51 PM
Screenshot 2024-03-01 at 12 21 39 PM

### Test 6
Mint, transfer x2, new burn in 1 txn 
Screenshot 2024-03-01 at 12 22 57 PM
Screenshot 2024-03-01 at 12 22 43 PM

### DIFF
--- a/rust/processor/src/models/token_v2_models/v2_token_activities.rs
+++ b/rust/processor/src/models/token_v2_models/v2_token_activities.rs
@@ -7,7 +7,6 @@
 
 use super::{
     v2_token_datas::TokenDataV2,
-    v2_token_ownerships::CurrentTokenOwnershipV2Query,
     v2_token_utils::{TokenStandard, TokenV2Minted, V2TokenEvent},
 };
 use crate::{
@@ -155,8 +154,6 @@ impl TokenActivityV2 {
         entry_function_id_str: &Option<String>,
         token_v2_metadata: &ObjectAggregatedDataMapping,
         tokens_minted: &TokenV2Minted,
-        // needed to find owner of the burnt nft
-        conn: &mut PgPoolConnection<'_>,
     ) -> anyhow::Result<Option<Self>> {
         let event_type = event.type_str.clone();
         if let Some(token_event) =
@@ -250,28 +247,8 @@ impl TokenActivityV2 {
                     transaction_timestamp: txn_timestamp,
                 }));
             }
-            // This should only happen in the case where we have a burn event and mint event in the same
-            // transaction. Basically token doesn't exist in metadata but there's a mint event
-            if tokens_minted.get(&token_data_id).is_some() {
-                return Ok(Some(Self {
-                    transaction_version: txn_version,
-                    event_index,
-                    event_account_address,
-                    token_data_id,
-                    property_version_v1: BigDecimal::zero(),
-                    type_: event_type,
-                    from_address: owner_address.clone(),
-                    to_address: None,
-                    token_amount: BigDecimal::one(),
-                    before_value: None,
-                    after_value: None,
-                    entry_function_id_str: entry_function_id_str.clone(),
-                    token_standard: TokenStandard::V2.to_string(),
-                    is_fungible_v2: Some(false),
-                    transaction_timestamp: txn_timestamp,
-                }));
-            }
-            // Otherwise the burn happened previously and we have to partially fill the row
+            // If the object metadata isn't found in the transaction, then it's a burn event.
+            // This should only happen in the case when the token is minted and burnt in the same transaction.
             return Ok(Some(Self {
                 transaction_version: txn_version,
                 event_index,
@@ -279,7 +256,7 @@ impl TokenActivityV2 {
                 token_data_id,
                 property_version_v1: BigDecimal::zero(),
                 type_: event_type,
-                from_address: owner_address,
+                from_address: owner_address.clone(),
                 to_address: None,
                 token_amount: BigDecimal::one(),
                 before_value: None,
--- a/rust/processor/src/models/token_v2_models/v2_token_activities.rs
+++ b/rust/processor/src/models/token_v2_models/v2_token_activities.rs
@@ -173,6 +173,12 @@ impl TokenActivityV2 {
                 V2TokenEvent::TransferEvent(inner) => inner.get_object_address(),
                 _ => event_account_address.clone(),
             };
+            // the new burn event has owner address now!
+            let owner_address = if let V2TokenEvent::Burn(inner) = token_event {
+                Some(inner.get_previous_owner_address())
+            } else {
+                None
+            };
 
             if let Some(metadata) = token_v2_metadata.get(&token_data_id) {
                 let object_core = &metadata.object.object_core;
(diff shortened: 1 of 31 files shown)


igor-aptos (approved):

rust/processor/src/models/token_v2_models/v2_token_activities.rs
```diff
@@ -303,6 +271,24 @@ impl TokenActivityV2 {
                     transaction_timestamp: txn_timestamp,
                 }));
             }
+            // Otherwise the burn happened previously and we have to partially fill the row
```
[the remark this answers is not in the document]
rtso: could we combine this and the above 
```if tokens_minted.get(&token_data_id).is_some() {```

File: rust/processor/src/models/token_v2_models/v2_token_ownerships.rs [hunk 2]
```diff
@@ -294,69 +294,78 @@ impl TokenOwnershipV2 {
...
                     token_data_id: token_data_id.clone(),
                     property_version_v1: BigDecimal::zero(),
-                    owner_address: Some(owner_address.clone()),
+                    owner_address: Some(previous_owner.clone()),
                     storage_id: storage_id.clone(),
                     amount: BigDecimal::zero(),
                     table_type_v1: None,
                     token_properties_mutated_v1: None,
-                    is_soulbound_v2: is_soulbound,
+                    is_soulbound_v2: Some(false),
```
rtso, line 349: @bowenyang007 is this ok? or should we make this `None`?

rtso: Once we have better deletion handling, I guess this doesn't matter much

rust/processor/src/processors/token_v2_processor.rs
```diff
@@ -482,6 +482,7 @@ async fn parse_v2_token(
 
             // Get burn events for token v2 by object
             let mut tokens_burned: TokenV2Burned = AHashSet::new();
+            let mut tokens_burned_map: TokenV2BurnedMap = AHashMap::new();
```
bowenyang007, line 485: why do we need both?

rtso: the hashset includes both old and new burn events. we still need to know if token got burnt with the old event for lookup. the map is for only new burn events to get the previous owner.

rust/processor/src/models/token_v2_models/v2_token_activities.rs
```diff
@@ -243,66 +245,26 @@ impl TokenActivityV2 {
                     is_fungible_v2: Some(false),
                     transaction_timestamp: txn_timestamp,
                 }));
-            } else {
```
bowenyang007, line 246: Oops I think that we might still need this for backfilling?

rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
bowenyang007: actually should we just add deletion handling here as well?

rust/processor/src/models/token_v2_models/v2_token_activities.rs
rtso: yes let me add the `else` back in

rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
rtso: i'll give it a shot

rtso: actually let's leave deletion handling for later. there's some edge cases we need to think about, like if the row doesn't exist yet.

bowenyang007: this data is wrong though.

bowenyang007: The simplest thing to do is to have another field called burnt_transaction_version and update that field when the token is burnt. That way the existing queries would work as intended because you're not updating last_transaction-version

rtso: so the idea is to have to the deletion query handle the `amount` field and the insertion query to handle updating the rest of the fields. how do we set `amount = 1` the first time and not the other times?

rtso: we could split the insertion query into 2 queries but that would double the # of insertion queries

rtso: a race condition between the mint/transfer and burn insertions can happen and cause some fields to be wrong, but we sync'd offline that data accuracy for burnt tokens/objects is best effort so this is fine. gonna consider this resolved.

bowenyang007 (commented):
What if there's mint + burn. Or transfer + burn. Or transfer + transfer + burn in the same transaction? 

Will this error or I guess maybe not b/c we're no longer batching across tables?

bowenyang007 (commented):
This does need more testing though. Can we at least manually create a case where this could happen? Or replay in cases where we had bugs before.

rust/processor/src/models/token_v2_models/v2_token_activities.rs
```diff
@@ -244,55 +240,24 @@ impl TokenActivityV2 {
...
+                    // To handle a case with the old burn events, when a token is minted and burnt in the same transaction
+                    None
+                };
 
-                // This should only happen in the case where we have a burn event where the token is gone
-                // and the previous instance of the token wasn't in the batch. We need to look up in the db
-                let latest_nft_ownership =
-                    match CurrentTokenOwnershipV2Query::get_latest_owned_nft_by_token_data_id(
-                        conn,
-                        &token_data_id,
```
bowenyang007, line 274: wait why don't we need this query anymore for the old events?

rust/processor/src/processors/token_v2_processor.rs
bowenyang007: we need to add comments here, and it might be better to merge them. E.g. for v1 event we can just have the value be null. It's very confusing to have both.

rust/processor/src/models/token_v2_models/v2_token_activities.rs
rtso: for the old burn events, we set from_address to None. is that fine?

bowenyang007 (approved):

PR #297, opened 2024-02-29
[rust] add transaction size info

bowenyang007:
* add sizes for transaction, events, and write sets. 
* fixing upgrade to signature type from 1.10

## testing
image


msmouse (approved):
Thanks boss!

PR #299, opened 2024-03-01
Introduce txn filter

CapCap:
Allow filtering transactions that reach the processor right after GRPC via config

The criteria are combined with `AND`, if a criteria is not set, it is ignored

adds a config section:
```
server_config:
  transaction_filter:
    # Only allow transactions from these contract addresses
    focus_contract_addresses:
      - "0x0"
    # Skip transactions from these sender addresses
    skip_sender_addresses:
      - "0x0"
    # Skip all transactions that aren't user transactions
    focus_user_transactions: true
  ```

### DIFF
--- a/rust/processor/src/transaction_filter.rs
+++ b/rust/processor/src/transaction_filter.rs
@@ -1,0 +1,77 @@
+use aptos_protos::transaction::v1::{
+    transaction::{TransactionType, TxnData},
+    transaction_payload::Payload,
+    Transaction,
+};
+use serde::{Deserialize, Serialize};
+
+/// Allows filtering transactions based on various criteria
+/// The criteria are combined with `AND`
+/// If a criteria is not set, it is ignored
+/// Criteria will be loaded from the config file
+#[derive(Debug, Clone, Deserialize, Serialize)]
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: rust/processor/src/transaction_filter.rs [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+        if !is_user_txn {
+            return true;
+        }
+
+        if let Some(txn_data) = transaction.txn_data.as_ref() {
+            if let TxnData::User(user_transaction) = txn_data {
+                if let Some(utr) = user_transaction.request.as_ref() {
+                    // Skip if sender is in the skip list
+                    if let Some(skip_sender_addresses) = &self.skip_sender_addresses {
+                        return !skip_sender_addresses.contains(&utr.sender);
```
just-in-chang, line 54: shouldn't we also be checking focus_contract_address instead of immediately returning

larry-aptos: if false then return; otherwise continue the checking

CapCap: good call

larry-aptos (approved):

PR #303, opened 2024-03-05
Fix fungible asset metadata

rtso:
## Description 
* Use the correct metadata address to lookup whether an object is a fungible asset
* Fix current_object is_fungible_asset to check for existence of FungibleAssetMetadata, not FungibleAssetStore

## Testing
https://explorer.aptoslabs.com/txn/313811207/userTxnOverview?network=mainnet
Run objects_processor with version 313811207. See that is_fungible_asset is true. 
Screenshot 2024-03-05 at 4 10 50 PM
Screenshot 2024-03-05 at 4 16 44 PM

Run fungible_asset_processor with version 313811207. Fungible asset store is indexed
Screenshot 2024-03-05 at 4 32 55 PM

### DIFF
--- a/rust/processor/src/processors/objects_processor.rs
+++ b/rust/processor/src/processors/objects_processor.rs
@@ -4,7 +4,7 @@
 use super::{ProcessingResult, ProcessorName, ProcessorTrait};
 use crate::{
     models::{
-        fungible_asset_models::v2_fungible_asset_utils::FungibleAssetStore,
+        fungible_asset_models::v2_fungible_asset_utils::FungibleAssetMetadata,
         object_models::{
             v2_object_utils::{
                 ObjectAggregatedData, ObjectAggregatedDataMapping, ObjectWithMetadata,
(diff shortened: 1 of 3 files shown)


rust/processor/src/processors/objects_processor.rs
```diff
@@ -119,12 +119,6 @@ fn insert_current_objects_query(
             .on_conflict(object_address)
             .do_update()
             .set((
-                owner_address.eq(excluded(owner_address)),
-                state_key_hash.eq(excluded(state_key_hash)),
-                allow_ungated_transfer.eq(excluded(allow_ungated_transfer)),
-                last_guid_creation_num.eq(excluded(last_guid_creation_num)),
-                last_transaction_version.eq(excluded(last_transaction_version)),
-                is_deleted.eq(excluded(is_deleted)),
```
bowenyang007, line 127: you can't remove this. I meant the objects table

bowenyang007 (changes requested):
fix the on conflict insert

rust/processor/src/processors/objects_processor.rs
rtso: the objects table already has it set

bowenyang007 (approved):

PR #311, opened 2024-03-07
Remove dependency on objects processor

rtso:
## Description 
To improve dev ex, we remove the dependency of token v2 / fungible asset processor on objects processor. 

In `fungible_asset_metadata`, introduce a column `is_token_v2`. Fungible asset processor will lookup `is_token_v2` to skip indexing fungible tokens. 

Token v2 processor will lookup `current_token_datas_v2.is_fungible_v2` to index the fungible token. 

## Test

Fungible token with metadata in same txn. https://explorer.aptoslabs.com/txn/351790067/userTxnOverview?network=mainnet
See that the fungible token is indexed in the tokens table and not the fungible balances table. 
Screenshot 2024-03-07 at 6 46 06 PM

Normal fungible asset indexing still works
Screenshot 2024-03-07 at 7 02 18 PM

Normal token indexing still works
Screenshot 2024-03-07 at 7 04 08 PM

Objects indexing still works
Screenshot 2024-03-07 at 7 12 49 PM
Screenshot 2024-03-07 at 7 12 45 PM

### DIFF
--- a/rust/processor/migrations/2024-03-07-224504_fungible_asset_metadata_is_token_v2/up.sql
+++ b/rust/processor/migrations/2024-03-07-224504_fungible_asset_metadata_is_token_v2/up.sql
@@ -1,0 +1,9 @@
+-- Your SQL goes here
+ALTER TABLE fungible_asset_metadata
+ADD COLUMN IF NOT EXISTS is_token_v2 BOOLEAN;
+ALTER TABLE objects
+DROP COLUMN IF EXISTS is_fungible_asset,
+DROP COLUMN IF EXISTS is_token;
+ALTER TABLE current_objects
+DROP COLUMN IF EXISTS is_fungible_asset,
+DROP COLUMN IF EXISTS is_token;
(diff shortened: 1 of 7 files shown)


rust/processor/migrations/2024-03-07-224504_fungible_asset_metadata_is_token_v2/up.sql
```diff
@@ -0,0 +1,9 @@
+-- Your SQL goes here
+ALTER TABLE fungible_asset_metadata
+ADD COLUMN IF NOT EXISTS is_token_v2 BOOLEAN;
+ALTER TABLE objects
+DROP COLUMN IF EXISTS is_fungible_asset,
+DROP COLUMN IF EXISTS is_token;
+ALTER TABLE current_objects
+DROP COLUMN IF EXISTS is_fungible_asset,
+DROP COLUMN IF EXISTS is_token;
```
bowenyang007, line 9: do we need to remove these? Maybe we keep it?

bowenyang007: Up to you tho.

bowenyang007 (approved):

rust/processor/migrations/2024-03-07-224504_fungible_asset_metadata_is_token_v2/up.sql
rtso: idk when someone would ever use this. they can just look at the fungible and token tables. and this is gonna change once we have a fungible token standard.

PR #323, opened 2024-03-12
[EI-347] Version tracking in a parallel processing world

rtso:
## Description 
With all the recent improvements we've been making, tracking the latest version processed is more difficult.
- Now we have multiple db writer tasks writing to different tables. For a batch of txn's, how do we know all tables that should be written to have been written to?
- When a batch of txn's are filtered out, we still need to mark those txn's as processed.

The proposal in this PR is to have a `VersionTracker` that receives VersionTrackerItem, which can be either type
- `CompleteBatch`: version tracker can mark this batch as completely processed 
- `PartialBatch`: Internally, `VersionTracker` counts the number of times it receives a `PartialBatch`. Once the count of partial batches reaches the number of tables we expect to be written, `VersionTracker` can mark the versions as completely processed. The `config.yaml` will need to set `number_db_tables`. 

## Testing 
Run user_transaction_processor with `number_db_tables: 2`. Check that 2 tables have been written to and processor_status is updated. 
Run cargo test


rust/processor/src/config.rs
```diff
@@ -30,15 +29,15 @@ pub struct IndexerGrpcProcessorConfig {
     pub starting_version: Option<u64>,
     // Version to end indexing at
     pub ending_version: Option<u64>,
+    // Number of tables this processor writes to. This is used in the version tracker.
+    // Since DB table writes are parallel, it needs to know when all tables have been written to before it can move the latest version forward.
+    pub number_db_tables: u64,
```
rtso, line 34: Having this requirement in config.yaml is not great. Thoughts on requiring a mapping for tablename -> # queries in the `ProcessorTrait`? On initialization, `VersionTracker` can count the total # queries per processor, excluding the ones in `skip_tables`.

bowenyang007 (changes requested):
This looks stale. Pushing back to your queue

PR #327, opened 2024-03-20
[rust] handle burn properly for partial burns

bowenyang007:
## Summary
This closes the loop on the messy logic that is burning NFTs: 
There could be a complete burn or a partial burn.
* In the complete burn case, we don’t have anything so we either have the previous_owner field from the new burn event, or have to do a db lookup
* In the partial burn case, we assumed (wrongly) that ObjectCore will be there.

This PR fixes the partial burn case. 

## Backfill
First burn event 
Testnet: 835574507
Mainnet: 151521169

## Testing
NFT burnt!
image

### DIFF
--- a/rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
+++ b/rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
@@ -339,19 +339,20 @@ impl TokenOwnershipV2 {
         query_retries: u32,
         query_retry_delay_ms: u64,
     ) -> anyhow::Result<Option<(Self, CurrentTokenOwnershipV2)>> {
-        if let Some(burn_event) = tokens_burned.get(token_address) {
+        let token_address = standardize_address(token_address);
+        if let Some(burn_event) = tokens_burned.get(&token_address) {
             // 1. Try to lookup token address in burn event mapping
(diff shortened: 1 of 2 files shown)


rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
```diff
@@ -296,20 +339,19 @@ impl TokenOwnershipV2 {
         query_retries: u32,
         query_retry_delay_ms: u64,
     ) -> anyhow::Result<Option<(Self, CurrentTokenOwnershipV2)>> {
-        let token_address = standardize_address(&write_resource.address.to_string());
-        if let Some(burn_event) = tokens_burned.get(&token_address) {
+        if let Some(burn_event) = tokens_burned.get(token_address) {
```
just-in-chang, line 342: Maybe call `standardize_address()` on `tokens_burned` here too in case we use the helper fn anywhere else in the future

just-in-chang (approved):

rust/processor/src/models/token_v2_models/v2_token_ownerships.rs
bowenyang007: ok yea good poitn.

PR #330, opened 2024-03-24
[hotfix] fix panic where coin type is too long

bowenyang007:
## Summary
If coin type is too long so we’re not tracking it under the new logic (> 1K length). But the lookup logic was still there so it caused some issues. 

This skips cases where the mapping isn't there. 

Also this only applies to coin v1.

## Test
No longer panics


CapCap (approved):
Is this dangerous in other ways/places?

bowenyang007:
> Is this dangerous in other ways/places?

We wanted to make this super safe, but so far we've never seen this fail (been like 6month+ so I think that this panic was overkill in the first place).

Issue #219, opened 2023-12-14
[Custom Processors] Add variant starting_version field to config for when a version is already in the DB


banool:
## Description
Currently we have only one `starting_version` field. This field applies unconditionally, even if there is a version in the DB. This means if the user wants to start from the version in the DB if present, but otherwise start from a given version, there is no way for them to do that without a tricky order of starts and stops with adding and removing `starting_version` from the config.

We should have two fields (more concise naming TBD):
- `starting_version_if_nothing_in_db`: If given, start from this version if nothing is in the DB.
- `starting_version_no_matter_what`: Start from this version even if there is something in DB.

PR #332, opened 2024-03-25
Refactor indexer version config

keyliaran:
Resolve issue https://github.com/aptos-labs/aptos-indexer-processors/issues/219.
Moved the configuration for indexer starting/ending versions into a separate entity. Additionally, a flag 'override_starting' has been introduced to enforce the use of the starting version from the configuration, even if 'starting_version' exists in the database.


larry-aptos:
nice work! 

one thing to point out is we're trying to avoid breaking changes introduced to our indexer stack. 

* community needs to change their processor config, otherwise crash due to`deny_unknown_fields`

is it possible to add a new optional field like `overriding_starting_version`(default to true) and if it's present and true or not present, we'll keep the existing behavior and if false, do the new logic.

keyliaran:
@larry-aptos I've implemented support for backward compatibility in the configuration. Please take a look.

rust/processor/README.md
```diff
@@ -25,8 +25,10 @@ Indexer GRPC parser is to indexer data processor that leverages the indexer grpc
         indexer_grpc_http2_ping_timeout_in_secs: 10
         number_concurrent_processing_tasks: 10
         auth_token: AUTH_TOKEN
-        starting_version: 0 # optional
```
larry-aptos, line 28: nit: update the document

keyliaran: Done

larry-aptos (approved):

rust/processor/README.md
larry-aptos: sorry for being late on this.

a question about this: if `starting_version` is specific, does it also mean that processing start from this version regardless of db version?

keyliaran: I've added a new field called "override_starting_version." The logic is straightforward:
1) If "override_starting_version" exists and "starting_version" is configured, use it regardless of the database version.
2) If "override_starting_version" doesn't exist, use the database version.

rtso (commented):
@keyliaran the current implementation is not backwards compatible with how the ecosystem uses `starting_version`. The current assumption is `starting_version` will override whatever is it DB. We should keep that assumption so we don't break ecosystem indexers. I propose to add a `starting_version_if_nothing_in_db` instead of the override.

keyliaran:
> @keyliaran the current implementation is not backwards compatible with how the ecosystem uses `starting_version`. The current assumption is `starting_version` will override whatever is it DB. We should keep that assumption so we don't break ecosystem indexers. I propose to add a `starting_version_if_nothing_in_db` instead of the override.

If the indexer is launched with starting_version set in the configuration, it will always resume from that version after each restart due to node crashes or other issues like panics. This behavior isn't ideal because, i guess, nobody wants to keep reindexing the entire blockchain from starting_version setted in config. To address this, I would need to start the indexer from the specified starting_version and then modify the configuration.

To improve this, you're suggesting adding a new field called starting_version_if_nothing_in_db. Here's how it would work: starting_version will override both the database value and the config value of starting_version_if_nothing_in_db. This means that if the database doesn't have a starting version stored, it will use starting_version from the config. So there are no usefull usecases of new field starting_version_if_nothing_in_db, it will be used only if there is nothing in database

keyliaran:
@rtso Pls take a look to the updated implementation. It shoudln't break backward compatible.

rtso:
@keyliaran I think the issue is originally suggesting this approach:
* If there's no latest version set in the db, it will use `starting_version_if_nothing_in_db` if set
* Otherwise it will use `starting_version` if set
* Otherwise, it will use the latest version in db

This makes it easier to develop the indexer. When you wipe the db, the config remembers the version the contract was deployed and start indexing from that version. If you need to start indexing from specific version, you can set an arbitrary version in `starting_version`. You won't have to set/unset the flag and `starting_version` every time.

rtso:
> starting_version_if_nothing_in_db, it will be used only if there is nothing in database

Yep you're absolutely correct! And if you want it to start from latest version, leave `starting_version` blank so that it starts from latest version in db.

keyliaran:
@rtso Made an update, pls take a look

rtso:
Could you also share your test plan in the description? TY

rust/processor/src/config.rs
```diff
@@ -27,6 +27,8 @@ pub struct IndexerGrpcProcessorConfig {
     pub auth_token: String,
     // Version to start indexing from
     pub starting_version: Option<u64>,
+    // Version to start indexing from if nothing in db
+    pub starting_version_if_nothing_in_db: Option<u64>,
```
grao1991, line 31: we probably want a better name..

also, for these two parameters, it's not clear which one will take precedence. we probably should either use an enum here, or ensure they are mutually exclusive.

rust/processor/src/worker.rs
```diff
@@ -141,27 +144,33 @@ impl Worker {
...
+            None => {
                 info!(
                     processor_name = processor_name,
                     service_type = PROCESSOR_SERVICE_TYPE,
                     "[Parser] No starting version from db so starting from version 0"
                 );
-                0
-            });
-
-        let starting_version = self.starting_version.unwrap_or(starting_version_from_db);
```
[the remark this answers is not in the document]
keyliaran: Sure! Fixed

rust/processor/src/config.rs
keyliaran: I believe we should avoid using enums to maintain backward compatibility. The current implementation works as follows:
- If starting_version exists in the config, it will override all existing values (both starting_version_if_nothing_in_db and the database value).
- If starting_version is not configured, we'll take a look to the database value. If the database has no value and starting_version_if_nothing_in_db is configured, we'll use that value; otherwise, we'll default to 0.

PR #333, opened 2024-03-25
coin_processor optimization: compile regex once

keyliaran:
Currently, the regular expression is compiled anew with each iteration of the loop. This compilation process is resource-intensive, and there's no built-in caching mechanism. Consider utilizing lazy_static to cache the compiled result for optimization.

### DIFF
--- a/rust/processor/src/models/coin_models/coin_utils.rs
+++ b/rust/processor/src/models/coin_models/coin_utils.rs
@@ -163,9 +163,7 @@ pub struct CoinInfoType {
     creator_address: String,
 }
 
-static RE: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r"(<(.*)>)").unwrap()
-});
+static RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"(<(.*)>)").unwrap());
 
 static COIN_RESOURCES: Lazy<[String; 2]> = Lazy::new(|| {
     [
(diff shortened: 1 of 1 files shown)


larry-aptos:
looks good! can you add how do you test this change? ty

keyliaran:
> looks good! can you add how do you test this change? ty

Currently, there are no unit tests for the indexer. I've conducted tests by running both the old and new versions on the same batch of events and confirmed that the resulting databases are identical.

larry-aptos (approved):

banool:
Hey thanks for the PR! Could you use once_cell::Lazy instead?

keyliaran:
> Hey thanks for the PR! Could you use once_cell::Lazy instead? I believe it is recommended these days over lazy_static, better API / types rather than macro magic.

Hi! Made an update, pls take a look!

larry-aptos (approved):

rust/Cargo.lock
just-in-chang, line 1: Hey! Mind removing the Cargo.lock changes from the PR? There aren't any changes to the Cargo.toml so it shouldn't be necessary to include

keyliaran: Done

grao1991 (approved):

keyliaran:
> Great! It seems CoinBalance::from_write_resource consumes 5x less CPU with your change. Can you fix the lint and have your PR merged?

Sure, fixed

PR #335, opened 2024-03-27
Compatibility Check Workflow

larry-aptos:
* Goal: if new `enum` variant or `oneof` field is added, this check will break. 
* The data quality/correctness will be checked separately. 

### Testing Plan

When new enum variant or oneof field is added, workflow failure:

image


yuunlimm (approved):
looks good!

PR #336, opened 2024-03-29
Add CPU profiling support.

grao1991:
Serves the CPU flamegraph at "host:health_check_port/profilez" page.

It took 10s to collect samples, and you will see the flamegraph afterwards.


larry-aptos (approved):

CapCap (approved):
Linux only? LGTM, but please add PR description and ideally link to the "how to use
 notion

grao1991:
> Linux only? LGTM, but please add PR description and ideally link to the "how to use notion

Link added..

I'm too lazy to figure out a way for non-linux :)

PR #337, opened 2024-04-02
Change the config structure.

grao1991:
Change the config to support various features, including:

- running multiple processors together in a server
- providing multiple db connection urls

etc.

Implementation will be provided in future PRs.


bowenyang007 (changes requested):
wait why are we doing this? And what's the potential downside of having multiple services? 

If this is for a particular test, can we make a custom processor for this? Instead of making this the new normal.

grao1991:
> wait why are we doing this? And what's the potential downside of having multiple services?
> 
> If this is for a particular test, can we make a custom processor for this? Instead of making this the new normal.

There are people asking for this feature. It could save a lot of bandwidth and resource since multiple processor can share the same fetcher.

bowenyang007 (changes requested):
Pushing back to your queue. This looks like it needs to be synced with the SDK workstream. Let's have online meetings about changes that affect the UX directly.

PR #346, opened 2024-04-15
Cleanup post processor

rtso:
Metrics are not used anymore since we switched to using txn timestamps everywhere in Grafana


larry-aptos (approved):

just-in-chang (approved):
Would we ever need this in the future? Could just keep the code here and remove the deployment

PR #350, opened 2024-04-17
Refactor the code to make it easier to support multiple db backends in the future.

grao1991:
Restructure the directories.
Rename several variables to remove "Pg".
Remove some unused methods.

Tested on the local machine, verified both manually running migration and running migration through rust code can work.


bowenyang007 (dismissed):
There's a problem. Running migrations manually the old way anymore. We have to cd into the new folder to run migrations. It also generates schema in a different location. Can you actually try it and update the readme? Also please add test plan.

grao1991:
> There's a problem. Running migrations manually the old way anymore. We have to cd into the new folder to run migrations. It also generates schema in a different location. Can you actually try it and update the readme? Also please add test plan.

Tested and added one line in README.

CapCap (approved):

rust/processor/src/db/common/mod.rs
CapCap, line 1: please make this `pub`

rust/processor/src/db/mod.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod common;
```
CapCap, line 1: pub again, no? otherwise how will SDK use any of this

rust/processor/src/db/common/mod.rs
grao1991: Done

rust/processor/src/db/mod.rs
grao1991: Done

rust/processor/src/db/common/mod.rs
rtso: The command should be run inside `src/db/postgres/`

PR #359, opened 2024-04-30
[token] Add is_deleted to token datas (v2 nfts only)

bowenyang007:
## Summary
For context we want to be able to know if a token was deleted from the current_token_datas_v2 table. This only works for v2 tokens because v1 tokens can be fungible and non fungible interchangeably so it's impossible to get this field accurately. 

## Backfill
Testnet: 553151941
Mainnet: 199271122

## Testing
is_deleted_v2 in!
image


rtso:
How big of an effort is it to index `is_deleted` for v1 tokens?

bowenyang007:
> How big of an effort is it to index `is_deleted` for v1 tokens?

If we can disregard fungible v1 tokens, then we can do it. Actually it's a good point they might want for v1 as well.

rtso:
@bowenyang007 yeah just disregard fungible tokens because the logic will be separated after my PR anyway

bowenyang007:
@rtso not v1 fungible tokens lol. v1 fungible tokens look the same as v1 nfts.

bowenyang007:
Ok confirmed with stakeholders that we only need v2 for now

rust/processor/src/models/token_v2_models/v2_token_datas.rs
```diff
@@ -139,13 +140,78 @@ impl TokenDataV2 {
...
+                    is_deleted_v2: Some(false),
                 },
             )))
         } else {
             Ok(None)
         }
     }
 
+    /// This handles the case where token is burned but objectCore is still there
+    pub async fn get_burned_nft_v2_from_write_resource(
```
rtso, line 152: isn't this case already covered by `get_burned_nft_v2_from_delete_resource` below?

rust/processor/src/processors/token_v2_processor.rs
```diff
@@ -893,6 +945,22 @@ async fn parse_v2_token(
                             );
                         }
 
+                        // Add burned NFT handling for token datas (can probably be merged with below)
```
rtso, line 948: yeah can we merge this with below since it's doing the same thing

rust/processor/src/models/token_v2_models/v2_token_datas.rs
bowenyang007: no. this is write resource.

rust/processor/src/processors/token_v2_processor.rs
bowenyang007: the issue is that the code below is more complex than needed here b/c it needs previous owner as well. I'm hoping that as you guys look at sdk you can also rewrite this whole logic. The logic has kind of ballooned out of control a bit. For now I don't see a clean way to merge the paths so just leaving a comment for future.

rtso: ok i'll take a stab at refactoring this for sdk

rust/processor/src/models/token_v2_models/v2_token_datas.rs
rtso: when a token gets burnt won't it always have a delete resource?

bowenyang007: What if it's a partial burn and there are resources left?

rtso (approved):

rust/processor/src/models/token_v2_models/v2_token_datas.rs
rtso: won't the token resource get deleted so there will always be a delete resource?

bowenyang007: no that's not how it works. unless the resource group is entirely destroyed it'll emit a write resource event :(

PR #367, opened 2024-05-09
add grpc response item timeout.

larry-aptos:
* Add timeout for transaction fetch receiver.

## Testing Plan

* Tested in devnet and looks good

image

### DIFF
--- a/rust/processor/src/config.rs
+++ b/rust/processor/src/config.rs
@@ -70,9 +70,9 @@ impl IndexerGrpcProcessorConfig {
         100_000
     }
 
-    /// Default timeout for grpc response item in seconds. Defaults to 300 seconds.
+    /// Default timeout for grpc response item in seconds. Defaults to 15 seconds.
     pub const fn default_grpc_response_item_timeout_in_secs() -> u64 {
-        60 * 5
+        15
     }
 }
 
--- a/rust/processor/src/config.rs
+++ b/rust/processor/src/config.rs
@@ -44,8 +44,8 @@ pub struct IndexerGrpcProcessorConfig {
     pub per_table_chunk_sizes: AHashMap<String, usize>,
     pub enable_verbose_logging: Option<bool>,
 
-    #[serde(default = "IndexerGrpcProcessorConfig::default_grpc_response_item_timeout_ms")]
-    pub grpc_response_item_timeout_ms: u64,
+    #[serde(default = "IndexerGrpcProcessorConfig::default_grpc_response_item_timeout_in_secs")]
+    pub grpc_response_item_timeout_in_secs: u64,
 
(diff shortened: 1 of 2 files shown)


rust/processor/src/config.rs
```diff
@@ -43,6 +43,10 @@ pub struct IndexerGrpcProcessorConfig {
     #[serde(default = "AHashMap::new")]
     pub per_table_chunk_sizes: AHashMap<String, usize>,
     pub enable_verbose_logging: Option<bool>,
+
+    #[serde(default = "IndexerGrpcProcessorConfig::default_grpc_response_item_timeout_ms")]
+    pub grpc_response_item_timeout_ms: u64,
```
rtso, line 48: nit: make this in secs to be consistent with the other config values

rust/processor/src/config.rs
```diff
@@ -65,6 +69,11 @@ impl IndexerGrpcProcessorConfig {
     pub const fn default_pb_channel_txn_chunk_size() -> usize {
         100_000
     }
+
+    /// Default timeout for grpc response item in milliseconds. Defaults to 300 seconds.
+    pub const fn default_grpc_response_item_timeout_ms() -> u64 {
```
rtso, line 74: could we make this return a duration instead of seconds? this is in line with other config values [https://github.com/aptos-labs/aptos-indexer-processors/blob/41461d455dbeb1699159ab1dd91f30b9b074c5db/rust/processor/src/config.rs#L122](link)

larry-aptos: done done done. sorry for the delay 🤦

rust/processor/src/config.rs
```diff
@@ -65,6 +69,11 @@ impl IndexerGrpcProcessorConfig {
     pub const fn default_pb_channel_txn_chunk_size() -> usize {
         100_000
     }
+
+    /// Default timeout for grpc response item in seconds. Defaults to 300 seconds.
+    pub const fn default_grpc_response_item_timeout_in_secs() -> u64 {
+        60 * 5
```
rtso, line 75: can we decrease this timeout any more so the processors reconnect faster? message size limit on grpc side is 15MB

larry-aptos: true; this number is from previous setting.

15MB / 10mbps = 15 seconds

larry-aptos: updated.

rtso (approved):
code lgtm, can you update test plan and deploy to networks

rust/processor/src/config.rs
larry-aptos: 15 is too aggressive; we want to relax this condition considering network contention may happen.

PR #376, opened 2024-05-15
[EI-428] Auto reconnect grpc if txn stream times out waiting for response

rtso:
## Description 
Two changes:
1. Auto reconnect to grpc if it times out waiting for txn's from channel. Moving the timeout out of the processor thread to the fetcher loop thread.
2. When processor reaches ending version, it stops gracefully instead of panic

## Testing 
Happy path still works. 

Processor with auto-reconnection
Screenshot 2024-05-15 at 12 15 47 PM

Processor with ending version
Screenshot 2024-05-15 at 11 47 47 AM

Panics in the fetcher thread works.
Screenshot 2024-05-17 at 11 09 42 AM

Panic in processor thread works. 
Screenshot 2024-05-17 at 11 19 37 AM

### DIFF
--- a/rust/processor/src/grpc_stream.rs
+++ b/rust/processor/src/grpc_stream.rs
@@ -308,6 +308,7 @@ pub async fn create_fetcher_loop(
     indexer_grpc_http2_ping_interval: Duration,
     indexer_grpc_http2_ping_timeout: Duration,
     indexer_grpc_reconnection_timeout_secs: Duration,
+    indexer_grpc_response_item_timeout_secs: Duration,
     starting_version: u64,
     request_ending_version: Option<u64>,
     auth_token: String,
@@ -358,8 +359,13 @@ pub async fn create_fetcher_loop(
     let mut send_ma = MovingAverage::new(3000);
 
     loop {
-        let is_success = match resp_stream.next().await {
-            Some(Ok(mut r)) => {
+        let is_success = match tokio::time::timeout(
+            indexer_grpc_response_item_timeout_secs,
+            resp_stream.next(),
+        )
+        .await
+        {
+            Ok(Some(Ok(mut r))) => {
                 reconnection_retries = 0;
                 let start_version = r.transactions.as_slice().first().unwrap().version;
                 let start_txn_timestamp =
@@ -540,7 +546,7 @@ pub async fn create_fetcher_loop(
                     .inc_by(num_filtered_txns as u64);
                 true
             },
-            Some(Err(rpc_error)) => {
+            Ok(Some(Err(rpc_error))) => {
                 tracing::warn!(
                     processor_name = processor_name,
                     service_type = crate::worker::PROCESSOR_SERVICE_TYPE,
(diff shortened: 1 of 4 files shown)


bowenyang007:
what if the processor thread is stuck? we've seen that in one of the SEVs before.

rtso:
@bowenyang007 iirc, it was stuck because the fetcher channel had no items, indicating we didn't receive any txn's from the stream, so the stream itself was stuck.

rust/server-framework/src/lib.rs
```diff
@@ -55,8 +55,8 @@ where
             bail!("Probes and metrics handler unexpectedly exited");
         },
         _ = main_task_handler => {
-            error!("Main task unexpectedly exited");
```
bowenyang007, line 58: are you sure this is OK to remove? I remember that at one point we didn't have proper error exiting and the panics just silently failed. Did you test this?

bowenyang007:
> @bowenyang007 iirc, it was stuck because the fetcher channel had no items, indicating the fetcher thread didn't receive any txn's from the stream, so the grpc stream itself was stuck

No we did have another sev where the fetcher channel was full and the processor channel was empty for some reason.

rust/processor/src/grpc_stream.rs
```diff
@@ -358,8 +359,13 @@ pub async fn create_fetcher_loop(
...
     loop {
-        let is_success = match resp_stream.next().await {
-            Some(Ok(mut r)) => {
+        let is_success = match tokio::time::timeout(
+            indexer_grpc_response_item_timeout_secs,
+            resp_stream.next(),
+        )
+        .await
+        {
+            Ok(Some(Ok(mut r))) => {
```
CapCap, line 368: could you please add comments as to what state each of these rows represents? Could also break out into a nested match, up to you

rust/server-framework/src/lib.rs
CapCap: agreed this change seems dangerous at first glance

rtso: This change is so the when ending_version is reached, the processor stops gracefully and doesn't log it as an error. 

Verified that panic in the fetcher thread and processor thread still work.
Screenshot 2024-05-17 at 11 19 37 AM
Screenshot 2024-05-17 at 11 09 42 AM

bowenyang007: Can you try to crash something like in the processors and add to test plan?

rtso: I added a panic! in the processors thread and it exited with error. Let me include the screenshots in test plan.

CapCap: I think we can do something like this (pls verify actual behavior of `JoinHandle` first), instead of throwing it away
```
        res = main_task_handler => {
            res.expect("Main task failed").expect("Main task failed");
        },
```

CapCap: dis https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html

rtso:
@bowenyang007 that one was because of db inserts taking too long and can be done as a followup oncall task

bowenyang007 (approved):
Let's publish a release for okx.

PR #381, opened 2024-05-22
[EI-461] Index royalty in royalty tables

rtso:
## Description 
* Separate royalty to`current_token_royalty_v1` table (note that v2 royalty requires collection id which isn't currently supported in this PR)
* Move the indexing to `token_v2_processor`

## Backfill
Testnet: 0
Mainnet: 0

## Test Plan
Mainnet transaction 613858541
Screenshot 2024-05-21 at 5 48 17 PM

Screenshot 2024-05-21 at 5 48 06 PM
Screenshot 2024-05-21 at 5 47 34 PM

### DIFF
--- a/rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
+++ b/rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
@@ -1,19 +1,23 @@
 -- Your SQL goes here
 CREATE TABLE token_royalty (
     transaction_version BIGINT NOT NULL,
-    token_data_id VARCHAR(64) NOT NULL,
-    payee_address VARCHAR(64) NOT NULL,
-    royalty_points_numerator_v1 INT NOT NULL,
-    royalty_points_denominator_v1 INT NOT NULL,
+    token_data_id VARCHAR(66) NOT NULL,
+    payee_address VARCHAR(66) NOT NULL,
+    royalty_points_numerator NUMERIC NOT NULL,
+    royalty_points_denominator NUMERIC NOT NULL,
     token_standard TEXT NOT NULL,
-    (transaction_version, token_data_id) PRIMARY KEY
+    inserted_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (transaction_version, token_data_id) 
 );
 
 CREATE TABLE current_token_royalty (
-    token_data_id VARCHAR(64) NOT NULL,
-    payee_address VARCHAR(64) NOT NULL,
-    royalty_points_numerator_v1 INT NOT NULL,
-    royalty_points_denominator_v1 INT NOT NULL,
+    token_data_id VARCHAR(66) NOT NULL,
+    payee_address VARCHAR(66) NOT NULL,
+    royalty_points_numerator NUMERIC NOT NULL,
+    royalty_points_denominator NUMERIC NOT NULL,
     token_standard TEXT NOT NULL,
-    token_data_id PRIMARY KEY
+    last_transaction_version BIGINT NOT NULL,
+    last_transaction_timestamp TIMESTAMP NOT NULL,
+    inserted_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (token_data_id)
 );
--- a/rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
+++ b/rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
@@ -1,24 +1,11 @@
 -- Your SQL goes here
-CREATE TABLE token_royalty (
-    transaction_version BIGINT NOT NULL,
-    token_data_id VARCHAR(66) NOT NULL,
+-- This'll only work with royalty v1 because royalty_v2 requires collection id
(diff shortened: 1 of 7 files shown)


rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
```diff
@@ -0,0 +1,23 @@
...
+    payee_address VARCHAR(66) NOT NULL,
+    royalty_points_numerator NUMERIC NOT NULL,
+    royalty_points_denominator NUMERIC NOT NULL,
+    token_standard TEXT NOT NULL,
+    inserted_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (transaction_version, token_data_id) 
+);
+
+CREATE TABLE current_token_royalty (
+    token_data_id VARCHAR(66) NOT NULL,
```
larry-aptos, line 14: I forgot why we don't have a unique id for the token.. but using hash value along as primary key can be dangerous since collision may happen.

rtso: That's a good point. Including `creator_address` in the primary key can help with collision. Let me add that.

CapCap (approved):

rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
```diff
@@ -0,0 +1,24 @@
...
+    payee_address VARCHAR(66) NOT NULL,
+    royalty_points_numerator NUMERIC NOT NULL,
+    royalty_points_denominator NUMERIC NOT NULL,
+    token_standard TEXT NOT NULL,
+    inserted_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (transaction_version, token_data_id) 
+);
+
+CREATE TABLE current_token_royalty (
```
bowenyang007, line 13: do we need any indices here?

bowenyang007: wait...where is collection_id. we need that to get the join key.

rtso: it's only going to join with `current_token_datas.token_data_id`

rtso: `token_data_id` is already part of the PK so no additional index needed

grao1991: I'd suggest not adding unnecessary columns as part of PK. 256 bits hash is used everywhere and assumes no collision.

rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
rtso: @larry-aptos explained to me how collision can happen, but i forgot the reason cc @grao1991

larry-aptos: hash id is not guaranteed to be unique. It may be a problem when two different entities have the same hash value.

it seems ahash uses 64bit and considering 10million items, we get this

image

If we consider the number of items(i.e. token) can go up to 10m, we should take the probability of collision into account and add a field to make pk unique, otherwise, we should be fine.

rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
bowenyang007: This won't work for v2 though. Let's rename the table to v1 then.

rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
```diff
@@ -0,0 +1,24 @@
+-- Your SQL goes here
+CREATE TABLE token_royalty (
```
bowenyang007, line 2: Given we're moving away from tracking historical I don't think that we need this table.

rust/processor/migrations/2024-05-21-221101_add_royalty_v1/up.sql
bowenyang007: wait what? these hashes aren't randomly generated though so can collision still happen? I don't think that creator address is necessary b/c it's part of the hash.

bowenyang007: Also the tokens tables don't have this level of protection so we'll end up with issues anyway.

rust/processor/src/processors/token_v2_processor.rs
```diff
@@ -1127,6 +1188,7 @@ async fn parse_v2_token(
                 &b.storage_id,
             ))
     });
+    current_token_royalties_v1.sort();
```
rtso, line 1191: Does it automatically know to sort by token_data_id (pk)?

bowenyang007: yea cuz I added the sort function remember. that was your idea.

rtso:
New changes lgtm. Thanks @bowenyang007 !

PR #385, opened 2024-05-22
[EI-468] Create v1 migration views

rtso:
## Description
As part of the v1 table deprecation, replace the v1 tables with views using the v2 tables. This will give the ecosystem a migration window to migrate their queries to the v2 tables and reduces ecosystem impact. 

## Test plan
Ran the SQL locally


rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,327 @@
...
+    SELECT
+        tav.transaction_version,
+        event_account_address,
+        -- is the below missing an issue?
+        0 as event_creation_number,
+        0 as event_sequence_number,
+        tdv.collection_id  as collection_data_id_hash,
+        substring(tav.token_data_id, 3) as token_data_id_hash,
+        property_version_v1,
+        "type",
```
ying-w, line 96: "type" as transfer_type,

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,327 @@
...
+CREATE OR REPLACE VIEW legacy_migration_v1.token_activities AS
+    SELECT
+        tav.transaction_version,
+        event_account_address,
+        -- is the below missing an issue?
+        0 as event_creation_number,
+        0 as event_sequence_number,
+        tdv.collection_id  as collection_data_id_hash,
+        substring(tav.token_data_id, 3) as token_data_id_hash,
+        property_version_v1,
```
ying-w, line 95: `property_version_v1 as property_version`

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,327 @@
...
+    FROM public.fungible_asset_balances
+    WHERE token_standard = 'v1'
+    ;
+-- CREATE INDEX lm1_ccb_ct_a_index ON legacy_migration_v1.current_coin_balances USING btree (coin_type, amount)
+-- CREATE INDEX lm1_ccb_insat_index ON legacy_migration_v1.current_coin_balances USING btree (inserted_at)
+-- CREATE INDEX lm1_ccb_oa_ct_index ON legacy_migration_v1.current_coin_balances USING btree (owner_address, coin_type)
+
+-- replace `token_activities` with `token_activities_v2`
+-- token_activities_v2.token_data_id is 0x prefixed, but token_activities.token_data_id is not. We need to create an index on the substring
+CREATE OR REPLACE VIEW legacy_migration_v1.token_activities AS
```
ying-w, line 86: missing the following columns from token_activities

1. creator_address
2. collection_name
3. name

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,327 @@
...
+        token_uri AS metadata_uri,
+        -- Join on new royalty table
+        payee_address,
+        royalty_points_numerator,
+        royalty_points_denominator,
+        TRUE AS maximum_mutable,
+        TRUE AS uri_mutable,
+        TRUE AS description_mutable,
+        TRUE AS properties_mutable,
+        TRUE AS royalty_mutable,
```
ying-w, line 216: This is not always TRUE

```
SELECT 
maximum_mutable,
uri_mutable,
description_mutable,
properties_mutable,
royalty_mutable,
COUNT(1)
FROM current_token_datas
GROUP BY 1,2,3,4,5
ORDER BY 1,2,3,4,5
```

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,327 @@
...
+-- CREATE INDEX lm1_ca_oa_ct_at_index ON legacy_migration_v1.coin_activities USING btree (owner_address, coin_type, activity_type, amount)
+-- CREATE INDEX lm1_ca_oa_igf_index ON legacy_migration_v1.coin_activities USING btree (owner_address, is_gas_fee)
+
+-- replace `coin_balances` with `fungible_asset_balances`
+CREATE OR REPLACE VIEW legacy_migration_v1.coin_balances AS
+    SELECT
+        transaction_version,
+        owner_address,
+        -- this is mainly for hashing the coin type for primary key
+        substring(asset_type, 66) as coin_type_hash,
```
ying-w, line 53: why are we truncating this? you can have `0x1::aptos_coin::AptosCoin` as asset_type
we should note somewhere that coin_balances.coin_type_hash is going to change

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,327 @@
...
+-- CREATE INDEX lm1_ta_addr_coll_name_pv_index ON legacy_migration_v1.token_activities USING btree (creator_address, collection_name, name, property_version)
+-- CREATE INDEX lm1_ta_from_ttyp_index ON legacy_migration_v1.token_activities USING btree (from_address, transfer_type)
+-- CREATE INDEX lm1_ta_insat_index ON legacy_migration_v1.token_activities USING btree (inserted_at)
+-- CREATE INDEX lm1_ta_tdih_pv_index ON legacy_migration_v1.token_activities USING btree (token_data_id_hash, property_version)
+-- CREATE INDEX lm1_ta_to_ttyp_index ON legacy_migration_v1.token_activities USING btree (to_address, transfer_type)
+
+-- replace `token_ownerships` with `token_ownerships_v2`
+CREATE OR REPLACE VIEW legacy_migration_v1.token_ownerships AS 
+    SELECT
+        tov.token_data_id AS token_data_id_hash,
```
ying-w, line 122: substring?

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
ying-w: actually this will do the mapping
```sql
encode(sha256(asset_type::bytea), 'hex')
```

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
rtso: @bowenyang007 validated that most of these are true, so it's fine to hardcode it to avoid having to migrate the v1 logic over to the v2 processor

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,319 @@
...
+        substring(asset_type, 64) as coin_type_hash,
+        asset_type as coin_type,
+        amount,
+        transaction_version as last_transaction_version,
+        transaction_timestamp as last_transaction_timestamp,
+        inserted_at
+    FROM public.current_fungible_asset_balances
+    WHERE token_standard = 'v1'
+    ;
+CREATE INDEX lm1_ccb_ct_a_index ON public.current_fungible_asset_balances USING btree (asset_type, amount)
```
CapCap, line 72: surprising this didn't already exist

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,319 @@
...
+CREATE INDEX lm1_ca_oa_ct_at_index ON public.fungible_asset_activities USING btree (owner_address, asset_type, "type", amount)
+CREATE INDEX lm1_ca_oa_igf_index ON public.fungible_asset_activities USING btree (owner_address, is_gas_fee)
+
+-- replace `coin_balances` with `fungible_asset_balances`
+CREATE OR REPLACE VIEW legacy_migration_v1.coin_balances AS
+    SELECT
+        transaction_version,
+        owner_address,
+        -- this is mainly for hashing the coin type for primary key
+        encode(sha256(asset_type::bytea), 'hex') as coin_type_hash,
```
CapCap, line 49: do we actually have sha256 support enabled in alloy?

CapCap: Is the intention to use this as a join? Because if so need to add this as an index on the underlying table

CapCap (approved):
a few small concerns, but otherwise lgtm

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
CapCap: what do you mean it's going to change? Is there a way to leave it as is?

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,319 @@
...
+    FROM current_token_ownerships_v2 ctov
+    JOIN current_token_datas_v2 ctdv
+    ON ctov.token_data_id = ctdv.token_data_id 
+    JOIN current_collections_v2 ccv
+    ON ctdv.collection_id = ccv.collection_id 
+    WHERE ctov.token_standard = 'v1' 
+;
+-- current_collections_v2 already has index on collection_id
+CREATE INDEX lm1_curr_to_oa_tt_am_ltv_index ON current_token_ownerships_v2 USING btree (owner_address, table_type_v1, amount, last_transaction_version DESC)
+CREATE INDEX lm1_curr_to_oa_tt_ltv_index ON current_token_ownerships_v2 USING btree (owner_address, table_type_v1, last_transaction_version DESC)
```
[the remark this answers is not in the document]
rtso: This index actually isn't used frequently. I only copy pasted the existing v1 indices to the underlying v2 tables.

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
rtso: This index is also used very infrequently. Usually queries have owner

rtso:
Need to add coin_infos

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
rtso: I ran this in dbeaver and it worked so it seems like it's supported. Also I forgot to add a migration view for coin_infos, which is the only table this could join with. Yes we should add indexes on the underlying tables.

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,319 @@
...
+        -- this is a bit concerning
+        '' AS table_handle,
+        creator_address,
+        collection_name,
+        tdv.token_name AS name,
+        owner_address,
+        amount,
+        table_type_v1 AS table_type,
+        tov.inserted_at,
+        tdv.collection_id AS collection_data_id_hash,
```
ying-w, line 126: values are different

rtso: can you give an example?

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
ying-w: i think we stopped using _hash fields in v2 since we have proper _id fields, for the purpose of backwards compatibility, not sure if we should put the _id fields in (since purpose of column should be unique identifier) which would change the value or hash to keep value the same

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
ying-w: similar to coins, logic would be `encode(
sha256(
(collection_creator_address || '::' || token_name)::bytea
), 'hex')`

from here https://github.com/aptos-labs/aptos-indexer-processors/blob/5244b84fa5ed872e5280dc8df032d744d62ad29d/python/utils/token_utils.py#L44

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
ying-w: > these are supposed to be used by marketplaces to indicate which fields are non modifiable. however basically everyone just used one minting contract that sets all of these to true so majority of these ended up being true. and nobody seems to care about these.
my thought is that we can remove them and nothing would happen since I haven’t seen any UI that exposes these. If someone ends up wanting them we could create a new table altogether (similar to royalty and other metadata that aren’t critical).

Field is not nullable, previously these could have different bool values

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
rtso: The processing code does the hash already - https://github.com/aptos-labs/aptos-indexer-processors/blob/48d779449caa5b10a51a6319c61992af0edb52ce/rust/processor/src/models/token_v2_models/v2_collections.rs#L264

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,319 @@
...
+        NULL AS coin_type,
+        NULL AS coin_amount,
+        tav.inserted_at,
+        tav.transaction_timestamp,
+        event_index
+    FROM public.token_activities_v2 tav 
+    JOIN token_datas_v2 tdv 
+    ON tav.token_data_id = tdv.token_data_id AND tav.transaction_version = tdv.transaction_version
+    JOIN collections_v2 cv
+    ON tdv.collection_id = cv.collection_id AND tdv.transaction_version = cv.transaction_version
```
ying-w, line 103: i dont think this works, token_activities_v2 has new entries when tokens have activities whereas collections_v2 has new entries when collection properties change

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
ying-w: just realized that, so just need `ltrim`

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
ying-w: nvm i see, the _id fields are just 0x + _hash fields

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
rtso: i guess the best we can do is joining on the current_ tables

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,319 @@
...
+        collection_name,
+        token_name AS "name",
+        maximum,
+        supply,
+        largest_property_version_v1 AS largest_property_version,
+        token_uri AS metadata_uri,
+        -- Join on new royalty table
+        payee_address,
+        royalty_points_numerator,
+        royalty_points_denominator,
```
rtso, line 199: Need to join on new royalty tables

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -0,0 +1,319 @@
...
+        event_index,
+        gas_fee_payer_address,
+        storage_refund_amount,
+    FROM public.fungible_asset_activities
+    WHERE token_standard = 'v1'
+  ;
+CREATE INDEX lm1_ca_ct_a_index ON public.fungible_asset_activities USING btree (asset_type, amount)
+CREATE INDEX lm1_ca_ct_at_a_index ON public.fungible_asset_activities USING btree (asset_type, "type", amount)
+CREATE INDEX lm1_ca_oa_ct_at_index ON public.fungible_asset_activities USING btree (owner_address, asset_type, "type", amount)
+CREATE INDEX lm1_ca_oa_igf_index ON public.fungible_asset_activities USING btree (owner_address, is_gas_fee)
```
bowenyang007, line 41: have we tested any of this in terms of performance?

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
bowenyang007: So 2 options here: 
1. make these nullable and set to null. this changes the schema and could cause some compatibility issues
2. keep as is. 

@ying-w do you prefer one or the other?

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
yuunlimm, line 41: @bowenyang007 this is where it's failing

rust/processor/migrations/2024-05-22-200847_add_v1_migration_views/up.sql
```diff
@@ -52,11 +50,26 @@ SELECT transaction_version,
     inserted_at
 FROM public.fungible_asset_balances
 WHERE token_standard = 'v1';
+-- replace `coin_infos` with `fungible_asset_metadata`
+CREATE OR REPLACE VIEW legacy_migration_v1.coin_infos AS
+SELECT encode(sha256(asset_type::bytea), 'hex'),
```
yuunlimm, line 55: do we need to rename this?  as `coin_type_hash`

bowenyang007: Yes! Good catch.

PR #386, opened 2024-05-23
Use GraphQL rather than REST for indexer-metrics

banool:
## Summary
API Gateway doesn't proxy the REST endpoints for Hasura. We spoke a while ago and decided we'd just not use them. This PR makes that the case.

## Test Plan
```
$ cat /tmp/config.yaml
health_check_port: 8080
server_config:
  chain_name: testnet
  hasura_graphql_endpoint: https://api.testnet.aptoslabs.com/v1/graphql
  fullnode_rest_api_endpoint: https://api.testnet.aptoslabs.com/v1
```
```
cargo run -p indexer-metrics -- -c config.yaml
```
Output: https://gist.github.com/banool/c9ead808cba82ece5689a26fa6abd7cf.

See the internal-ops half of this: https://github.com/aptos-labs/internal-ops/pull/3265.


banool:
Done, see https://github.com/aptos-labs/internal-ops/pull/3265.

just-in-chang (approved):

PR #360, opened 2024-05-01
[event_v2] indexer logic update to handle migration 

lightmark:
# PR description

In this PR we add event v2 migration indexing support.

PRs to handle:

1. https://github.com/aptos-labs/aptos-core/pull/10532
2. https://github.com/aptos-labs/aptos-core/pull/11688
3. https://github.com/aptos-labs/aptos-core/pull/13147

## Events to migrate

* V1 + V2: means dedupe is required
* only V1 or V2: means no dedupe is required due to no double emission.

| Module            | Event Name                        | Migration Note          |
| ----------------- | --------------------------------- | ----------------------- |
| account           | KeyRotation                       | not indexed             |
| aptos_account     | DirectCoinTransferConfigUpdated   | not indexed             |
| coin              | Deposit                           | only V1 or V2; not both |
|                   | Withdraw                          | only V1 or V2; not both |
| fungible_asset    | Deposit                           | only V1 or V2; not both |
|                   | Withdraw                          | only V1 or V2; not both |
|                   | Frozen                            | only V1 or V2; not both |
| object            | Transfer                     | V1 + V2                 |
| aptos_governance  | CreateProposal                    | not indexed             |
|                   | Vote                              | V1 + V2                 |
|                   | UpdateConfig                      | not indexed             |
| block             | NewBlock                          | not indexed             |
|                   | UpdateEpochInterval               | not indexed             |
| multisig_account  | \-- all events --                 | not indexed             |
| reconfiguration   | NewEpoch                          | not indexed             |
| stake             | RegisterValidatorCandidate        | not indexed             |
|                   | SetOperator                       | not indexed             |
|                   | AddStake                          | V1 + V2                 |
|                   | ReactivateStakeEvent              | not indexed             |
|                   | RotateConsensusKey                | not indexed             |
|                   | UpdateNetworkAndFullnodeAddresses | not indexed             |
|                   | IncreaseLockup                    | not indexed             |
|                   | JoinValidatorSet                  | not indexed             |
|                   | DistributeRewards                 | V1 + V2                 |
|                   | UnlockStake                       | not indexed             |
|                   | WithdrawStake                     | not indexed             |
|                   | LeaveValidatorSet                 | not indexed             |
| staking_contract  | \-- all events --                 | not indexed             |
| vesting           | \-- all events --                 | not indexed             |
| voting            | \-- all events --                 | not indexed             |
| token             | Mutation                          | V1 + V2                 |
| token_event_store | \-- all events --                 | not indexed             |
[...]

PR #388, opened 2024-05-23
[event_v2] indexer logic update to handle migration (#360)

lightmark:
revert https://github.com/aptos-labs/aptos-indexer-processors/commit/8c41d26171446837edcbf6c7a4084e59fd43f4ba
with a fix


rust/processor/src/models/coin_models/coin_utils.rs
```diff
@@ -288,16 +302,38 @@ impl CoinResource {
...
+        txn_version: i64,
+    ) -> Result<Option<(CoinEvent, Option<String>)>> {
         match data_type {
-            "0x1::coin::WithdrawEvent" => {
-                serde_json::from_str(data).map(|inner| Some(CoinEvent::WithdrawCoinEvent(inner)))
+            "0x1::coin::WithdrawEvent" => serde_json::from_str(data)
+                .map(|inner| Some((CoinEvent::WithdrawCoinEvent(inner), None))),
+            "0x1::coin::DepositEvent" => serde_json::from_str(data)
+                .map(|inner| Some((CoinEvent::WithdrawCoinEvent(inner), None))),
+            t if t.starts_with("0x1::coin::Withdraw") => {
```
rtso, line 320: why not explicitly match the event name?

rust/processor/src/models/mod.rs
```diff
@@ -16,3 +20,111 @@ pub mod token_models;
...
+            "0x3::token_transfers::TokenClaimEvent",
+            "0x3::token_transfers::TokenClaim",
+        );
+        m.insert("0x4::token::MutationEvent", "0x4::token::Mutation");
+        m.insert("0x1::object::TransferEvent", "0x1::object::TransferEvent");
+        m
+    };
+    pub static ref EVENTS_MIGRATION_PREFIX_MAP: HashMap<&'static str, &'static str> = {
+        let mut m = HashMap::new();
+        m.insert("0x1::coin::WithdrawEvent", "0x1::coin::Withdraw<");
```
rtso, line 77: Can you add a comment with example of what this is for?

rust/processor/src/models/mod.rs
```diff
@@ -16,3 +20,111 @@ pub mod token_models;
...
+        m.insert("0x1::coin::DepositEvent", "0x1::coin::Deposit<");
+        m
+    };
+}
+
+pub(crate) fn should_skip(index: usize, event: &Event, events: &[Event]) -> bool {
+    index > 0
+        && (EVENTS_MIGRATION_MAP.get(event.type_str.as_str())
+            == Some(&&*events[index - 1].type_str)
```
rtso, line 86: Can you add a comment with the assumption this is making? I'm guessing the assumption is that an event v2 will always come right after an event v1?

rust/processor/src/models/mod.rs
```diff
@@ -16,3 +20,111 @@ pub mod token_models;
...
+    };
+    pub static ref EVENTS_MIGRATION_PREFIX_MAP: HashMap<&'static str, &'static str> = {
+        let mut m = HashMap::new();
+        m.insert("0x1::coin::WithdrawEvent", "0x1::coin::Withdraw<");
+        m.insert("0x1::coin::DepositEvent", "0x1::coin::Deposit<");
+        m
+    };
+}
+
+pub(crate) fn should_skip(index: usize, event: &Event, events: &[Event]) -> bool {
```
rtso, line 83: The logic may be simpler if we do a first pass of the events to check if it's events v1 or v2. Then should_skip(is_events_v2, event_name) just checks if the event_name is in the v1 set or v2 set.

rust/processor/src/processors/fungible_asset_processor.rs
```diff
@@ -433,6 +434,9 @@ async fn parse_v2_coin(
 
         // Loop to handle events and collect additional metadata from events for v2
         for (index, event) in events.iter().enumerate() {
+            if should_skip(index, event, events) {
+                continue;
+            };
```
rtso, line 439: confirmed with @lightmark we don't need this for fungible asset

rust/processor/src/models/mod.rs
rtso, line 77: What's the rationale for having coin type generic in the event name? This makes it so much harder to index bc we need to parse out the generic part of the event name. Why not have coin_type as part of the event struct?

bowenyang007 (changes requested):
pushing back to your queue. Is it still relevant? If so let's address the comments.

rust/processor/src/processors/token_v2_processor.rs
```diff
@@ -770,41 +771,66 @@ async fn parse_v2_token(
...
+                            // note that these will be multiplied by -1 so that it doesn't conflict with wsc index
+                            let index = if index == 0 {
+                                user_txn.events.len()
+                            } else {
+                                index
+                            };
+                            aggregated_data
+                                .transfer_events
+                                .push((index as i64, transfer_events));
+                        }
```
rtso, line 831: Why is this changed so much? I thought `should_skip` handles skipping the v1 events

rtso (changes requested):

rust/processor/src/db/common/models/mod.rs
```diff
@@ -16,3 +23,63 @@ pub mod token_models;
 pub mod token_v2_models;
 pub mod transaction_metadata_model;
 pub mod user_transactions_models;
+
+static EVENT_V2_ENABLED: AtomicBool = AtomicBool::new(false);
```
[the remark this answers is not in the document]
lightmark: how do you know whether to skip it when you see a v1?

rust/processor/src/processors/fungible_asset_processor.rs
```diff
@@ -596,6 +597,9 @@ async fn parse_v2_coin(
 
         // Loop to handle events and collect additional metadata from events for v2
         for (index, event) in events.iter().enumerate() {
+            if should_skip(event) {
```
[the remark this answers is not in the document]
lightmark: yes. so it does not hurt. This line is added by Larry :(

rust/processor/src/processors/token_v2_processor.rs
lightmark: ...Because... Igor makes them mixed now. go to the smart contracts and then you will know why. That's why I want to separate a fix for kent first.

rust/processor/src/db/common/models/mod.rs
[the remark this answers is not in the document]
rtso:
```
let mapping = {
 withdrawV1 -> withdrawV2
}
should_skip(event_name, prev_event_name):
if event_name is in mapping and mapping[event_name] == prev_event_name: true
else: false

```

rust/processor/src/processors/token_v2_processor.rs
rtso: Chatted offline about this. There could be v1 and v2 event next to each other for different collections and could accidentally skip a v1 event 

To get around this, we'll create `should_skip` that only skips if event names AND collection address matches.

python/processors/nft_orderbooks/parsers/okx_parser.py
bowenyang007, line 1: can we remove these? we're no longer maintaining the python stuff anyway so it's not needed and we don't want to test it anyway.

lightmark: larry added these. 
what to do with this file?

bowenyang007: uh just revert to how it was.

bowenyang007:
Need a better summary and test plan. The summary cannot be "revert of a revert".

rust/processor/src/db/common/models/coin_models/coin_activities.rs
```diff
@@ -186,6 +187,16 @@ impl CoinActivity {
...
+            if should_skip(
+                event,
+                if index == 0 {
+                    None
+                } else {
+                    Some(&events[index - 1])
+                },
+            ) {
+                continue;
+            }
```
bowenyang007, line 199: What's happening here?

rust/processor/src/db/common/models/coin_models/coin_activities.rs
```diff
@@ -222,23 +233,44 @@ impl CoinActivity {
...
+                inner.amount.clone(),
+                Some(inner.coin_type.clone()),
+            ),
         };
-        let coin_type =
+        let coin_type = if let Some(coin_type) = coin_type_option {
+            coin_type
+        } else {
+            let event_move_guid = EventGuidResource {
+                addr: standardize_address(event.key.as_ref().unwrap().account_address.as_str()),
```
bowenyang007, line 262: this is just owner_account from above right?

rust/processor/src/db/common/models/coin_models/coin_activities.rs
```diff
@@ -222,23 +233,44 @@ impl CoinActivity {
...
+                Some(inner.coin_type.clone()),
+            ),
+            CoinEvent::DepositCoinEventV2(inner) => (
+                standardize_address(&inner.account),
+                inner.amount.clone(),
+                Some(inner.coin_type.clone()),
+            ),
         };
-        let coin_type =
+        let coin_type = if let Some(coin_type) = coin_type_option {
```
bowenyang007, line 258: So for v2 what happens?

lightmark:
we will not do double emission.
refactored.

rtso (approved):

PR #389, opened 2024-05-25
[WIP] More Advanced Transaction Filtering

CapCap:
an MVP version of the transaction filtering crate

### Performance
On a random 3mb proto of 1000txns during taptos, running the "everything" filter (looped) gives:
```
BENCH: Took 13.225167ms for 1000000 transactions (13ns each)
```

Random 10mb proto from April 25th with "everything" filter:
```
BENCH: Took 21.815375ms for 1000000 transactions (21ns each)
```

Graffio 100mb proto ("everything" filter, but with `or`): 
```
BENCH: Took 31.559333ms for 1000000 transactions (31ns each)
```


rust/transaction-filter/src/filter_operator.rs
```diff
@@ -0,0 +1,336 @@
...
+                    if let TxnData::User(u) = txn_data {
+                        u.request
+                            .as_ref()
+                            .map(|req| ut_filter.is_allowed(req))
+                            .unwrap_or(false)
+                    } else {
+                        false
+                    }
+                })
+                .unwrap_or(false),
```
banool, line 50: Nit, if possible it'd be cleaner if this logic was just inside UserTransactionRequestFilter.is_allowed.

rust/transaction-filter/src/filter_operator.rs
```diff
@@ -0,0 +1,336 @@
...
+#[serde(deny_unknown_fields)]
+#[serde(tag = "type")]
+pub enum APIFilter {
+    TransactionRootFilter(TransactionRootFilter),
+    UserTransactionRequestFilter(UserTransactionRequestFilter),
+    EventFilter(EventFilter),
+    WriteSetChangeFilter(WriteSetChangeFilter),
+}
+
+impl Filterable<Transaction> for APIFilter {
```
banool, line 24: You could use enum_dispatch / ambassador to help eliminate some of the boilerplate here.

rust/transaction-filter/src/traits.rs
```diff
@@ -0,0 +1,94 @@
+use anyhow::Error;
+
+/// Simple trait to allow for filtering of items of type T
+pub trait Filterable<T> {
+    /// Whether this filter is correctly configured/initialized
+    /// Any call to `is_valid` is responsible for recursively checking the validity of any nested filters
+    fn is_valid(&self) -> Result<(), Error>;
+
+    fn is_allowed(&self, item: &T) -> bool;
```
banool, line 9: A doc comment here would be nice.

rust/transaction-filter/src/traits.rs
```diff
@@ -0,0 +1,94 @@
+use anyhow::Error;
+
+/// Simple trait to allow for filtering of items of type T
+pub trait Filterable<T> {
+    /// Whether this filter is correctly configured/initialized
+    /// Any call to `is_valid` is responsible for recursively checking the validity of any nested filters
+    fn is_valid(&self) -> Result<(), Error>;
```
banool, line 7: I imagine the reason we have `is_valid` is because we can't do it in constructors because these structs might get built directly from configuration, and we can't do it in the parsing logic because serde doesn't provide powerful enough attributes to let us do validity checking at parsing time, yeah?

rust/transaction-filter/src/filter_operator.rs
```diff
@@ -0,0 +1,336 @@
...
+
+impl Filterable<Transaction> for LogicalAnd {
+    fn is_valid(&self) -> Result<(), Error> {
+        for filter in &self.and {
+            filter.is_valid()?;
+        }
+        Ok(())
+    }
+
+    fn is_allowed(&self, item: &Transaction) -> bool {
```
banool, line 125: Good old Rust, feels like these tiny is_allowed functions are the real meat and potatoes and everything else is boilerplate hahah.

rust/transaction-filter/src/filters/event_filter.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct EventFilter {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub data: Option<JsonSearchTerm>,
+    // Only for events that have a struct as their generic
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub struct_type: Option<MoveStructTagFilter>,
+    // TODO: address? etc
```
banool, line 14: Like the address that emitted the event for v1 events?

rust/transaction-filter/src/filters/event_filter.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+impl Filterable<Event> for EventFilter {
+    fn is_valid(&self) -> Result<(), Error> {
+        if self.data.is_none() && self.struct_type.is_none() {
+            return Err(Error::msg(
```
banool, line 20: Not 100% sure what the errors would look like but I'd suggest making sure it's obvious to the user which part of their query is invalid. If all they see is `At least one of data or struct_type must be set` it doesn't make it clear which part is the problem.

rust/transaction-filter/src/filters/move_module.rs
```diff
@@ -0,0 +1,76 @@
...
+        self.address.is_allowed(&module_id.address) && self.name.is_allowed(&module_id.name)
+    }
+}
+#[derive(Debug, Deserialize, PartialEq, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct EntryFunctionFilter {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub module: Option<MoveModuleFilter>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub method: Option<String>,
```
banool, line 33: Let's go with `function`, it's what they call em in Move.

rust/transaction-filter/src/filter_operator.rs
```diff
@@ -0,0 +1,336 @@
...
+    fn is_valid(&self) -> Result<(), Error> {
+        match self {
+            APIFilter::TransactionRootFilter(filter) => filter.is_valid(),
+            APIFilter::UserTransactionRequestFilter(filter) => filter.is_valid(),
+            APIFilter::EventFilter(filter) => filter.is_valid(),
+            APIFilter::WriteSetChangeFilter(filter) => filter.is_valid(),
+        }
+    }
+
+    fn is_allowed(&self, txn: &Transaction) -> bool {
```
banool, line 34: I don't wanna nerd snipe you but it'd be neat if this function at the top level told you which filter allowed it if this is true 😛

rust/transaction-filter/src/traits.rs
CapCap: in the constructors of what? 

My assumption for these filters is they will be around for a long time, and different refs will flow through them- there's only ever one copy of these

Injecting it into the serialization would be on the GRPC side, which would be inside of poem, and I wouldn't want to mess with that 😬

rust/transaction-filter/src/filters/event_filter.rs
CapCap: yeah- I just removed this comment, don't think it's worth it

rust/transaction-filter/src/filters/event_filter.rs
CapCap: it would be some kind of catch + rewrap... doable, can do it in a follow up PR

rust/transaction-filter/src/filter_operator.rs
CapCap: what do you mean?

rust/transaction-filter/src/filters/move_module.rs
CapCap: un-nested into:
```
#[derive(Debug, Deserialize, PartialEq, Serialize)]
#[serde(deny_unknown_fields)]
pub struct EntryFunctionFilter {
    #[serde(skip_serializing_if = "Option::is_none")]
    pub address: Option<String>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub name: Option<String>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub method: Option<String>,
}
```

will rename

rust/transaction-filter/src/filter_operator.rs
rtso: Like if a transaction got filtered out, it'd be cool to see which filter exactly it failed on.

CapCap: sigh :-p

this is not technically hard, just a pain in the ass hahaha... Follow up PR?

rust/transaction-filter/src/filter_operator.rs
CapCap: enum dispatchhhhh

PR #390, opened 2024-05-28
Add parquet processor

yuunlimm:
This is to migrate some of tables in AlloyDB to BigQuery to fully deprecate those tables that are only used for analytics.

### Components added:
1. generic_parquet_processor:
   -  it store structs passed in from parquet_handler in in-memory buffer and whenever it reaches the max size, it writes to a parquet file and calls gcs_handler to upload the file to gcs bucket. 
2. parquet_default_processor
   - basically replica of deafult_processor that is responsible for
     - spawning a task that receives parquet structs from the kanal channel and invoke parquet_manager.handle
     - converting TransactionPB into structs.
     - sending parquet structs to its channel, current structs being handled: 
       - move_resources, table_items, transactions, write_set_changes.
3. parquet_handler:
   - util where it has a function that handles spawning a task for parquet_receiver to poll structs from the channel and calling parquet_manager.handle function.
4. gcs_handler:
 - responsible for reading local parquet files and uploading to GCS bucket. it retries 3 times with expontential backoffs.

Test Plan
 tested in testing environment and running locally.


rust/processor/src/config.rs
```diff
@@ -79,6 +86,21 @@ impl IndexerGrpcProcessorConfig {
...
     async fn run(&self) -> Result<()> {
+
+        // Establish GCS client
+        let gcs_config = GCSClientConfig::default()
+        .with_auth()
+        .await
+        .unwrap_or_else(|e| {
+            error!(
+                error = ?e,
+                "Failed to create gRPC client config"
```
CapCap, line 97: GCS client?

rust/processor/src/parquet_gap_detector.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    pub fn process_versions(
+        &mut self,
+        result: ParquetProcessingResult,
+    ) -> anyhow::Result<ParquetFileGapDetectorResult> {
+        // Update counts of structures for each transaction version
+        for (version, count) in result.txn_version_to_struct_count.iter() {
+            if !self.version_counters.contains_key(version) {
+                self.version_counters.insert(*version, (*count as i64).try_into().unwrap());
```
CapCap, line 46: you dont need the try into or unwrap

rust/processor/src/parquet_manager.rs
```diff
@@ -0,0 +1,554 @@
...
+        }
+
+        // let file_cloned = file.try_clone()?;
+        let mut writer: SerializedFileWriter<File> = self.writer.take().unwrap(); // Take out the writer temporarily
+        info!("Current buffer size: {:?}", self.buffer_size_bytes);
+
+        // loop the vector of struct and write to the parquet file 
+        for parquet_struct in parquet_structs.clone() {
+            self.buffer.push(parquet_struct.clone());
+            self.update_buffer_size(&parquet_struct);
```
CapCap, line 244: this seems to be duplicated- maybe some kind of generic method instead?

Also when you open the file, do you want to truncate if it already exists?

rust/processor/src/parquet_manager.rs
```diff
@@ -0,0 +1,554 @@
...
+        gcs_client: &GCSClient,
+        gap_detector_sender: &kanal::AsyncSender<ParquetProcessingResult>, 
+        changes: ParquetData) -> Result<ParquetProcessingResult>
+    {
+        let processing_start: std::time::Instant = std::time::Instant::now();
+
+        let props = WriterProperties::builder()
+            .set_compression(parquet::basic::Compression::LZ4) 
+            .build();
+        let props_arc = Arc::new(props.clone());
```
CapCap, line 122: you do this even if self.props_writer is some- when is it none? is it only during instantiation?

rust/processor/src/parquet_manager.rs
```diff
@@ -0,0 +1,554 @@
...
+        ParquetStruct::Transaction(_) => "transactions".to_string(),
+    }
+}
+
+
+fn get_schema(parquet_struct: ParquetStruct) -> Arc<parquet::schema::types::Type> {
+    let schema = match Some(parquet_struct) {
+        Some(ParquetStruct::MoveResource(struct_model)) => {
+            let mut parquet_rust_struct_vec = Vec::new();
+            parquet_rust_struct_vec.push(struct_model.clone());
```
CapCap, line 425:
```
            let mut parquet_rust_struct_vec = vec![struct_model];
            parquet_rust_struct_vec.as_slice().schema().unwrap().clone()
```
you don't need the clone, either

rust/processor/src/parquet_manager.rs
```diff
@@ -0,0 +1,554 @@
...
+                            error = ?_e,
+                            "[Parquet Handler] Parquet manager channel has been closed",
+                        );
+                        ParquetData {                  // maybe add a flag that can tell it's empty
+                            data: Vec::new(),
+                            last_transaction_timestamp: None,
+                            transaction_version_to_struct_count: AHashMap::new(),
+                        }
+                    },
+                };
```
CapCap, line 486: it seems like if you know the type you're spawning, instead of a `match`, this might be much more DRY with either a trait, or a generic, or both

rust/processor/src/parquet_manager.rs
```diff
@@ -0,0 +1,554 @@
...
+
+                let result = parquet_manager.write_and_upload(txn_pb_res, &gcs_client, new_gap_detector_sender.clone()).await;
```
CapCap, line 488: instead of cloning the `new_gap_detector_sender` every time, maybe the parquet manager should have it's own copy of it?

rust/processor/src/parquet_manager.rs
yuunlimm: yes, we should truncate if it already exists to avoid excessive duplicates

rust/processor/src/parquet_processors/move_resource_processor.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl ParquetHandlerTrait for MoveResourceParquetHandler {
+
+    fn update_buffer_size(&mut self, parquet_struct: &ParquetStruct) {
+        if let ParquetStruct::MoveResource(move_resource) = parquet_struct {
+            self.buffer_size_bytes += move_resource.size_of();
+        }
+    }
+    
+    async fn handle_transaction(
```
yuunlimm, line 54: this needs a further refactoring

rust/processor/src/parquet_processors/move_resource_processor.rs
```diff
@@ -0,0 +1,155 @@
...
+                    let new_file_path: PathBuf = PathBuf::from(format!("{}_{}.parquet", table_name.clone(), chrono::Utc::now().timestamp()));
+                    rename(file_path.clone(), new_file_path.clone())?;
+                    
+                    let mut row_group_writer = writer.next_row_group()?;
+                    let struct_buffer: Vec<_> = self.buffer.iter()
+                        .filter_map(|item| {
+                            if let ParquetStruct::MoveResource(move_resource) = item {
+                                Some(move_resource.clone())
+                            } else {
+                                None
```
yuunlimm, line 98: This is the only different lines of code among all three processors: move_resources, transactions, and write_set_changes. I was trying to move this code to ParquetHandlerTrait but with those fields from this struct, it seems not straightforward. @CapCap any insight on how I can reduce this boilerplate?!

rust/moving-average/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub struct MovingAverage {
 
 impl MovingAverage {
     pub fn new(window_millis: u64) -> Self {
-        let now = chrono::Utc::now().naive_utc().timestamp_millis() as u64;
+        let now = chrono::Utc::now().naive_utc().and_utc().timestamp_millis() as u64;
```
ying-w, line 18: this looked weird to me, should it be `chrono::Utc::now().unwrap().timestamp_millis()`? (i don't write rust code)

rust/processor/src/models/default_models/parquet_move_resources.rs
```diff
@@ -0,0 +1,162 @@
...
+use parquet_derive::{ParquetRecordWriter};
+
+
+#[derive(
+    Clone, Debug, Deserialize, FieldCount, Serialize, ParquetRecordWriter
+)]
+pub struct MoveResource {
+    pub transaction_version: i64,
+    pub write_set_change_index: i64,
+    pub transaction_block_height: i64,
```
ying-w, line 23: nit: block_height

rust/processor/src/models/default_models/parquet_transactions.rs
```diff
@@ -0,0 +1,390 @@
...
+
+use field_count::FieldCount;
+use serde::{Deserialize, Serialize};
+use parquet_derive::{ParquetRecordWriter};
+
+#[derive(Clone, Debug, Deserialize, FieldCount, Identifiable, Serialize, ParquetRecordWriter)]
+#[diesel(primary_key(version))]
+#[diesel(table_name = transactions)]
+pub struct Transaction {
+    pub version: i64,
```
ying-w, line 35: pref:

- version -> txn_version
- txn_hash 
- block_height
- block_timestamp 
- epoch
- txn_type
- payload
- payload_type

rest is fine, are you sure `state_checkpoint_hash` is optional?

rust/processor/src/models/default_models/parquet_write_set_changes.rs
```diff
@@ -0,0 +1,239 @@
...
+use field_count::FieldCount;
+use serde::{Deserialize, Serialize};
+use parquet_derive::{ParquetRecordWriter};
+
+
+#[derive(
+     Clone, Debug, Deserialize, FieldCount, Serialize, ParquetRecordWriter
+)]
+pub struct WriteSetChange {
+    pub transaction_version: i64,
```
ying-w, line 23: - transaction_version -> txn_version
- block_height
- block_timestamp
- index -> write_set_change_index
- txn_hash -> state_key_hash
-  txn_type -> change_type
-  address -> resource_address

ying-w (commented):
would be really nice to have but prob out of scope:

- having txn_version of previous state in move_resources and write_set_changes

rust/processor/src/models/default_models/parquet_transactions.rs
yuunlimm: Screenshot 2024-05-31 at 2 18 07 PM
It seems like it. 
Screenshot 2024-05-31 at 2 20 02 PM
state_change_hash seems required though

yuunlimm:
Gap Detector isn't working, but since I checked that all txns exist in BigQuery. this is a less concern for now. so I am prioritizing shipping this processor so that we can start backfilling mainnet
cc. @CapCap

rust/processor/src/parquet_processors/transaction_processor.rs
```diff
@@ -0,0 +1,157 @@
...
+                    let mut row_group_writer = writer.next_row_group()?;
+                    let struct_buffer: Vec<_> = self.buffer.iter()
+                        .filter_map(|item| {
+                            if let ParquetStruct::Transaction(transaction) = item {
+                                Some(transaction.clone())
+                            } else {
+                                None
+                            }
+                        })
+                        .collect();
```
yuunlimm, line 103: as long as this could be more generic, then we just need one processor for all tables.

rust/Cargo.toml
```diff
@@ -109,3 +109,15 @@ warp = { version = "0.3.5", features = ["tls"] }
...
+arrow = "52.0.0"
+parquet = "52.0.0"
+num = "0.4.0"
+google-cloud-storage = "0.13.0"
+hyper = { version = "0.14.18", features = ["full"] }
+parquet_derive = { version = "52.0.0" }
+canonical_json = "0.5.0"
+get-size = "0.1.3"
+get-size-derive = "0.1.3"
+uuid = { version = "1.8.0", features = ["v4"] }
```
CapCap, line 123: add newline at end pls 
and please rn cargo-machete to verify you ned all :-)

rust/moving-average/src/lib.rs
CapCap: theres a deprecation warning :-(

rust/processor/src/models/default_models/mod.rs
```diff
@@ -5,5 +5,8 @@ pub mod block_metadata_transactions;
 pub mod move_modules;
 pub mod move_resources;
 pub mod move_tables;
+pub mod parquet_move_resources;
+pub mod parquet_transactions;
+pub mod parquet_write_set_changes;
```
CapCap, line 10: can we move these to their own section below the others?

rust/processor/src/models/default_models/parquet_move_resources.rs
```diff
@@ -0,0 +1,190 @@
...
+            resource_type: "".to_string(),
+            module: "".to_string(),
+            fun: "".to_string(),
+            is_deleted: false,
+            generic_type_params: None,
+            data: None,
+            state_key_hash: "".to_string(),
+        }
+    }
+}
```
CapCap, line 65: what prevents you from deriving default?

rust/processor/src/models/default_models/parquet_move_resources.rs
```diff
@@ -0,0 +1,190 @@
...
+use aptos_protos::transaction::v1::{
+    DeleteResource, MoveStructTag as MoveStructTagPB, WriteResource,
+};
+use field_count::FieldCount;
+use get_size::GetSize;
+use get_size_derive::GetSize;
+use parquet_derive::ParquetRecordWriter;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, FieldCount, GetSize, Serialize, ParquetRecordWriter)]
```
CapCap, line 20: does `GetSize` follow pointers, arc, etc? cc @just-in-chang

just-in-chang: I was using [Allocative](https://crates.io/crates/allocative) for IMC, and that seemed to work fine, but I haven't tested it extensively. Not too sure about `GetSize` tho

CapCap: ideally we can unify on one as we're doing the same thing

CapCap:
@yuunlimm please rebase this on main, will make the review easier for us :-)

rust/processor/src/models/default_models/parquet_move_resources.rs
just-in-chang: True, allocative says they're "light weight". Not too sure if it's worth benchmarking to see which is faster tho. Also allocative is from Facebook 😃

CapCap: I think if ya'll just see which one is easier, or if they're the same go with the most updated one

rust/processor/src/db/common/models/default_models/parquet_move_resources.rs
```diff
@@ -0,0 +1,153 @@
...
+                    Ok(Some(
+                        serde_json::to_string(move_type).context("Failed to parse move type")?,
+                    ))
+                })
+                .collect::<Result<Option<String>>>()
+                .unwrap_or(None),
+        }
+    }
+
+    pub fn get_outer_type_from_resource(write_resource: &WriteResource) -> String {
```
banool, line 136: This method and `convert_move_struct_tag` don't have anything to do with `MoveResource`. Can you move them to the relevant struct or just make them freestanding?

As a general rule, if a method doesn't take `self` as a param or return `Self` there is no reason for it to a method on a struct.

rust/processor/src/db/common/models/default_models/parquet_transactions.rs
```diff
@@ -0,0 +1,381 @@
...
+    const TABLE_NAME: &'static str = "transactions";
+}
+
+impl HasVersion for Transaction {
+    fn version(&self) -> i64 {
+        self.txn_version
+    }
+}
+
+impl Default for Transaction {
```
banool, line 61: How come we have to implement Default? There is no "default transaction" so I'm not sure it makes sense to implement Default for Transaction. Same comment for many of these. If some trait / library is forcing us to implement Default, we should at least use `unreachable` so we never use it, or if something _does_ use it, for the fields we should use default, like `txn_version: Default::default()`, `event_root_hash: Default::default()`, etc.

rust/processor/src/db/common/models/default_models/parquet_transactions.rs
```diff
@@ -0,0 +1,381 @@
...
+            payload_type: None,
+            #[allow(deprecated)]
+            block_timestamp: chrono::NaiveDateTime::from_timestamp(0, 0),
+        }
+    }
+}
+
+impl Transaction {
+    fn from_transaction_info(
+        info: &TransactionInfo,
```
banool, line 88: Not 100% sure the use case but if possible, it'd be good to just take in the owned object, it'd be more efficient. No worries if it doesn't make sense for this use case.

banool (dismissed):
This is a refactor PR but it's my first time seeing lots of this code, so I left comments all over the place anyway. For any comment that's not relevant to the refactoring directly, please open an issue / task somewhere to track it for later!

I only reviewed the first bit for now. Since this is the refactoring PR I see there is now a mountain of new code hahah. I'm happy to do a full audit of the code later, but it's probably helpful for me to leave reviews in chunks rather than like 50 comments all at once.

rust/processor/src/db/common/models/default_models/parquet_move_resources.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub block_height: i64,
+    #[allocative(skip)]
+    pub block_timestamp: chrono::NaiveDateTime,
+    pub resource_address: String,
+    pub resource_type: String,
+    pub module: String,
+    pub fun: String,
+    pub is_deleted: bool,
+    pub generic_type_params: Option<String>,
+    pub data: Option<String>,
```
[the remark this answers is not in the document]
yuunlimm: During the design phase, DE team preferred storing fields as String in BigQuery.

rust/processor/src/db/common/models/default_models/parquet_move_resources.rs
```diff
@@ -0,0 +1,153 @@
...
+        txn_version: i64,
+        block_height: i64,
+        block_timestamp: chrono::NaiveDateTime,
+    ) -> Self {
+        let parsed_data = Self::convert_move_struct_tag(
+            delete_resource
+                .r#type
+                .as_ref()
+                .expect("MoveStructTag Not Exists."),
+        );
```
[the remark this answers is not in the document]
yuunlimm: synced offline, will follow up in the next pr.

rust/processor/src/db/common/models/default_models/parquet_move_resources.rs
yuunlimm: synced offline, will follow up in the next pr.

rust/processor/src/db/common/models/default_models/parquet_transactions.rs
yuunlimm: synced offline, will follow up in the next pr.

rust/processor/src/db/common/models/default_models/parquet_transactions.rs
yuunlimm: there is a usecase, where we have derive the parquet schema from the instance, and the we need to know this schema in advance. please take a look at hasParquetSchema trait in the generic_parquet_processor.rs. and we use this schema() to construct parquetHandler for each struct type. that's the only usecase of default so I am not too sure if there will any other workaround.

banool (approved):
Major items have been addressed. I'll leave a full audit here: https://github.com/aptos-labs/aptos-indexer-processors/pull/420.

rust/processor/src/db/common/models/default_models/parquet_transactions.rs
rtso: @banool for the non-Parquet processor, we use Default for transaction types where we can skip indexing most of its contents

PR #391, opened 2024-05-29
Add supply_v2, maximum_v2 to metadata

rtso:


hasura-api/metadata-json/unified.json
```diff
@@ -1,5 +1,5 @@
 {
-  "resource_version": 399,
+  "resource_version": 319,
```
jillxuu, line 2: why the version actually goes backward?

rtso: i don't know, this is just the file that hasura gave when i exported.

yuunlimm (approved):

PR #392, opened 2024-05-31
Simpler Transaction Filtering

CapCap:


rust/transaction-filter/src/traits.rs
```diff
@@ -0,0 +1,166 @@
...
+    #[inline]
+    fn is_allowed(&self, item: &String) -> bool {
+        match self {
+            Some(filter) => filter == item,
+            None => true,
+        }
+    }
+}
+
+impl Filterable<i32> for Option<i32> {
```
rtso, line 89: We use `i64` in most places in the processors, should we use `i64` here too?

rust/transaction-filter/src/filters/move_module.rs
```diff
@@ -0,0 +1,58 @@
...
+            return Err(anyhow!(
+                "At least one of address, module or name must be set"
+            ));
+        };
+        Ok(())
+    }
+
+    #[inline]
+    fn is_allowed(&self, struct_tag: &MoveStructTag) -> bool {
+        self.address.is_allowed(&struct_tag.address)
```
rtso, line 54: Does it expect that the address is 64 length? Should we support both truncated and standardized addresses?

rust/transaction-filter/src/filters/write_set_change_filter.rs
```diff
@@ -0,0 +1,306 @@
...
+                    {
+                        return false;
+                    }
+                }
+                if let Some(key) = &self.key {
+                    if !dti.data.as_ref().map_or(false, |dtd| key == &dtd.key) {
+                        return false;
+                    }
+                }
+            },
```
rtso, line 285: Some filters `&&` all the fields and some filters `||`. We should keep it consistent.

rust/transaction-filter/src/filters/write_set_change_filter.rs
```diff
@@ -0,0 +1,306 @@
...
+use aptos_protos::transaction::v1::{
+    write_set_change::Change, DeleteModule, DeleteResource, DeleteTableItem, WriteModule,
+    WriteResource, WriteSetChange, WriteTableItem,
+};
+use serde::{Deserialize, Serialize};
+
+/// This is a wrapper around ChangeItemFilter, which differs because:
+/// While `ChangeItemFilter` will return false if the Event does not match the filter,
+/// `ChangeItemFilter` will return true- i.e `WriteSetChangeFilter` *only* tries to match if the
+/// change type matches its internal change type
```
rtso, line 12: Do we need both the `ChangeItemFilter` and `WriteSetChangeFilter`? If it's possible, I prefer having a single one to keep things simpler. 

Also how will we check a transaction if a transaction is allowed? Is it if _any_ of its wsc is allowed?

CapCap: currently this `simple` version- yeah

In the follow up PR, there is a `AND`, `OR`, and `NOT`

rust/transaction-filter/src/filters/move_module.rs
CapCap: it does expect standardized addresses, yeah... I was hoping this could be a convention, but we could also put it in custom serde?

PR #394, opened 2024-06-04
skip writing deprecated tables in default processor

yuunlimm:
### Description
As part of the v1 table deprecation and parquet migration, we will stop writing to deprecated tables like move resources, write_set_changes, and transations.

### Test plan
tested locally 
Screenshot 2024-06-04 at 11 22 13 AM

### DIFF
--- a/rust/processor/src/config.rs
+++ b/rust/processor/src/config.rs
@@ -9,7 +9,7 @@ use ahash::AHashMap;
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 use server_framework::RunnableConfig;
-use std::time::Duration;
+use std::{collections::HashSet, time::Duration};
 use url::Url;
 
 pub const QUERY_DEFAULT_RETRIES: u32 = 5;
@@ -49,8 +49,8 @@ pub struct IndexerGrpcProcessorConfig {
 
     #[serde(default)]
     pub transaction_filter: TransactionFilter,
-    // flag to skip db writes for the deprecated tables
-    pub skip_deprecated_tables: Option<bool>,
+    // String vector for deprecated tables to skip db writes
+    pub deprecated_tables: HashSet<String>,
 }
 
 impl IndexerGrpcProcessorConfig {
@@ -97,7 +97,7 @@ impl RunnableConfig for IndexerGrpcProcessorConfig {
             self.enable_verbose_logging,
             self.transaction_filter.clone(),
             self.grpc_response_item_timeout_in_secs,
-            self.skip_deprecated_tables,
+            self.deprecated_tables.clone(),
         )
         .await
         .context("Failed to build worker")?;
--- a/rust/processor/src/config.rs
+++ b/rust/processor/src/config.rs
@@ -49,6 +49,8 @@ pub struct IndexerGrpcProcessorConfig {
 
     #[serde(default)]
     pub transaction_filter: TransactionFilter,
+    // flag to skip db writes for the deprecated tables
+    pub skip_deprecated_tables: Option<bool>,
 }
 
 impl IndexerGrpcProcessorConfig {
@@ -95,6 +97,7 @@ impl RunnableConfig for IndexerGrpcProcessorConfig {
             self.enable_verbose_logging,
             self.transaction_filter.clone(),
             self.grpc_response_item_timeout_in_secs,
+            self.skip_deprecated_tables,
         )
         .await
         .context("Failed to build worker")?;
--- a/rust/processor/src/processors/default_processor.rs
+++ b/rust/processor/src/processors/default_processor.rs
@@ -3,13 +3,18 @@
 
 use super::{ProcessingResult, ProcessorName, ProcessorTrait};
 use crate::{
-    db::common::models::default_models::{
-        block_metadata_transactions::{BlockMetadataTransaction, BlockMetadataTransactionModel},
-        move_modules::MoveModule,
-        move_resources::MoveResource,
-        move_tables::{CurrentTableItem, TableItem, TableMetadata},
-        transactions::TransactionModel,
-        write_set_changes::{WriteSetChangeDetail, WriteSetChangeModel},
+    db::common::models::{
+        default_models::{
+            block_metadata_transactions::{
+                BlockMetadataTransaction, BlockMetadataTransactionModel,
+            },
+            move_modules::MoveModule,
+            move_resources::MoveResource,
+            move_tables::{CurrentTableItem, TableItem, TableMetadata},
+            transactions::TransactionModel,
+            write_set_changes::{WriteSetChangeDetail, WriteSetChangeModel},
+        },
+        TableName,
     },
     schema,
     utils::database::{execute_in_chunks, get_config_table_chunk_size, ArcDbPool},
@@ -24,8 +29,8 @@ use diesel::{
     query_builder::QueryFragment,
     ExpressionMethods,
 };
-use tokio::join;
 use std::{collections::HashSet, fmt::Debug};
+use tokio::join;
 use tracing::error;
 
 pub struct DefaultProcessor {
--- a/rust/processor/src/processors/default_processor.rs
+++ b/rust/processor/src/processors/default_processor.rs
@@ -3,18 +3,13 @@
 
 use super::{ProcessingResult, ProcessorName, ProcessorTrait};
 use crate::{
-    db::common::models::{
-        default_models::{
-            block_metadata_transactions::{
-                BlockMetadataTransaction, BlockMetadataTransactionModel,
-            },
-            move_modules::MoveModule,
-            move_resources::MoveResource,
-            move_tables::{CurrentTableItem, TableItem, TableMetadata},
-            transactions::TransactionModel,
-            write_set_changes::{WriteSetChangeDetail, WriteSetChangeModel},
-        },
-        TableName,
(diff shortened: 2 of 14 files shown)


rust/processor/src/config.rs
```diff
@@ -49,6 +49,8 @@ pub struct IndexerGrpcProcessorConfig {
 
     #[serde(default)]
     pub transaction_filter: TransactionFilter,
+    // flag to skip db writes for the deprecated tables
+    pub skip_deprecated_tables: Option<bool>,
```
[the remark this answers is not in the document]
yuunlimm: we will probably need this for other processors too for tranche 2.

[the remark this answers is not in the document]
grao1991: Given there are other people running our stack and they might choose to deprecate the tables on a different schedule and/or keep some of them. I think it's better to make one flag for each table, and put them under the config of DefaultProcessor.

rust/processor/src/processors/default_processor.rs
yuunlimm, line 1: what would be a simple way to join res concurrently with those conditional inserts? maybe mock the future?

rust/processor/src/processors/default_processor.rs
```diff
@@ -23,20 +23,34 @@ use diesel::{
...
 use std::fmt::Debug;
-use tokio::join;
 use tracing::error;
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct DefaultProcessorConfig {
+    pub skip_move_resources: Option<bool>,
+    pub skip_transactions: Option<bool>,
+    pub skip_write_set_changes: Option<bool>,
```
bowenyang007, line 35: It might be better to make a vec of strings (table names) b/c this could get tedious as we remove more tables.

rust/processor/migrations/2024-06-04-211042_replace_move_resource_view/up.sql
bowenyang007, line 1: This isn't needed right? let's just have this as a part of the other PR 
https://github.com/aptos-labs/aptos-indexer-processors/pull/385

rust/processor/src/processors/default_processor.rs
```diff
@@ -76,76 +82,84 @@ async fn insert_to_db(
...
+            ),
+        )
+        .await?;
+    };
 
-    for res in [
-        txns_res, bmt_res, wst_res, mm_res, mr_res, ti_res, cti_res, tm_res,
-    ] {
-        res?;
-    }
```
rtso, line 148: I think we still need this to make sure the execution is successful. We can create a vec of `res` and then join and ensure success at the very end.

rtso, line 148: I think we still need this to make sure the execution is successful. We can create a vec of `res` and then join and ensure success at the very end.

yuunlimm: makes sense, will add this back to join and make sure the db writes were successful!

rust/processor/src/processors/default_processor.rs
```diff
@@ -76,76 +82,84 @@ async fn insert_to_db(
...
         insert_table_metadata_query,
         table_metadata,
         get_config_table_chunk_size::<TableMetadata>("table_metadatas", per_table_chunk_sizes),
-    );
+    )
+    .await?;
 
-    let (txns_res, bmt_res, wst_res, mm_res, mr_res, ti_res, cti_res, tm_res) =
-        join!(txns_res, bmt_res, wst_res, mm_res, mr_res, ti_res, cti_res, tm_res);
+    if !deprecated_tables.contains(&"transactions".to_string()) {
```
bowenyang007, line 131: Adding the deprecation here means that we're still processing the models though.

I would think that the better way would be having a function in the structs that says fn should_process that gets called in the processor code.

yuunlimm: yeah that's what I meant in the standup we can avoid processing the models instead of just not write to db.

CapCap: I suggested to merge this one with struct parsing but not DB writing, and then follow up with one that disables them further upstream after

rust/processor/src/processors/default_processor.rs
```diff
@@ -76,76 +82,84 @@ async fn insert_to_db(
...
+        )
+        .await?;
+    };
 
-    for res in [
-        txns_res, bmt_res, wst_res, mm_res, mr_res, ti_res, cti_res, tm_res,
-    ] {
-        res?;
-    }
+    if !deprecated_tables.contains(&"move_resources".to_string()) {
```
CapCap, line 154: this is expensive- instead of strings, lets use an enum for this type and https://github.com/bitflags/bitflags, instead of array iteration/search for each txn*structs

rust/processor/src/models/default_models/transactions.rs
```diff
@@ -186,13 +194,35 @@ impl Transaction {
...
             .expect("Transaction timestamp doesn't exist!");
+
+        let (wsc, wsc_detail) = WriteSetChangeModel::from_write_set_changes(
+            &transaction_info.changes,
+            version,
+            block_height,
+            deprecated_tables,
+        );
+
+        if deprecated_tables.contains(TableFlags::TRANSACTIONS) {
```
bowenyang007, line 205: It might be cleaner to do this inside of the models themselves, e.g. BlockMetadataTransaction::from_transaction. 

Maybe add a todo and we can fix later? This is a nit.

rust/processor/src/processors/default_processor.rs
```diff
@@ -80,53 +92,62 @@ async fn insert_to_db(
...
+        ),
     );
+
     let bmt_res = execute_in_chunks(
         conn.clone(),
         insert_block_metadata_transactions_query,
         block_metadata_transactions,
         get_config_table_chunk_size::<BlockMetadataTransactionModel>(
-            "block_metadata_transactions",
+            BlockMetadataTransaction::table_name(),
```
rtso, line 106: I think @CapCap was looking into a way to automatically get the table name from diesel type, so we wouldn't have to do something like this...

rust/processor/src/worker.rs
```diff
@@ -45,6 +47,20 @@ use url::Url;
...
+    #[derive(Debug, Clone, Copy)]
+    pub struct TableFlags: u32 {
+        const TRANSACTIONS = 0b00000001;
+        const BLOCK_METADATA_TRANSACTIONS = 0b00000010;
+        const WRITE_SET_CHANGES = 0b00000100;
+        const MOVE_MODULES = 0b00001000;
+        const MOVE_RESOURCES = 0b00010000;
+        const TABLE_ITEMS = 0b00100000;
+        const CURRENT_TABLE_ITEMS = 0b01000000;
+        const TABLE_METADATA = 0b10000000;
```
bowenyang007, line 60: if we need to add more than 32 tables what do we do?

rust/processor/src/processors/default_processor.rs
bowenyang007: it's a hashset lookup. Should be O(1). The to_string is unnecessary though.

rust/processor/src/models/default_models/transactions.rs
```diff
@@ -186,13 +194,35 @@ impl Transaction {
...
             .expect("Transaction timestamp doesn't exist!");
+
+        let (wsc, wsc_detail) = WriteSetChangeModel::from_write_set_changes(
+            &transaction_info.changes,
+            version,
+            block_height,
+            deprecated_tables,
+        );
+
+        if Transaction::skip_processing(deprecated_tables) {
```
rtso, line 205: It's kinda messy that we tied the struct to the table name, because that's not always the case that 1 struct maps cleanly to 1 table. As we see here, `Transaction` actually parses a lot more other structs than just `transactions`. 

I liked @CapCap 's idea of having the filter only inside `insert_to_db`. Performance is not an issue because CPU is "cheap" compared to IO.

yuunlimm: agreed, it's really messy that we pass flags around every place. I misunderstood, I thought Max suggested to have this logic within the struct parsing but not DB writing. 

We should be able to get rid of this anyway once we have a parquet deployed or this could be replaced with a further upstream filter or transaction filter what max is working on?

rust/processor/src/worker.rs
yuunlimm: we can use 64 or 128

bowenyang007 (approved):

rust/processor/src/processors/default_processor.rs
```diff
@@ -138,11 +148,11 @@ async fn insert_to_db(
         get_config_table_chunk_size::<TableMetadata>("table_metadatas", per_table_chunk_sizes),
     );
 
-    let (txns_res, bmt_res, wst_res, mm_res, mr_res, ti_res, cti_res, tm_res) =
-        join!(txns_res, bmt_res, wst_res, mm_res, mr_res, ti_res, cti_res, tm_res);
+    let (txns_res, wst_res, mr_res, bmt_res, mm_res, ti_res, cti_res, tm_res) =
+        join!(txns_res, wst_res, mr_res, bmt_res, mm_res, ti_res, cti_res, tm_res);
```
CapCap, line 152: why the reorder?

CapCap (approved):

PR #395, opened 2024-06-04
add bigquery load script for parquet files

yuunlimm:
these are the template that we could use to load parquet files to Big Query both in testnet and mainnet. we just have to specify the right path to the table. 

parquet file with schema in gcs
transaction
Screenshot 2024-06-04 at 4 02 08 PM
move_resources
Screenshot 2024-06-04 at 4 02 01 PM
write_set_changes
Screenshot 2024-06-04 at 4 00 50 PM

### Test plan
aptos-indexer-data-raw.testnet_nrt.transactions new table added with the script above
Screenshot 2024-06-04 at 4 10 44 PM
Screenshot 2024-06-04 at 4 09 44 PM


rust/parquet-bq-scripts/write_set_changes_create.sql
```diff
@@ -0,0 +1,16 @@
...
+  txn_version INT64,
+  write_set_change_index INT64,
+  state_key_hash STRING,
+  change_type STRING,
+  resource_address STRING,
+  block_height INT64,
+  block_timestamp TIMESTAMP,
+  --
+  bq_inserted_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP(),
+  PRIMARY KEY(txn_version, index) NOT ENFORCED
```
ying-w, line 12: index -> write_set_change_index

rust/parquet-bq-scripts/write_set_changes_create.sql
```diff
@@ -0,0 +1,16 @@
...
+)
+PARTITION BY TIMESTAMP_TRUNC(block_timestamp, DAY)
+CLUSTER BY txn_version, change_type, block_height
```
ying-w, line 15: also add `state_key_hash` to cluster

ying-w (approved):
looks good

nit: add extra newline to bottom of these sql files

PR #397, opened 2024-06-10
Make a crate of grpc_stream

rtso:
## Description
Clean up `grpc_stream.rs` so we can make a crate of it, which will be used in the SDK. 

## Testing
`cargo run --release -- -c config.yaml`

### DIFF
--- a/rust/processor/src/grpc_stream.rs
+++ b/rust/processor/src/grpc_stream.rs
@@ -35,6 +35,8 @@ pub const RECONNECTION_MAX_RETRIES: u64 = 5;
 /// 256MB
 pub const MAX_RESPONSE_SIZE: usize = 1024 * 1024 * 256;
 
+const PROCESSOR_SERVICE_TYPE: &str = "processor";
+
 #[derive(Clone)]
 pub struct TransactionsPBResponse {
     pub transactions: Vec<Transaction>,
@@ -296,13 +298,7 @@ pub async fn get_chain_id(
     }
 }
 
-/// Gets a batch of transactions from the stream. Batch size is set in the grpc server.
-/// The number of batches depends on our config
-/// There could be several special scenarios:
-/// 1. If we lose the connection, we will try reconnecting X times within Y seconds before crashing.
-/// 2. If we specified an end version and we hit that, we will stop fetching, but we will make sure that
-///    all existing transactions are processed
-pub async fn create_fetcher_loop(
+pub struct TransactionStream {
     txn_sender: AsyncSender<TransactionsPBResponse>,
     indexer_grpc_data_service_address: Url,
     indexer_grpc_http2_ping_interval: Duration,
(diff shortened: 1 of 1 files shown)


rust/aptos-indexer-transaction-stream/src/utils/counters.rs
```diff
@@ -0,0 +1,104 @@
...
+                "[Parser] Processor finished processing one batch of transaction"
+            },
+            ProcessorStep::ProcessedMultipleBatches => {
+                "[Parser] Processor finished processing multiple batches of transactions"
+            },
+        }
+    }
+}
+
+/// These metrics are temporary (suffixed with _temp to avoid conflict with metrics in processor crate)
```
banool, line 42: Does it matter if they clash, don't they still mean the same thing?

rust/aptos-indexer-transaction-stream/src/utils/counters.rs
```diff
@@ -0,0 +1,104 @@
...
+    register_gauge_vec, register_int_counter_vec, register_int_gauge_vec, GaugeVec, IntCounterVec,
+    IntGaugeVec,
+};
+
+pub enum ProcessorStep {
+    ReceivedTxnsFromGrpc,
+    // Received transactions from GRPC. Sending transactions to channel.
+    ProcessedBatch,
+    // Processor finished processing one batch of transaction
+    ProcessedMultipleBatches, // Processor finished processing multiple batches of transactions
```
banool, line 15: Can we do this instead:
```
pub enum ProcessorStep {
    /// Received transactions from GRPC. Sending transactions to channel.
    ReceivedTxnsFromGrpc,
    /// Processor finished processing one batch of transaction
    ProcessedBatch,
    /// Processor finished processing multiple batches of transactions
    ProcessedMultipleBatches,
}
```

banool: Also what do we mean by "Sending transactions to channel.". Did that happen as part of this step or is that what happens next?

rust/processor/src/worker.rs
```diff
@@ -214,7 +214,7 @@ impl Worker {
                 "[Parser] Starting fetcher thread"
             );
 
-            crate::grpc_stream::create_fetcher_loop(
+            aptos_indexer_transaction_stream::transaction_stream::create_fetcher_loop(
```
banool, line 217: Maybe `create_fetcher_loop` should be lifted to the root of the crate.

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
```diff
@@ -296,14 +299,7 @@ pub async fn get_chain_id(
...
 
-/// Gets a batch of transactions from the stream. Batch size is set in the grpc server.
-/// The number of batches depends on our config
-/// There could be several special scenarios:
-/// 1. If we lose the connection, we will try reconnecting X times within Y seconds before crashing.
-/// 2. If we specified an end version and we hit that, we will stop fetching, but we will make sure that
-///    all existing transactions are processed
-pub async fn create_fetcher_loop(
-    txn_sender: AsyncSender<TransactionsPBResponse>,
+pub struct TransactionStream {
```
banool, line 302: Doc comments at the struct and field level would be good since this is a public struct. Just for the more potentially confusing fields.

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
```diff
@@ -313,93 +309,150 @@ pub async fn create_fetcher_loop(
...
-        end_version = request_ending_version,
-        "[Parser] Successfully connected to GRPC stream",
-    );
+    resp_stream: Option<Streaming<TransactionsResponse>>,
+    connection_id: Option<String>,
+    next_version_to_fetch: u64,
+    reconnection_retries: u64,
+    last_fetched_version: i64,
+    fetch_ma: MovingAverage,
+}
```
banool, line 320: Maybe this change doesn't belong in this PR but I'd like us to think about how we would actually configure this from a config file. Half the fields here are just configuration, perhaps they could be moved into a `TransactionStreamConfig` struct (which would configure serde derives, have default values, etc) and then this struct has a `config: TransactionStreamConfig` field. We could then use this at a higher level rather than repeating the same fields all over again in the worker config or whatever.

This makes it much easier to using in an SDK setting, whereas today all the configuration stuff like defaults are detached from the actual place they're used.

I sort of attempted this in my SDK crate ages ago (I think I sent you the branch), you could use that for inspo.

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
```diff
@@ -313,93 +309,150 @@ pub async fn create_fetcher_loop(
...
+            resp_stream: None,
+            connection_id: None,
+            next_version_to_fetch: starting_version,
+            reconnection_retries: 0,
+            last_fetched_version: starting_version as i64 - 1,
+            fetch_ma: MovingAverage::new(3000),
+        }
+    }
+
+    pub async fn init_stream(&mut self) {
```
banool, line 362: I'm not a huge fan of this `init_stream` approach. It means we need to have fields like `resp_stream` be an `Option` and then unwrap them every time we try to use them, even though there is no case where they should ever be `None` once you've actually initiated the stream. We could make this much more type safe by just doing all this in the `new` method and making `resp_stream` not an `Option` (and therefore eliminating runtime checks like `.expect("[Parser] GRPC stream is not initialized. Did you call init_stream?")`.

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
```diff
@@ -313,93 +309,150 @@ pub async fn create_fetcher_loop(
...
+            service_type = PROCESSOR_SERVICE_TYPE,
+            stream_address = self.indexer_grpc_data_service_address.to_string(),
+            connection_id = self.connection_id,
+            start_version = self.starting_version,
+            end_version = self.request_ending_version,
+            "[Parser] Successfully connected to GRPC stream",
+        );
+    }
+
+    /// Gets a batch of transactions from the stream. Batch size is set in the grpc server.
```
banool, line 399: Again, maybe this specific PR isn't the right place for it, but we should really stop panicking everywhere. Let's make any function that could fail return a Result instead. This is the right time to make these changes, the SDK is our chance to clean up all our tech debt.

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
```diff
@@ -313,93 +309,150 @@ pub async fn create_fetcher_loop(
...
+    /// The number of batches depends on our config
+    /// There could be several special scenarios:
+    /// 1. If we lose the connection, we will try reconnecting X times within Y seconds before crashing.
+    /// 2. If we specified an end version and we hit that, we will stop fetching, but we will make sure that
+    ///    all existing transactions are processed
+    ///
+    /// Returns
```
banool, line 406: Nit, it returns `TransactionStreamOutput` where the `should_continue_fetching` is true if... blah blah.

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
```diff
@@ -582,42 +580,224 @@ pub async fn create_fetcher_loop(
...
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        if self.reconnection_retries >= RECONNECTION_MAX_RETRIES {
+            error!(
+                processor_name = self.processor_name,
+                service_type = PROCESSOR_SERVICE_TYPE,
+                stream_address = self.indexer_grpc_data_service_address.to_string(),
+                "[Parser] Reconnected more than 100 times. Will not retry.",
+            );
+            panic!("[Parser] Reconnected more than 100 times. Will not retry.")
```
banool, line 642: Same vibe here, I'd love if we didn't panic and instead returned a Result.

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
banool: To explain myself, panicking is vaguely acceptable when we're just doing it in our own service and we know it's the "top level" component of the system. The processor SDK could be used anywhere, it might just be one component of a larger system, we don't know how the user might use it. Our code shouldn't kill their entire process if it has an issue, we should let them handle it.

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
```diff
@@ -582,42 +580,224 @@ pub async fn create_fetcher_loop(
...
+            self.auth_token,
+            self.processor_name,
+        )
+        .await
+    }
+}
+
+pub async fn create_fetcher_loop(
+    txn_sender: AsyncSender<TransactionsPBResponse>,
+    indexer_grpc_data_service_address: Url,
```
banool, line 698: Same sorta thing here, this dozen args could be just `config: TransactionStreamConfig`, amazing!

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
```diff
@@ -582,42 +580,224 @@ pub async fn create_fetcher_loop(
...
+        starting_version,
+        request_ending_version,
+        auth_token,
+        processor_name.clone(),
+        transaction_filter,
+        pb_channel_txn_chunk_size,
+    );
+
+    transaction_stream.init_stream().await;
+    let mut send_ma = MovingAverage::new(3000);
```
banool, line 726: Should this 3k value be configurable? (with a default).

banool (changes requested):
Apologies if I left comments for things that are just copy pasted. If this isn't the right place to address them, please open issues / a tracker item for those rather than fixing them here.

rust/aptos-indexer-transaction-stream/src/utils/counters.rs
rtso: It will try to register the metric with the same name and will error. I can't import the metrics from processor crate because of cyclic dep.

rust/aptos-indexer-transaction-stream/src/utils/counters.rs
rtso: `Sending transactions to channel` is what happens next. This is copy/pasted from the processors crate. We'll redo all these metrics as part of SDK, so this is just throwaway for backwards compatibility for our dashboards.

CapCap: Can we kill this enum and giant, single overloaded logging method as a pattern? It's quite awful and confusing hahaha

rust/aptos-indexer-transaction-stream/src/utils/counters.rs
CapCap: I suggest renaming them and changing them- ideally each crate is responsible for emitting its own metrics; if anything, we can allow for a "metrics_prefix" and then allow dynamically creating them (instead of a single global lazy static) per instance of the streamer. 

We can then create nice and specific metrics- maybe an "items received", and then a "filtered" either book or separate series, etc

Ie there should not be any overlap between crates and the SDK because the SDK will theoretically never have the same context on what's happening inside a given step as the step does (especially for complicated things like this)

CapCap: Ie we should move away from thinking "this is the processor GRPC client" and think of it as "this is the way anyone in any application, even python, should use our GRPC APIs, and will put it on crates.io and it'll go viral and the world will love us"- I find that generalized approach helps create nice, hermetic, batteries-included crates, vs crates that rely on specific behavior from other crates

CapCap: Also duplicate metrics- ie the GRPC emitting channel size and also processor with the instrumented channel- it's fine :-)

rust/aptos-indexer-transaction-stream/src/utils/counters.rs
rtso: > giant, single overloaded logging method as a pattern

We can kill it in this create. For more context, "giant, single overloaded logging method as a pattern" got introduced in the indexer grpc stack to reduce duplicated code and make sure we are logging all the same metrics in every relevant place. I'm open to other suggestions on how to do this :)

rust/aptos-indexer-transaction-stream/src/utils/counters.rs
banool: Got it, didn't realize you'd be using this crate alongside the old code and the registration itself would clash.

These clients support setting a prefix. If we don't do that today, I agree we should. Example from API Gateway: https://github.com/aptos-labs/api-gateway/blob/fa335677d1c80de9dc103396d563d4e47006a661/src/metrics/labels.rs#L37.

rust/aptos-indexer-transaction-stream/src/utils/counters.rs
banool: I actually don't mind this enum ¯\_(ツ)_/¯. I like it for the reason Renee mentions, it's easier to do things consistently. If anything I'd go further and wrap this in a struct that also contains all the relevant labels, and then some methods on that struct bump the relevant counts (so the counters become private to this file). You can see the API Gateway code for some inspiration on that: https://github.com/aptos-labs/api-gateway/blob/fa335677d1c80de9dc103396d563d4e47006a661/src/metrics/labels.rs#L134-L135. In this case we only bump the relevant metrics when the struct is dropped, but you could also have explicit functions for it (which makes more sense in this case).

rust/aptos-indexer-transaction-stream/src/transaction_stream.rs
rtso: Removed panics everywhere and replaced them with Result

PR #407, opened 2024-06-14
maintain backward compatibility

bowenyang007:
Add default to value. Confirmed that this doesn't break even when config is missing.


banool (approved):

yuunlimm:
Thanks for handling this!

PR #411, opened 2024-06-14
processor proto update.

larry-aptos:
* Add support to new transaction type `BlockEpilogueTransaction`, in details:
  * new transaction is indexed under `Transactions` table
    * Transaction data will not be indexed in any table, i.e.,
      * fields will be ignored.  
```
"block_end_info": {
  "block_gas_limit_reached": false,
  "block_output_limit_reached": false,
  "block_effective_block_gas_units": 0,
  "block_approx_output_size": 1244
},
```   

## next step

* update the dependency to use a tagged version

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -19,7 +19,7 @@ aptos-moving-average = { path = "moving-average" }
 
 ahash = { version = "0.8.7", features = ["serde"] }
 anyhow = "1.0.62"
-aptos-protos = { git = "https://github.com/aptos-labs/aptos-core.git", tag = "d76b5bb423b78b2b9affc72d3853f0d973d3f11f" }
+aptos-protos = { git = "https://github.com/aptos-labs/aptos-core.git", rev = "d76b5bb423b78b2b9affc72d3853f0d973d3f11f" }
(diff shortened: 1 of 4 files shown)


rust/Cargo.toml
```diff
@@ -19,7 +19,7 @@ aptos-moving-average = { path = "moving-average" }
 
 ahash = { version = "0.8.7", features = ["serde"] }
 anyhow = "1.0.62"
-aptos-protos = { git = "https://github.com/aptos-labs/aptos-core.git", tag = "aptos-node-v1.12.1" }
+aptos-protos = { git = "https://github.com/aptos-labs/aptos-core.git", rev = "d76b5bb423b78b2b9affc72d3853f0d973d3f11f" }
```
banool, line 22: Not a blocker but any reason we're not using a tag now?

banool (approved):

rust/Cargo.toml
larry-aptos: it's in devnet; tag will only be available for testnet/mainet release candidate.

1. we can change back to tag once it's in testnet.
2. we can request devnet tag to make life easier.

PR #417, opened 2024-06-20
fix a bug not returning entry function id for multisig

yuunlimm:
bug: https://aptos-org.slack.com/archives/C03MN5F7WUV/p1718806535035359 
we weren't returning the string in the match statement

Test plan
Screenshot 2024-06-20 at 2 59 32 PM
Screenshot 2024-06-20 at 2 59 28 PM


rtso (approved):
nice fix!

rtso:
@mizu-flex yep we're gonna run the backfill for mainnet and testnet today!

PR #420, opened 2024-06-23
Add Rust back

banool:


rust/processor/src/db/common/models/default_models/parquet_transactions.rs
```diff
@@ -0,0 +1,381 @@
...
+impl Transaction {
+    fn from_transaction_info(
+        info: &TransactionInfo,
+        txn_version: i64,
+        epoch: i64,
+        block_height: i64,
+    ) -> Self {
+        Self {
+            txn_version,
+            block_height,
```
yuunlimm, line 95: Not 100% sure the use case but if possible, it'd be good to just take in the owned object, it'd be more efficient. No worries if it doesn't make sense for this use case.

rust/processor/src/db/common/models/default_models/parquet_transactions.rs
```diff
@@ -0,0 +1,381 @@
...
+    const TABLE_NAME: &'static str = "transactions";
+}
+
+impl HasVersion for Transaction {
+    fn version(&self) -> i64 {
+        self.txn_version
+    }
+}
+
+impl Default for Transaction {
```
yuunlimm, line 61: How come we have to implement Default? There is no "default transaction" so I'm not sure it makes sense to implement Default for Transaction. Same comment for many of these. If some trait / library is forcing us to implement Default, we should at least use unreachable so we never use it, or if something does use it, for the fields we should use default, like txn_version: Default::default(), event_root_hash: Default::default(), etc.

Member
Author
@[yuunlimm](https://github.com/yuunlimm) yuunlimm [3 days ago](https://github.com/aptos-labs/aptos-indexer-processors/pull/390#discussion_r1649266202) • 
there is a usecase, where we have derive the parquet schema from the instance, and the we need to know this schema in advance. please take a look at hasParquetSchema trait in the generic_parquet_processor.rs. and we use this schema() to construct parquetHandler for each struct type. that's the only usecase of default but I am not too sure if there is any other workaround.

rust/processor/src/worker.rs
```diff
@@ -0,0 +1,894 @@
...
+            service_type = PROCESSOR_SERVICE_TYPE,
+            stream_address = self.indexer_grpc_data_service_address.as_str(),
+            concurrent_tasks,
+            "[Parser] Spawning concurrent parallel processor tasks",
+        );
+
+        let mut processor_tasks = vec![fetcher_task];
+        for task_index in 0..concurrent_tasks {
+            let join_handle: JoinHandle<()> = self
+                .launch_processor_task(
```
yuunlimm, line 339: I see this function takes two Options, one for the default sender and one for the parquet sender. I see we always have exactly one Some and one Option. Feels like launch_processor_task could be made more safe by encoding this in the type system, e.g. by using an enum:

enum GapDetectorSender {
    Normal(AsyncSender<ProcessingResult>),
    Parquet(AsyncSender<ProcessingResult>),
}

rust/processor/src/db/common/models/default_models/parquet_move_resources.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    pub fn from_delete_resource(
+        delete_resource: &DeleteResource,
+        write_set_change_index: i64,
+        txn_version: i64,
+        block_height: i64,
+        block_timestamp: chrono::NaiveDateTime,
+    ) -> Self {
```
yuunlimm, line 93: ) -> Self {
        let parsed_data = Self::convert_move_struct_tag(
            delete_resource
                .r#type
                .as_ref()
                .expect("MoveStructTag Not Exists."),
        );

rust/processor/src/db/common/models/default_models/parquet_move_resources.rs
```diff
@@ -0,0 +1,153 @@
...
+                    Ok(Some(
+                        serde_json::to_string(move_type).context("Failed to parse move type")?,
+                    ))
+                })
+                .collect::<Result<Option<String>>>()
+                .unwrap_or(None),
+        }
+    }
+
+    pub fn get_outer_type_from_resource(write_resource: &WriteResource) -> String {
```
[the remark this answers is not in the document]
yuunlimm: `get_outer_type_from_resource` isn't being used anywhere, will remove.  and made convert_move_struct_tag freestadanding

PR #423, opened 2024-06-24
handle payload filtered txns for default processor

yuunlimm:
https://aptos-org.slack.com/archives/C07A7CWF788

### Test plan
Screenshot 2024-06-24 at 3 44 53 PM

tested locally.

### DIFF
--- a/rust/processor/src/db/common/models/default_models/transactions.rs
+++ b/rust/processor/src/db/common/models/default_models/transactions.rs
@@ -202,8 +202,8 @@ impl Transaction {
 
                 let (payload_cleaned, payload_type) = match request.payload.as_ref() {
                     Some(payload) => {
-                        let payload_cleaned = get_clean_payload(&payload, version);
-                        (payload_cleaned, Some(get_payload_type(&payload)))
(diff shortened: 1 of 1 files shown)


rust/processor/src/db/common/models/default_models/transactions.rs
```diff
@@ -195,20 +195,34 @@ impl Transaction {
...
+                let mut payload_cleaned = None;
+                let (payload, payload_type) = match request.payload.as_ref() {
+                    Some(payload) => {
+                        payload_cleaned = get_clean_payload(&payload, version);
+                        (payload_cleaned, Some(get_payload_type(&payload)))
+                    },
+                    None => (None, None),
+                };
+
+                let (payload_cleaned, payload_type) = request
```
rtso, line 212: nit: combine these 2 branches into single one

rtso: actually these seem to be copies of the same code?

yuunlimm: removed unused

rtso (approved):

PR #428, opened 2024-06-26
use option to replace expect with context and avoid creating two gap detectors

yuunlimm:
### Summary
- made bucket root to be configurable for each stage.
- fix gap detector issue where the parquet gap detector wasn't updating its processor status to db.
- move unrelated functions in MoveResource to outside of the struct impl
- removed unnecessary fields in the parquetProcessingResult

### Test Plan
Tested locally and verified status being updated
Screenshot 2024-06-25 at 4 54 51 PM

### DIFF
--- a/rust/processor/src/bq_analytics/gcs_handler.rs
+++ b/rust/processor/src/bq_analytics/gcs_handler.rs
@@ -13,7 +13,6 @@ use tokio::{
     time::{sleep, timeout, Duration},
 };
 use tracing::{debug, error, info};
-const BUCKET_REGULAR_TRAFFIC: &str = "devnet-airflow-continue";
 const MAX_RETRIES: usize = 3;
 const INITIAL_DELAY_MS: u64 = 500;
 const TIMEOUT_SECONDS: u64 = 300;
@@ -22,6 +21,7 @@ pub async fn upload_parquet_to_gcs(
     file_path: &PathBuf,
     table_name: &str,
     bucket_name: &str,
+    bucket_root: &str,
 ) -> Result<(), ParquetProcessorError> {
     let mut file = TokioFile::open(&file_path)
         .await
@@ -54,13 +54,8 @@ pub async fn upload_parquet_to_gcs(
     let highwater_s = start_of_month.timestamp_millis();
     let highwater_ms = now.timestamp_millis();
     let counter = 0; // THIS NEED TO BE REPLACED OR REIMPLEMENTED WITH AN ACTUAL LOGIC TO ENSURE FILE UNIQUENESS.
-    let object_name: PathBuf = generate_parquet_file_path(
-        BUCKET_REGULAR_TRAFFIC,
-        table_name,
-        highwater_s,
-        highwater_ms,
-        counter,
-    );
+    let object_name: PathBuf =
+        generate_parquet_file_path(bucket_root, table_name, highwater_s, highwater_ms, counter);
 
     let file_name = object_name.to_str().unwrap().to_owned();
     let upload_type: UploadType = UploadType::Simple(Media::new(file_name.clone()));
--- a/rust/processor/src/bq_analytics/gcs_handler.rs
+++ b/rust/processor/src/bq_analytics/gcs_handler.rs
@@ -6,7 +6,7 @@ use google_cloud_storage::{
     http::objects::upload::{Media, UploadObjectRequest, UploadType},
 };
 use hyper::Body;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use tokio::io::AsyncReadExt; // for read_to_end()
 use tokio::{
     fs::File as TokioFile,
@@ -18,10 +18,10 @@ const INITIAL_DELAY_MS: u64 = 500;
 const TIMEOUT_SECONDS: u64 = 300;
 pub async fn upload_parquet_to_gcs(
     client: &GCSClient,
-    file_path: &PathBuf,
+    file_path: &Path,
     table_name: &str,
     bucket_name: &str,
-    bucket_root: &str,
+    bucket_root: &Path,
 ) -> Result<(), ParquetProcessorError> {
     let mut file = TokioFile::open(&file_path)
         .await
@@ -103,14 +103,14 @@ pub async fn upload_parquet_to_gcs(
 }
 
 fn generate_parquet_file_path(
-    gcs_bucket_root: &str,
+    gcs_bucket_root: &Path,
     table: &str,
     highwater_s: i64,
     highwater_ms: i64,
     counter: u32,
 ) -> PathBuf {
-    PathBuf::from(format!(
-        "{}/{}/{}/{}_{}.parquet",
-        gcs_bucket_root, table, highwater_s, highwater_ms, counter
+    gcs_bucket_root.join(format!(
+        "{}/{}/{}_{}.parquet",
+        table, highwater_s, highwater_ms, counter
     ))
 }
--- a/rust/processor/src/bq_analytics/generic_parquet_processor.rs
+++ b/rust/processor/src/bq_analytics/generic_parquet_processor.rs
@@ -117,7 +117,7 @@ where
         bucket_name: String,
         bucket_root: String,
         gap_detector_sender: kanal::AsyncSender<ProcessingResult>,
-        schema: Arc<parquet::schema::types::Type>,
+        schema: Arc<Type>,
     ) -> Result<Self> {
         // had to append unique id to avoid concurrent write issues
         let file_path = format!("{}_{}.parquet", ParquetType::TABLE_NAME, Uuid::new_v4());
@@ -158,9 +158,9 @@ where
         // for now, it's okay to go little above the buffer_size, given that we will keep max size as 200 MB
         if self.buffer_size_bytes >= max_buffer_size {
             let start_version = self.buffer.first().unwrap().version();
-            let end_version = self.buffer.last().unwrap().version();
-            let last_transaction_timestamp =
-                naive_datetime_to_timestamp(self.buffer.last().unwrap().get_timestamp());
+            let last = self.buffer.last().unwrap();
+            let end_version = last.version();
+            let last_transaction_timestamp = naive_datetime_to_timestamp(last.get_timestamp());
 
             let txn_version_to_struct_count = process_struct_count_map(
                 &self.buffer,
@@ -190,12 +190,13 @@ where
                 end_version = end_version,
                 "Max buffer size reached, uploading to GCS."
             );
+            let bucket_root = PathBuf::from(&self.bucket_root);
             let upload_result = upload_parquet_to_gcs(
                 gcs_client,
                 &new_file_path,
                 ParquetType::TABLE_NAME,
                 &self.bucket_name,
-                &self.bucket_root,
+                &bucket_root,
             )
             .await;
             self.buffer_size_bytes = 0;
--- a/rust/processor/src/bq_analytics/generic_parquet_processor.rs
+++ b/rust/processor/src/bq_analytics/generic_parquet_processor.rs
@@ -2,7 +2,10 @@ use super::ParquetProcessingResult;
 use crate::{
     bq_analytics::gcs_handler::upload_parquet_to_gcs,
     gap_detectors::ProcessingResult,
-    utils::counters::{PARQUET_HANDLER_BUFFER_SIZE, PARQUET_STRUCT_SIZE},
+    utils::{
+        counters::{PARQUET_HANDLER_BUFFER_SIZE, PARQUET_STRUCT_SIZE},
+        util::naive_datetime_to_timestamp,
+    },
 };
 use ahash::AHashMap;
 use allocative::Allocative;
@@ -24,9 +27,6 @@ use uuid::Uuid;
 #[derive(Debug, Default, Clone)]
 pub struct ParquetDataGeneric<ParquetType> {
     pub data: Vec<ParquetType>,
-    pub first_txn_version: u64,
-    pub last_txn_version: u64,
-    pub last_transaction_timestamp: Option<aptos_protos::util::timestamp::Timestamp>,
     pub transaction_version_to_struct_count: AHashMap<i64, i64>,
 }
 
@@ -40,6 +40,10 @@ pub trait HasVersion {
 
 pub trait HasParquetSchema {
     fn schema() -> Arc<parquet::schema::types::Type>;
+}
+
+pub trait GetTimeStamp {
+    fn get_timestamp(&self) -> chrono::NaiveDateTime;
 }
 
 /// Auto-implement this for all types that implement `Default` and `RecordWriter`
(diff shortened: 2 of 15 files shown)


ying-w:
can you also fix the first folder name? right now it's `devnet-airflow-continue`, I used it to differentiate between routine/continue (get next versions) vs backfill

rust/processor/src/bq_analytics/gcs_handler.rs
```diff
@@ -13,7 +13,6 @@ use tokio::{
     time::{sleep, timeout, Duration},
 };
 use tracing::{debug, error, info};
-const BUCKET_REGULAR_TRAFFIC: &str = "devnet-airflow-continue";
```
ying-w, line 16: i used this to differentiate between routine/continue (get next versions) vs backfill/test. Other way is to use different project

yuunlimm: let's keep this way? since this would be more straight forward unless using different project gives more benefits.

rust/processor/src/gap_detectors/mod.rs
```diff
@@ -41,6 +49,7 @@ pub async fn create_gap_detector_status_tracker_loop(
     processor: Processor,
     starting_version: u64,
     gap_detection_batch_size: u64,
+    is_parquet_processor: bool,
```
rtso, line 52: Instead of this bool flag, we could make it take in a GapDetectorTrait

rust/processor/src/gap_detectors/mod.rs
```diff
@@ -132,24 +144,20 @@ pub async fn create_gap_detector_status_tracker_loop(
                                     // We don't panic as everything downstream will panic if it doesn't work/receive
                                 }
 
-                                if let Some(res_last_success_batch) = res.last_success_batch {
```
rtso, line 135: Why do we need to remove this?

rust/processor/src/gap_detectors/mod.rs
```diff
@@ -28,9 +28,17 @@ pub enum GapDetector {
 
 #[enum_dispatch(GapDetectorTrait)]
 pub enum GapDetectorResult {
-    DefaultGapDetectorResult,
-    ParquetFileGapDetectorResult,
+    DefaultGapDetectorResult(DefaultGapDetectorResult),
+    ParquetFileGapDetectorResult(ParquetFileGapDetectorResult),
```
rtso, line 32: This is annoying that we have to use an enum like this to handle both types of processors... hopefully in SDK we can create a DefaultVersionTracker and a ParquetVersionTracker

rust/processor/src/gap_detectors/mod.rs
yuunlimm: I've decided to remove the last_success_batch field from the DefaultGapDetectorResult due to the complexity and limited utility it offered in tracking progress within the parquet processor. Given the processor's architecture, where 10 tasks each manage a buffer containing distinct transaction ranges, the end_version of a batch isn't necessarily sequential. This non-sequential nature made it challenging to effectively use end_version and last_transaction_timestamp for status updates.

Instead, I'm proposing that we update the processor status using next_version_process - 1. This approach aligns better with the purpose of the status ensuring data integrity. 

But as I am writing this, we are running 10 tasks only for backfilling, but I tuned to 1 task for regular traffic, which doesn't apply to the case above. let me come back with a better one that is applicable to both cases.

rust/processor/src/gap_detectors/mod.rs
yuunlimm: Agreed!! need to use so many enums to handle both types. Looking forward to cleaning this up using SDK !

rust/processor/src/gap_detectors/mod.rs
yuunlimm: I thought about it, maybe keep this behaviorfor now since backfilling is our priority, and then improve later to support both cases as a follow-up?

rust/processor/src/bq_analytics/gcs_handler.rs
banool: idk for sure the situation but in general it'd be better to rely on metadata about the deployment (e.g. the configuration name, project, etc), just less duplication that way.

rust/processor/src/bq_analytics/gcs_handler.rs
```diff
@@ -22,6 +21,7 @@ pub async fn upload_parquet_to_gcs(
     file_path: &PathBuf,
     table_name: &str,
     bucket_name: &str,
+    bucket_root: &str,
```
banool, line 24: Let's make this a Path / PathBuf if possible, less room for error. For example, if someone starts / ends this with a slash it'll cause issues.

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -153,6 +159,8 @@ where
         if self.buffer_size_bytes >= max_buffer_size {
             let start_version = self.buffer.first().unwrap().version();
             let end_version = self.buffer.last().unwrap().version();
+            let last_transaction_timestamp =
+                naive_datetime_to_timestamp(self.buffer.last().unwrap().get_timestamp());
```
banool, line 163: Nit
```suggestion
            let last =  self.buffer.last().unwrap();
            let end_version = last.version();
            let last_transaction_timestamp =
                naive_datetime_to_timestamp(last.get_timestamp());
```

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -153,6 +159,8 @@ where
         if self.buffer_size_bytes >= max_buffer_size {
             let start_version = self.buffer.first().unwrap().version();
             let end_version = self.buffer.last().unwrap().version();
```
banool, line 161: We know for sure that the buffer has at least one item in it?

rust/processor/src/db/common/models/default_models/parquet_write_set_changes.rs
```diff
@@ -58,98 +58,117 @@ impl Default for WriteSetChange {
...
                     inner,
                     write_set_change_index,
                     txn_version,
                     block_height,
                     block_timestamp,
-                )),
-            ),
+                );
+
+                if let Some(resource) = resource_option.unwrap() {
```
banool, line 132: Are we confident this will never fail? If so, can we use `expect` with a message that explains why, plus a comment? Unless we're 100% certain, it'd be better to just `.context("explanation")?` and bubble up the error, since this function also returns a Result.

rust/processor/src/db/common/models/default_models/parquet_write_set_changes.rs
```diff
@@ -58,98 +58,117 @@ impl Default for WriteSetChange {
...
+                    return Ok(Some((
+                        Self {
+                            txn_version,
+                            state_key_hash: standardize_address(
+                                hex::encode(inner.state_key_hash.as_slice()).as_str(),
```
banool, line 137: We should make a version of `standardize_address` that takes in a slice, so we don't have to do this hex::encode first everywhere.

rust/processor/src/db/common/models/default_models/parquet_write_set_changes.rs
```diff
@@ -58,98 +58,117 @@ impl Default for WriteSetChange {
...
+                            ),
+                            block_height,
+                            change_type,
+                            resource_address: standardize_address(&inner.address.to_string()),
```
banool, line 141: Is the `to_string` necessary? I pulled the PR and it seems not.

rust/processor/src/db/common/models/default_models/parquet_write_set_changes.rs
```diff
@@ -185,7 +204,7 @@ impl WriteSetChange {
                     block_height,
                     block_timestamp,
                 );
-                (
+                return Ok(Some((
```
banool, line 207: Nit, the explicit return is not necessary here, I'm surprised clippy didn't complain, I guess we don't have that check configured.

rust/processor/src/db/common/models/default_models/parquet_write_set_changes.rs
```diff
@@ -209,21 +228,32 @@ impl WriteSetChange {
...
-                )
+                ) {
+                    Ok(Some((change, detail))) => Some((change, detail)),
+                    Ok(None) => None,
+                    Err(e) => {
+                        tracing::error!(
+                            "Failed to convert write set change: {:?} with error: {:?}",
+                            write_set_change,
+                            e
+                        );
```
banool, line 249: Is this safe? I feel like normally in the processors we prefer to fail rather than log an error and move on. This way the processor can "succeed"  but drop write set changes. If there is a good reason why this is okay here, could you leave a comment?

rust/processor/src/gap_detectors/gap_detector.rs
```diff
@@ -24,7 +20,10 @@ pub struct DefaultGapDetectorResult {
 }
 
 impl GapDetectorTrait for DefaultGapDetector {
-    fn process_versions(&mut self, result: ProcessingResult) -> anyhow::Result<GapDetectorResult> {
+    fn process_versions(
+        &mut self,
+        result: ProcessingResult,
+    ) -> anyhow::Result<GapDetectorResult, Box<dyn std::error::Error + Send + Sync>> {
```
banool, line 26: How come this error type is required? 

If it is required, might as well just use `Result`, `anyhow::Result` is just an alias for a result where the error type is `anyhow::Error`.

rust/processor/src/gap_detectors/mod.rs
```diff
@@ -28,18 +28,26 @@ pub enum GapDetector {
 
 #[enum_dispatch(GapDetectorTrait)]
 pub enum GapDetectorResult {
-    DefaultGapDetectorResult,
-    ParquetFileGapDetectorResult,
+    DefaultGapDetectorResult(DefaultGapDetectorResult),
```
banool, line 31: I can't comment on it but on line 23, is it meaningful to have `#[enum_dispatch(GapDetectorTrait)]` on `GapDetector`? It's an enum without variants, what is it dispatching to? I tried removing it locally and the code was still fine.

rust/processor/src/gap_detectors/mod.rs
```diff
@@ -28,18 +28,26 @@ pub enum GapDetector {
...
+    ParquetFileGapDetectorResult(ParquetFileGapDetectorResult),
 }
+
+pub trait GapDetectorTrait: Send {
+    fn process_versions(
```
banool, line 36: For traits we should have comments on the trait and the methods of that trait explaining how implementations of the trait should behave.

rust/processor/src/gap_detectors/mod.rs
banool: I'm bouncing back and forth between the new and old code and can't quite figure out what changed. At least here it seems like we're still logging the same stuff as before, but more reliably now, since the fields are always set. With the new code the structs seem less deep / cleaner. tldr I need to read the code further to figure out what's going on here, but don't let it block you, I'll catch up later.

rust/processor/src/worker.rs
```diff
@@ -298,44 +301,42 @@ impl Worker {
...
-                let processor = build_processor(
-                    &self.processor_config,
-                    self.per_table_chunk_sizes.clone(),
-                    self.deprecated_tables,
-                    self.db_pool.clone(),
-                    Some(gap_detector_sender.clone()),
-                );
-                let gap_detection_batch_size: u64 = self.parquet_gap_detection_batch_size;
+        let is_parquet_processor = self.processor_config.is_parquet_processor();
+        let (processor, gap_detection_batch_size) = if is_parquet_processor {
```
banool, line 305: Nit, you could just return `maybe_gap_detector_sender` and `gap_detection_batch_size` from this block and then only have `build_processor` in one place.

rust/processor/src/worker.rs
```diff
@@ -298,44 +301,42 @@ impl Worker {
...
-                    processor,
-                    gap_detection_batch_size,
-                    Some(gap_detector_sender),
-                )
-            };
+        let gap_detector: Box<dyn GapDetectorTrait + Send> = if is_parquet_processor {
+            Box::new(ParquetFileGapDetector::new(starting_version))
+        } else {
+            Box::new(DefaultGapDetector::new(starting_version))
+        };
```
banool, line 333: Rather than using a box, make use of enum_dispatch on GapDetector. That way you can just pass around GapDetector instead of a trait object. Nicer types and more efficient.

rust/processor/src/gap_detectors/mod.rs
banool: Ah actually, seems like we should expand the use of this enum. See my comment on line 329 in rust/processor/src/worker.rs.

rust/processor/src/worker.rs
```diff
@@ -298,44 +301,42 @@ impl Worker {
...
-                let processor = build_processor(
-                    &self.processor_config,
-                    self.per_table_chunk_sizes.clone(),
-                    self.deprecated_tables,
-                    self.db_pool.clone(),
-                    None,
-                );
-                let gap_detection_batch_size = self.gap_detection_batch_size;
+            (processor, gap_detection_batch_size)
+        };
```
banool, line 327: If it makes sense, it might be nice to use launch_processor_task to reduce code dupe. Nbd though.

banool (changes requested):

rust/processor/src/db/common/models/default_models/parquet_write_set_changes.rs
yuunlimm: I think if there isn't, that's a problematic txn. and it seems like we have been panicing if there is a problematic one as I see expect in the default model. (move_resource.rs, and write_set_changes.rs). 

but if it makes sense to use `context`  here as it already returns a Result.

rust/processor/src/bq_analytics/generic_parquet_processor.rs
yuunlimm: it should, and I have a follow-up pr that prevents this happening too

rust/processor/src/bq_analytics/gcs_handler.rs
yuunlimm: good call!

rust/processor/src/db/common/models/default_models/parquet_write_set_changes.rs
yuunlimm: changed to fail rather than log an error and move on.

rust/processor/src/worker.rs
yuunlimm: Used GapDetector instead of a trait object

rust/processor/src/worker.rs
yuunlimm: this will be a lot cleaner once we have SDK lol.

banool (approved):
Nice!

PR #429, opened 2024-06-26
add skip db write flags for tables in parquet migration tranche 2 and 3

yuunlimm:
### Summary
We have categorized our tables into four tranches to manage database better writes and prioritize our parquet processing efforts. The goal is to enable skipping database writes for tables in Tranche 2 and Tranche 3 by adding appropriate flags in each processor responsible for these models.

Here is the list of tables we have: https://docs.google.com/spreadsheets/d/1gSuE27GsGKLCBLOxhlzonvHh_bzF5i1Uzx27Mf61uh0/edit?gid=0#gid=0

### Tranche Categories:
   Tranche 1: tables that are huge and high priority 
   Tranche 2: tables that are huge and low usage
   Tranche 3: v1 tables to replace with v2 tables.
   Tranche 4: tables to revisit. 

This enables skipping db writes for tables included in Tranche 2 and 3 by adding a flag in each processor responsible for each model. 

We can add table names in the `deprecated_tables` field in the config as needed

## Revision 2
- excluded the flags for transaction_metadata_processor as clients can stop running this processor completely, and these were for analytics purposes only. 

### test plan
Screenshot 2024-06-26 at 1 09 21 PM
Screenshot 2024-06-26 at 1 09 17 PM

tested when CURRENT_COIN_BALANCES was added

### DIFF
--- a/rust/processor/src/processors/coin_processor.rs
+++ b/rust/processor/src/processors/coin_processor.rs
@@ -14,6 +14,7 @@ use crate::{
     gap_detectors::ProcessingResult,
     schema,
     utils::database::{execute_in_chunks, get_config_table_chunk_size, ArcDbPool},
+    worker::TableFlags,
 };
 use ahash::AHashMap;
 use anyhow::{bail, Context};
@@ -30,13 +31,19 @@ use tracing::error;
 pub struct CoinProcessor {
     connection_pool: ArcDbPool,
     per_table_chunk_sizes: AHashMap<String, usize>,
+    deprecated_tables: TableFlags,
 }
 
 impl CoinProcessor {
-    pub fn new(connection_pool: ArcDbPool, per_table_chunk_sizes: AHashMap<String, usize>) -> Self {
+    pub fn new(
+        connection_pool: ArcDbPool,
+        per_table_chunk_sizes: AHashMap<String, usize>,
+        deprecated_tables: TableFlags,
+    ) -> Self {
         Self {
             connection_pool,
             per_table_chunk_sizes,
+            deprecated_tables,
         }
     }
 }
@@ -216,10 +223,10 @@ impl ProcessorTrait for CoinProcessor {
         let last_transaction_timestamp = transactions.last().unwrap().timestamp.clone();
 
         let (
-            all_coin_activities,
+            mut all_coin_activities,
             all_coin_infos,
-            all_coin_balances,
-            all_current_coin_balances,
+            mut all_coin_balances,
+            mut all_current_coin_balances,
         ) = tokio::task::spawn_blocking(move || {
             let mut all_coin_activities = vec![];
             let mut all_coin_balances = vec![];
@@ -258,6 +265,19 @@ impl ProcessorTrait for CoinProcessor {
         })
         .await
         .context("spawn_blocking for CoinProcessor thread failed")?;
+
+        if self.deprecated_tables.contains(TableFlags::COIN_ACTIVITIES) {
+            all_coin_activities.clear();
+        }
+        if self.deprecated_tables.contains(TableFlags::COIN_BALANCES) {
+            all_coin_balances.clear();
+        }
+        if self
+            .deprecated_tables
+            .contains(TableFlags::CURRENT_COIN_BALANCES)
+        {
+            all_current_coin_balances.clear();
+        }
 
         let processing_duration_in_secs = processing_start.elapsed().as_secs_f64();
         let db_insertion_start = std::time::Instant::now();
--- a/rust/processor/src/processors/coin_processor.rs
+++ b/rust/processor/src/processors/coin_processor.rs
@@ -14,7 +14,6 @@ use crate::{
     gap_detectors::ProcessingResult,
     schema,
     utils::database::{execute_in_chunks, get_config_table_chunk_size, ArcDbPool},
-    worker::TableFlags,
 };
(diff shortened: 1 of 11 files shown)


rust/processor/src/processors/coin_processor.rs
```diff
@@ -216,10 +223,10 @@ impl ProcessorTrait for CoinProcessor {
         let last_transaction_timestamp = transactions.last().unwrap().timestamp.clone();
 
         let (
-            all_coin_activities,
+            mut all_coin_activities,
             all_coin_infos,
```
rtso, line 227: I think we are missing coin_infos in the table deprecation plan

rust/processor/src/processors/coin_processor.rs
```diff
@@ -30,13 +31,19 @@ use tracing::error;
 pub struct CoinProcessor {
     connection_pool: ArcDbPool,
     per_table_chunk_sizes: AHashMap<String, usize>,
+    deprecated_tables: TableFlags,
 }
 
 impl CoinProcessor {
-    pub fn new(connection_pool: ArcDbPool, per_table_chunk_sizes: AHashMap<String, usize>) -> Self {
+    pub fn new(
```
rtso, line 38: Hmm do we need the flags in coin_processor and token_processor? I think we can tell them to stop running it and only run fungible_asset_processor and token_v2_processor, respectively.

rust/processor/src/processors/fungible_asset_processor.rs
```diff
@@ -367,6 +374,21 @@ impl ProcessorTrait for FungibleAssetProcessor {
...
+            .contains(TableFlags::FUNGIBLE_ASSET_BALANCES)
+        {
+            fungible_asset_balances.clear();
+        }
+
+        if self
+            .deprecated_tables
+            .contains(TableFlags::CURRENT_FUNGIBLE_ASSET_BALANCES)
+        {
+            current_fungible_asset_balances.clear();
```
rtso, line 389: Just a note: this table is on a different deprecation timeline than the other ones for the coin -> FA migration. It's fine to add the flag, but we don't want to turn it off just yet.

rust/processor/src/processors/coin_processor.rs
yuunlimm: yeah, I think we wanted to revisit this, because it has lots of usages. The notes says we can replace with fungible_asset_metadata. we can add this to the plan. then we can tell them to stop running this processor.

rust/processor/src/processors/coin_processor.rs
yuunlimm: can `current_collection_datas` and `current_token_pending_claims` be replaced with any other tables? we mark them as No in [the doc](https://docs.google.com/spreadsheets/d/1gSuE27GsGKLCBLOxhlzonvHh_bzF5i1Uzx27Mf61uh0/edit?gid=0#gid=0)

rust/processor/src/processors/fungible_asset_processor.rs
yuunlimm: we need to re-eval this table `current_fungible_asset_balances` with @ying-w

yuunlimm: we don't need this in parquet, we can delete this table. according to Ying, we can replace with fungible_asset_balances

rust/processor/src/processors/coin_processor.rs
yuunlimm: + `transaction_metadata_processor`

rust/processor/src/processors/fungible_asset_processor.rs
yuunlimm: Will start a separate migration plan for this table with the different timeline.

rust/processor/src/processors/coin_processor.rs
yuunlimm: `coin_infos` can be included in the next table deprecation plan along with `current_fungible_asset_balances`

rtso: Do we have a migration view for it? If not, let's create a new PR for it

yuunlimm: I don't think there is , are we going to include this in the current migration plan?

rtso: Oh wait we do have it, so we can include it in Tranche 2
https://github.com/aptos-labs/aptos-indexer-processors/blob/2397cd85566186f4189cbae520de3aa50cec7b5c/rust/processor/src/db/postgres/migrations/2024-05-22-200847_add_v1_migration_views/up.sql#L48

yuunlimm: oh we do, yes then I will include t he flag for this

rust/processor/src/processors/fungible_asset_processor.rs
rtso: Wait `fungible_asset_balances` is going to Parquet, but we need to keep `current_fungible_asset_balances` for now for the coin -> FA migration, which I'll deal with .

yuunlimm:
Migration view pr for `current_collection_datas` and migrating `current_token_pending_claims` logic to v2 will be followed up in separate pr

rtso (approved):

PR #437, opened 2024-07-01
add property maps indexing to collections v2

jillxuu:
Screenshot 2024-07-01 at 2 05 24 PM

backfill:
testnet: 18589
mainnet: 169280

### DIFF
--- a/rust/processor/src/db/postgres/migrations/2024-07-01-204112_add_collection_properties_to_collections/down.sql
+++ b/rust/processor/src/db/postgres/migrations/2024-07-01-204112_add_collection_properties_to_collections/down.sql
@@ -1,5 +1,5 @@
 ALTER TABLE current_collections_v2
-DROP COLUMN collection_properties;
+DROP COLUMN IF EXISTS collection_properties;
 
 ALTER TABLE collections_v2
-DROP COLUMN collection_properties;
+DROP COLUMN IF EXISTS collection_properties;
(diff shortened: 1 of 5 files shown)


rust/processor/src/db/postgres/migrations/2024-07-01-204112_add_collection_properties_to_collections/down.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE current_collections_v2
+DROP COLUMN collection_properties;
```
rtso, line 2: add IF EXISTS

rtso:
Need to also update the insert query and backfill query.

jillxuu:
> Need to also update the insert query and backfill query.

could you elaborate?

rust/processor/src/processors/token_v2_processor.rs
```diff
@@ -354,6 +354,7 @@ fn insert_current_collections_v2_query(
                 table_handle_v1.eq(excluded(table_handle_v1)),
                 token_standard.eq(excluded(token_standard)),
                 last_transaction_version.eq(excluded(last_transaction_version)),
+                collection_properties.eq(excluded(collection_properties)),
```
rtso, line 357: You also need to update insert_collections_v2_query so that it backfills the new column

jillxuu: updated!

yuunlimm (approved):

PR #444, opened 2024-07-09
enable time based parquet upload for parquet processor

yuunlimm:
## Description
- support time-based upload, which to avoid files not being uplaoded for too long and cause an alert in analytics for missing data. 
- fix to handle filtered payload, where the behavior changed now due to the newly introduced. filtering feature.
- moved parquet related processors to its own directory as we will have more parquet processors. 

## Test Plan
tested locally and parquet files uploaded successfully when even it hasn't reached the max file size but when it has elapsed over the interval duration specified. 

Screenshot 2024-07-09 at 12 20 03 PM


rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -153,82 +161,91 @@ where
...
-                        .await
-                        .expect("[Parser] Failed to send versions to gap detector");
-                    Ok(())
-                },
-                Err(e) => {
-                    error!("Failed to upload file to GCS: {}", e);
-                    Err(anyhow!("Failed to upload file to GCS: {}", e))
-                },
-            };
+            if self.buffer_size_bytes >= self.max_buffer_size {
```
larry-aptos, line 165: do i understand correctly that if buffer is not full, there is no upload?

yuunlimm: yeah. it only uploads when it reaches the max buffer size, and until then we don't upload.

larry-aptos: if the TPS is low, is there a possibility that we won't upload anything for hours?

yuunlimm: if we configure the max buffer size to be big, yes. but this change prevents this from not uploading for hours as you can specify the interval duration. for example, if we put  3600(seconds) in the config, it will check if we uploaded within the past 1 hour, if not, we upload whatever we have in the buffer.

rust/processor/src/bq_analytics/mod.rs
```diff
@@ -66,3 +85,120 @@ impl From<parquet::errors::ParquetError> for ParquetProcessorError {
...
+            .expect("Failed to create GCS client config");
+        let gcs_client = Arc::new(GCSClient::new(gcs_config));
+
+        let mut last_upload_time = Instant::now();
+
+        loop {
+            let now = Instant::now();
+            // Check if the interval has elapsed
+            if now.duration_since(last_upload_time) >= upload_interval {
+                info!("Time has elamped more since last update.");
```
larry-aptos, line 143: does `last_upload_time.elaspsed()` work here?

rust/processor/src/bq_analytics/generic_parquet_processor.rs
larry-aptos: ah i see; add some comments below

rust/processor/src/bq_analytics/mod.rs
```diff
@@ -66,3 +85,120 @@ impl From<parquet::errors::ParquetError> for ParquetProcessorError {
...
+                        last_upload_time = Instant::now();
+                    },
+                    Err(e) => {
+                        error!("Failed to upload buffer: {}", e);
+                        panic!("Error processing parquet files: {:?}", e);
+                    },
+                }
+            }
+
+            match parquet_receiver.recv().await {
```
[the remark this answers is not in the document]
yuunlimm: yeah encapsulating the logic inside parquet-manager makes more sense and wold lead to a cleaner handling loop.

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -153,82 +164,102 @@ where
...
+                if let Err(e) = self.upload_buffer(gcs_client).await {
```
larry-aptos, line 169: nit: you can add a different log here to say buffer size large and we upload; this is for easily debugging.

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -133,15 +140,19 @@ where
...
     }
 
     pub async fn handle(
         &mut self,
         gcs_client: &GCSClient,
         changes: ParquetDataGeneric<ParquetType>,
-        max_buffer_size: usize,
-    ) -> Result<()> {
+    ) -> Result<ParquetProcessingState> {
+        let mut state = ParquetProcessingState::Buffered;
```
larry-aptos, line 154: is this `state` used in handle? looks like it's always `buffered`;  maybe we can remove it and make downstream life easier?

rust/processor/src/bq_analytics/mod.rs
```diff
@@ -66,3 +84,97 @@ impl From<parquet::errors::ParquetError> for ParquetProcessorError {
...
+        let gcs_config = GcsClientConfig::default()
+            .with_auth()
+            .await
+            .expect("Failed to create GCS client config");
+        let gcs_client = Arc::new(GCSClient::new(gcs_config));
+
+        loop {
+            match parquet_receiver.recv().await {
+                Ok(txn_pb_res) => {
+                    let result = parquet_manager.handle(&gcs_client, txn_pb_res).await;
```
larry-aptos, line 139: yea this is the part you can simplify without this state... Ok/Error as state looks enough

larry-aptos (changes requested):

rust/processor/src/bq_analytics/mod.rs
yuunlimm: I initially thought knowing what has been done would give us more info when debugging. but since we have that info! when we upload I guess we can just return Ok/Error too

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -153,84 +161,106 @@ where
...
+            }
         }
 
         PARQUET_HANDLER_BUFFER_SIZE
             .with_label_values(&[ParquetType::TABLE_NAME])
             .set(self.buffer.len() as i64);
         Ok(())
     }
+
+    pub async fn upload_buffer(&mut self, gcs_client: &GCSClient) -> Result<()> {
```
larry-aptos, line 193: nit: remove `pub`? this leaves the `handle` to be the only operation for the buffer manager.

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -153,84 +161,106 @@ where
...
+
+        let txn_version_to_struct_count =
+            process_struct_count_map(&self.buffer, &mut self.transaction_version_to_struct_count);
+
+        let new_file_path: PathBuf = PathBuf::from(format!(
+            "{}_{}.parquet",
+            ParquetType::TABLE_NAME,
+            Uuid::new_v4()
+        ));
+        rename(&self.file_path, &new_file_path)?; // this fixes an issue with concurrent file access issues
```
larry-aptos, line 210: nit: document more around the file rename... what's the issue here btw?

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -153,84 +161,106 @@ where
...
+        let upload_result = upload_parquet_to_gcs(
+            gcs_client,
+            &new_file_path,
+            ParquetType::TABLE_NAME,
+            &self.bucket_name,
+            &bucket_root,
+        )
+        .await;
+        self.buffer_size_bytes = 0;
+        remove_file(&new_file_path)?;
```
larry-aptos, line 238: question: if gcs operation fails, how to recover? this reads to me that we panic, right? removal is to make sure we don't have any thing that's not cleaned up.

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -153,84 +161,106 @@ where
...
+
+        match upload_result {
```
larry-aptos, line 240: nit: flatten this with `?`

larry-aptos (changes requested):

rust/processor/src/bq_analytics/generic_parquet_processor.rs
yuunlimm: if gcs operation fails after retries, and it doesn't panic her right? it cleans up and it bubbles up the error and panic at the `create_parquet_handler_loop`. and if it fails, we should restart from where it failed.

rust/processor/src/bq_analytics/generic_parquet_processor.rs
yuunlimm: it's been a while.. but I recall that there is a issue with file not found or something like after first upload. I can revisit this later and see if we can remove this behavior

larry-aptos: can you elaborate a little? do you mean file not found right when reading into bigquery right after upload?

larry-aptos (approved):

PR #446, opened 2024-07-10
fix: primary key conflict in upsert

bbaktaeho:
You do not need to enter the column 'ON CONFLICT' to process 'DO NOTHING' when the Primary Key CONFLICT occurs.

For example, 
if you create a partition table using `pg_partman` extention in PostgreSQL, the parent table might not have a PK.

```SQL
create table test_table (id int, tt timestamp) partition by range (tt);
create table test_table_template (LIKE test_table);
alter table test_table_template add constraint test_table_pk primary key (id);

select partman.create_parent('temp.test_table', 'tt', 'native', '1 month', p_template_table := 'temp.test_table_template', p_premake := 2);

select constraint_name, constraint_type from information_schema.table_constraints where table_name = 'test_table';
-- no exists primary key
-- +---------------+---------------+
-- |constraint_name|constraint_type|
-- +---------------+---------------+

```
Create a parent table as above and inject pk into the partitioning table through the template table.
As a result, there is no PK in the parent table.

```SQL
-- success
insert into test_table (id, tt) values (1, now());

-- there is no unique or exclusion constraint matching the ON CONFLICT specification
insert into test_table (id, tt)
values (1, now()) on conflict (id) do nothing ;

-- success
insert into test_table (id, tt)
values (1, now()) on conflict do nothing ;
```

So, it looks better to use on_conflict_do_nothing() when it's PK CONFLICT.

### DIFF
--- a/rust/processor/src/processors/coin_processor.rs
+++ b/rust/processor/src/processors/coin_processor.rs
@@ -170,8 +170,7 @@ fn insert_coin_balances_query(
     (
         diesel::insert_into(schema::coin_balances::table)
             .values(items_to_insert)
-            .on_conflict((transaction_version, owner_address, coin_type_hash))
-            .do_nothing(),
+            .on_conflict_do_nothing(),
         None,
     )
 }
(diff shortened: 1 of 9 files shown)


rust/processor/src/processors/coin_processor.rs
```diff
@@ -170,8 +170,7 @@ fn insert_coin_balances_query(
     (
         diesel::insert_into(schema::coin_balances::table)
             .values(items_to_insert)
-            .on_conflict((transaction_version, owner_address, coin_type_hash))
```
larry-aptos, line 173: This is `on conflict` on a `unique index`

PR #454, opened 2024-07-11
[Validator API Fix]support validator txn with its events.

larry-aptos:
## Description

* This is to supports events from validator transactions with events migrated from BlockMetadataTransaction.

### Backfill 
* First validator transaction in mainnet: 554229017
* First validator transaction in testnet: 942054324

## Test Plan

Tested in devnet and looks good. 

image

image

### DIFF
--- a/rust/processor/src/db/common/models/stake_models/delegator_activities.rs
+++ b/rust/processor/src/db/common/models/stake_models/delegator_activities.rs
@@ -50,6 +50,7 @@ impl DelegatedStakingActivity {
         let events = match txn_data {
             TxnData::User(txn) => &txn.events,
             TxnData::BlockMetadata(txn) => &txn.events,
+            TxnData::Validator(txn) => &txn.events,
             _ => return Ok(delegator_activities),
         };
         for (index, event) in events.iter().enumerate() {
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -160,8 +160,8 @@ dependencies = [
 
 [[package]]
 name = "aptos-protos"
-version = "1.3.0"
-source = "git+https://github.com/aptos-labs/aptos-core.git?rev=d76b5bb423b78b2b9affc72d3853f0d973d3f11f#d76b5bb423b78b2b9affc72d3853f0d973d3f11f"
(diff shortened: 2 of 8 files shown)


rust/processor/src/db/common/models/stake_models/delegator_activities.rs
rtso, line 1: Besides delegated staking activities, are there other staking models that need to be indexed? I'd imagine we also need to index the wsc of validator txn's as well right?

rtso:
Could you update the description with testnet and mainnet backfill versions?

rust/processor/src/db/common/models/stake_models/delegator_activities.rs
larry-aptos: updated; all i did is for the places that requires `match TxnData::` branching logic, I added Validator condition.. 

but for `if TxnData::User`, I ignored them since they're usually specific to user transactions.

larry-aptos:
Addressed the comments; ready for another review.

rust/processor/src/db/common/models/stake_models/delegator_activities.rs
rtso: We need to parse all the wsc in ValidatorTransaction now, because there might be delegator resource changes in them. Like here: https://github.com/aptos-labs/aptos-indexer-processors/blob/98386c6a27665f6ba4e7e7caf233e3f4581f5695/rust/processor/src/db/common/models/stake_models/delegator_pools.rs#L119

larry-aptos: wait.. but this event was moved from BlockMetadtaTransaciton, does this mean we forgot to handle the writesetchanges in BlockMetadataTransaciton?

rtso: Do you have the proto txn? We can take a look at what's in it.

larry-aptos: https://gist.github.com/larry-aptos/bfb876716dc617c5717c312d2f7c8255 checking

rtso (approved):

PR #456, opened 2024-07-11
add parquet_fungible_asset_processor

yuunlimm:


rust/processor/src/processors/parquet_processors/parquet_fungible_asset_processor.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+impl ParquetFungibleAssetProcessor {
+    pub fn new(
+        connection_pool: ArcDbPool,
+        config: ParquetFungibleAssetProcessorConfig,
+        new_gap_detector_sender: AsyncSender<ProcessingResult>,
+    ) -> Self {
+        if let Some(credentials) = config.google_application_credentials.clone() {
+            std::env::set_var(GOOGLE_APPLICATION_CREDENTIALS, credentials);
```
larry-aptos, line 62: is this something we want to put in each processor or we can move to generic component?

rust/processor/src/db/common/models/fungible_asset_models/parquet_v2_fungible_asset_balances.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+impl GetTimeStamp for FungibleAssetBalance {
+    fn get_timestamp(&self) -> chrono::NaiveDateTime {
+        self.block_timestamp
+    }
+}
+
+impl FungibleAssetBalance {
```
larry-aptos, line 65: do i understand correctly that this code is duplicated but with a different output(FungibleAssetBalance in Parquet vs FungibleAssetBalance in disiel)? can we abstract the transformation?

larry-aptos (approved):

PR #457, opened 2024-07-11
add parquet_objects_processor

yuunlimm:
### Description
Add parquet object processor that handles objects table

### Test Plan
Screenshot 2024-07-11 at 2 31 07 PM

### DIFF
--- a/rust/processor/src/db/common/models/object_models/parquet_v2_objects.rs
+++ b/rust/processor/src/db/common/models/object_models/parquet_v2_objects.rs
@@ -67,22 +67,22 @@ impl Object {
     ) -> anyhow::Result<Option<(Self, CurrentObject)>> {
         let address = standardize_address(&write_resource.address.to_string());
         if let Some(object_aggregated_metadata) = object_metadata_mapping.get(&address) {
-            // do something
-            let object_with_metadata = object_aggregated_metadata.object.clone();
-            let object_core = object_with_metadata.object_core;
+            let object_with_metadata = &object_aggregated_metadata.object;
+            let object_core = &object_with_metadata.object_core;
 
             let untransferrable = if object_aggregated_metadata.untransferable.as_ref().is_some() {
                 true
             } else {
                 !object_core.allow_ungated_transfer
             };
+            let state_key_hash = &object_with_metadata.state_key_hash;
             Ok(Some((
                 Self {
                     txn_version,
                     write_set_change_index,
                     object_address: address.clone(),
                     owner_address: object_core.get_owner_address(),
-                    state_key_hash: object_with_metadata.state_key_hash.clone(),
+                    state_key_hash: state_key_hash.clone(),
                     guid_creation_num: object_core.guid_creation_num.clone().to_u64().unwrap(),
                     allow_ungated_transfer: object_core.allow_ungated_transfer,
                     is_deleted: false,
@@ -92,7 +92,7 @@ impl Object {
                 CurrentObject {
                     object_address: address,
                     owner_address: object_core.get_owner_address(),
-                    state_key_hash: object_with_metadata.state_key_hash,
+                    state_key_hash: state_key_hash.clone(),
                     allow_ungated_transfer: object_core.allow_ungated_transfer,
                     last_guid_creation_num: object_core.guid_creation_num.clone(),
                     last_transaction_version: txn_version,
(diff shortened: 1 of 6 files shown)


rust/processor/src/db/common/models/object_models/parquet_v2_objects.rs
```diff
@@ -0,0 +1,183 @@
...
+            let resource = MoveResource::from_delete_resource(
+                delete_resource,
+                0, // Placeholder, this isn't used anyway
+                txn_version,
+                0, // Placeholder, this isn't used anyway
+            );
+            let previous_object = if let Some(object) = object_mapping.get(&resource.address) {
+                object.clone()
+            } else {
+                match DefaultObject::get_current_object(
```
rtso, line 131: This tries to do a DB lookup to get prev owner, if it doesn't find the object core in the transaction. We probably don't want to keep this lookup for the Parquet processor. 

I think we're only migrating the `objects` table right? So we could just make the owner (and other columns) null. 
cc @ying-w

rust/processor/src/db/common/models/object_models/parquet_v2_objects.rs
```diff
@@ -0,0 +1,183 @@
...
+#[derive(
+    Allocative, Clone, Debug, Default, Deserialize, FieldCount, ParquetRecordWriter, Serialize,
+)]
+pub struct Object {
+    pub txn_version: i64,
+    pub write_set_change_index: i64,
+    pub object_address: String,
+    pub owner_address: String,
+    pub state_key_hash: String,
+    pub guid_creation_num: u64,
```
ying-w, line 36: what is this? @bowenyang007

rust/processor/src/processors/parquet_processors/parquet_objects_processor.rs
```diff
@@ -0,0 +1,243 @@
...
+
+            // First pass to get all the object cores
+            for wsc in changes.iter() {
+                if let Change::WriteResource(wr) = wsc.change.as_ref().unwrap() {
+                    let address = standardize_address(&wr.address.to_string());
+                    if let Some(object_with_metadata) =
+                        ObjectWithMetadata::from_write_resource(wr, txn_version).unwrap()
+                    {
+                        // Object core is the first struct that we need to get
+                        object_metadata_helper.insert(address.clone(), ObjectAggregatedData {
```
larry-aptos, line 145: I think you can do
ObjectAggregatedData {
   object: blabla,
   ...ObjectAggregatedData::default()
}

rust/processor/src/db/common/models/object_models/parquet_v2_objects.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub fn from_write_resource(
+        write_resource: &WriteResource,
+        txn_version: i64,
+        write_set_change_index: i64,
+        object_metadata_mapping: &ObjectAggregatedDataMapping,
+        block_timestamp: chrono::NaiveDateTime,
+    ) -> anyhow::Result<Option<(Self, CurrentObject)>> {
+        let address = standardize_address(&write_resource.address.to_string());
+        if let Some(object_aggregated_metadata) = object_metadata_mapping.get(&address) {
+            // do something
```
larry-aptos, line 70: nit: remove or update the comments

rust/processor/src/db/common/models/object_models/parquet_v2_objects.rs
```diff
@@ -0,0 +1,183 @@
...
+            let object_with_metadata = object_aggregated_metadata.object.clone();
```
larry-aptos, line 71: is the clone here required? is ref ok?

rust/processor/src/processors/parquet_processors/parquet_objects_processor.rs
```diff
@@ -0,0 +1,243 @@
...
+                        }
+                    },
+                    _ => {},
+                };
+            }
+        }
+
+        let objects_parquet_data = ParquetDataGeneric {
+            data: all_objects,
+            transaction_version_to_struct_count: transaction_version_to_struct_count.clone(),
```
larry-aptos, line 222: why cloning the map? we only care about the current state? it reads to me that, transaction 0 with map(0), transaction 1 with map(0, 1), transaction 2, with map(0, 1, 2).

yuunlimm: it's for the batch of transactions

larry-aptos: yea makes sense

PR #460, opened 2024-07-13
[Processor] Use multiple threads to speed up FA processor.

grao1991:
Most part of this PR is just from auto-formatter.

This PR is trying to use rayon to process transactions in multiple threads. I moved all the hashmaps inside the loop (so assumption here is everything txn can be processed independently), and produces per txn output. The output is merged together at the very end.

The performance improvement depends on the machine spec. On my dev machine processing p2p transfer data is 10x+ faster.

### DIFF
--- a/rust/processor/src/processors/fungible_asset_processor.rs
+++ b/rust/processor/src/processors/fungible_asset_processor.rs
@@ -40,6 +40,7 @@ use diesel::{
     query_builder::QueryFragment,
     ExpressionMethods,
 };
+use rayon::prelude::*;
 use std::fmt::Debug;
 use tracing::error;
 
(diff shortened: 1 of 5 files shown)


rust/processor/src/processors/fungible_asset_processor.rs
```diff
@@ -453,294 +454,318 @@ async fn parse_v2_coin(
...
+            )
+        })
+        .collect();
+
+    for (faa, fab, acs, cfab, fam) in data {
+        fungible_asset_activities.extend(faa);
+        fungible_asset_balances.extend(fab);
+        all_coin_supply.extend(acs);
+        current_fungible_asset_balances.extend(cfab);
+        fungible_asset_metadata.extend(fam);
```
rtso, line 768: Is this necessary?

grao1991: I don't want to change the insert_to_db interface for now.

rtso (approved):

PR #461, opened 2024-07-15
add parquet txn metadata processor that handles write set size info

yuunlimm:
### Description
Add write_set_size_info table to parquet

### Test Plan
Screenshot 2024-07-15 at 2 10 54 PM

### DIFF
--- a/rust/processor/src/db/common/models/transaction_metadata_model/parquet_write_set_size_info.rs
+++ b/rust/processor/src/db/common/models/transaction_metadata_model/parquet_write_set_size_info.rs
@@ -1,0 +1,54 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(clippy::extra_unused_lifetimes)]
+
+use crate::bq_analytics::generic_parquet_processor::{GetTimeStamp, HasVersion, NamedTable};
+use allocative_derive::Allocative;
+use aptos_protos::transaction::v1::WriteOpSizeInfo;
+use field_count::FieldCount;
+use parquet_derive::ParquetRecordWriter;
+use serde::{Deserialize, Serialize};
+
+#[derive(
+    Allocative, Clone, Debug, Default, Deserialize, FieldCount, ParquetRecordWriter, Serialize,
+)]
+pub struct WriteSetSize {
+    pub txn_version: i64,
+    pub change_index: i64,
+    pub key_bytes: i64,
+    pub value_bytes: i64,
+    pub total_bytes: i64,
+}
+
+impl NamedTable for WriteSetSize {
+    const TABLE_NAME: &'static str = "write_set_size";
+}
+
+impl HasVersion for WriteSetSize {
+    fn version(&self) -> i64 {
+        self.txn_version
+    }
+}
+
+impl GetTimeStamp for WriteSetSize {
+    fn get_timestamp(&self) -> chrono::NaiveDateTime {
+        chrono::NaiveDateTime::default()
+    }
+}
+
+impl WriteSetSize {
+    pub fn from_transaction_info(
+        info: &WriteOpSizeInfo,
+        txn_version: i64,
+        change_index: i64,
+    ) -> Self {
+        WriteSetSize {
+            txn_version,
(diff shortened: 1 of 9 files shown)


rust/processor/src/processors/parquet_processors/parquet_transaction_metadata_processor.rs
```diff
@@ -0,0 +1,147 @@
...
+            .send(write_set_size_info_parquet_data)
+            .await
+            .context("Error sending write set size info to parquet handler")?;
+
+        Ok(ProcessingResult::ParquetProcessingResult(
+            ParquetProcessingResult {
+                start_version: start_version as i64,
+                end_version: end_version as i64,
+                last_transaction_timestamp: last_transaction_timestamp.clone(),
+                txn_version_to_struct_count: AHashMap::new(),
```
rtso, line 139: Do we need to give it `transaction_version_to_struct_count`?

yuunlimm: here we don't have to. I think this is a flaw that we would like to improve. b/c transaction_version_to_struct_count is only being used when the structs are processed (insert into buffer) in the parquet handler.

rust/processor/src/db/common/models/transaction_metadata_model/parquet_write_set_size_info.rs
```diff
@@ -0,0 +1,54 @@
...
+use aptos_protos::transaction::v1::WriteOpSizeInfo;
+use field_count::FieldCount;
+use parquet_derive::ParquetRecordWriter;
+use serde::{Deserialize, Serialize};
+
+#[derive(
+    Allocative, Clone, Debug, Default, Deserialize, FieldCount, ParquetRecordWriter, Serialize,
+)]
+pub struct WriteSetSize {
+    pub txn_version: i64,
```
yuunlimm, line 17: @ying-w let me know if it's missing any fields

ying-w: i thought we had talked about not using change_index but rather state_key_hash

yuunlimm: sync offline. this is fine to stay as it is

rust/processor/src/processors/parquet_processors/parquet_transaction_metadata_processor.rs
```diff
@@ -0,0 +1,150 @@
...
+                },
+            };
+            for (index, write_set_size_info) in size_info.write_op_size_info.iter().enumerate() {
+                write_set_sizes.push(WriteSetSize::from_transaction_info(
+                    write_set_size_info,
+                    txn_version,
+                    index as i64,
+                    block_timestamp,
+                ));
+                transaction_version_to_struct_count
```
larry-aptos, line 119: what will happen if `txn.size_info.as_ref()` is None? dose gap detector can handle this? if not, we should upgrade to panic or something

yuunlimm: if that is None, wouldn't that proceed to next txn?

larry-aptos: yea i was thinking there were some consecutive constrain required by Gap detector and now gap detector would look like [v1, v2, v4, v5..], thus the question

yuunlimm: no, the current behavior is, if there is no structs to process (e.g. the v3 will have 0 count),  it will proceed to next one to process. b/c we loop through starting_version ..endIng_version to populate the map first aside from the result we handle here.

larry-aptos (approved):

yuunlimm:
Screenshot 2024-07-24 at 5 27 17 PM
Screenshot 2024-07-24 at 5 29 00 PM
there is no write_set_size_info for version 9999999, and if you look at processor status for parquet_transaction_metadat_processor. it still updated

PR #468, opened 2024-07-18
fix parquet to start from where it left off when it's re-deployed

yuunlimm:
## Context 

There was a gap issue within the analytics pipeline. The new deployment restarted parquet processors, and it started from version stored in the `processor_status` alloydb table. however it wasn't working correctly b/c the table gets updated whenever the new parquet file gets uploaded to GCS, and the problem is that each table has a different frequency of uploads.  For example, for `move_resources`, because its size is bigger than any other structs, it gets uplaoded very frequency whereas move_modules reaches its max buffer size once every 30mins. and each file includes different rage of txn versions so that `processor status` is flaky. the version can go back and forth.  

Example failing scenario:
- last parquet file for `transactions` table hold a version 0-100000 -> file 1
- last parquet file for `move_resources` table hold a version 120000-150000. - file 2

1. file 1 uploaded at 12:00
2. processor-status now has a last_success_version: 100000
3. file 2 uplaoaded at 12
4. processor-status now has a last_success_version: 150000
5. new deployment kicks in
6. it reads the  last_success_version from db, which is 150000 and starts processing from 150000

ends up having a gap of 100001 - 149999.

### Previous behavior
parquet processors
- sends parquet data along with transaction_version_to_struct_count to channel that each table handler polls data from. this is where it clones maps N times (N = num of tables processor writes to)
- returns processing result, which wasn't handled anywhere 

parquet_gap_detector
- adde

generic parquet processor

### New behavior
parquet processors
- sends parquet data along with transaction_version_to_struct_count to channel that each table handler polls data from. this is where it clones maps N times (N = num of tables processor writes to)
- returns processing result with transaction_version_to_struct_count. and gap_detector in worker.rs updates its own map which is shared by all parquet handles per table 

parquet_gap_detector
- added a new function to update its version_counters, which is used to handle the processing_result returned from the parquet processor after sending all data over to channels.
- refactored out a function `update_next_version_to_process` which is responsible for updating `next_version_to_process`

generic parquet processor

### other changes
- removed tables that are not in scope like current)table_items from parquet_default_prcoessor
- added `parquet_processed_structs` and `table_name` to `ParquetProcessingResult` 
- added metrics. 

### Test Plan
GCS bucket used for devnet: https://console.cloud.google.com/storage/browser/aptos-indexer-data-etl-parquet-devnet/devnet-airflow-backfill?project=aptos-data-staging&supportedpurview=project&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))
load test resulting doc: https://www.notion.so/aptoslabs/Parquet-Load-Test-Plan-ac4e090c1210483a98d49cfdb65a26bf?pvs=4
Dashboard: https://aptoslabs.grafana.net/d/ads9cvy5s47pcc/indexer-parquet-processors-debug?from=now-30m&to=now&timezone=America%2FLos_Angeles&var-chain_name=devnet&var-processor_name=$__all&var-job=$__all&orgId=1&refresh=auto

- verified locally that `processor_status` now holds the min(max(txn_version) for `transactions`, max(txn_version) for `move_resources`, max(txn_version) for `write_set_changes`...) 
- verified that when we restart the processor with starting_version not specified, it picks up the version above. 

Screenshot 2024-07-18 a
![Screenshot 2024-07-18 at 1 45 36 PM
t 1 45 27 PM](https://github.com/user-attachments/assets/0c19c02c-ddc5-450d-985e-052c1ef6139c)

### DIFF
--- a/rust/processor/src/bq_analytics/gcs_handler.rs
+++ b/rust/processor/src/bq_analytics/gcs_handler.rs
@@ -65,7 +65,10 @@ pub async fn upload_parquet_to_gcs(
 
         match upload_result {
             Ok(Ok(result)) => {
-                info!("File uploaded successfully to GCS: {}", result.name);
+                info!(
+                    table_name = table_name,
+                    "File uploaded successfully to GCS: {}", result.name
+                );
                 return Ok(());
             },
             Ok(Err(e)) => {
--- a/rust/processor/src/bq_analytics/gcs_handler.rs
+++ b/rust/processor/src/bq_analytics/gcs_handler.rs
@@ -67,7 +67,9 @@ pub async fn upload_parquet_to_gcs(
             Ok(Ok(result)) => {
                 info!(
                     table_name = table_name,
-                    "File uploaded successfully to GCS: {}", result.name
+                    uploaded_time_iso = now.to_rfc3339(),
+                    "File uploaded successfully to GCS: {}",
+                    result.name
                 );
                 return Ok(());
             },
--- a/rust/processor/src/bq_analytics/gcs_handler.rs
+++ b/rust/processor/src/bq_analytics/gcs_handler.rs
@@ -1,7 +1,4 @@
-use crate::{
-    bq_analytics::ParquetProcessorError,
-    utils::counters::{PARQUET_BUFFER_SIZE, PARQUET_BUFFER_SIZE_AFTER_UPLOAD},
-};
+use crate::{bq_analytics::ParquetProcessorError, utils::counters::PARQUET_BUFFER_SIZE};
 use anyhow::{Context, Result};
 use chrono::{Datelike, Timelike};
 use google_cloud_storage::{
@@ -18,7 +15,7 @@ const INITIAL_DELAY_MS: u64 = 500;
 const TIMEOUT_SECONDS: u64 = 300;
 pub async fn upload_parquet_to_gcs(
     client: &GCSClient,
-    mut buffer: Vec<u8>,
+    buffer: Vec<u8>,
     table_name: &str,
     bucket_name: &str,
     bucket_root: &Path,
@@ -81,10 +78,6 @@ pub async fn upload_parquet_to_gcs(
                     file_name = result.name,
                     "File uploaded successfully to GCS",
                 );
-                buffer.clear();
-                PARQUET_BUFFER_SIZE_AFTER_UPLOAD
-                    .with_label_values(&[table_name])
-                    .set(buffer.len() as i64);
                 return Ok(());
             },
             Ok(Err(e)) => {
--- a/rust/processor/src/bq_analytics/gcs_handler.rs
+++ b/rust/processor/src/bq_analytics/gcs_handler.rs
@@ -1,23 +1,28 @@
-use crate::bq_analytics::ParquetProcessorError;
-use anyhow::Result;
+use crate::{
+    bq_analytics::ParquetProcessorError,
+    utils::counters::{PARQUET_BUFFER_SIZE, PARQUET_BUFFER_SIZE_AFTER_UPLOAD},
+};
+use anyhow::{Context, Result};
 use chrono::{Datelike, Timelike};
 use google_cloud_storage::{
     client::Client as GCSClient,
     http::objects::upload::{Media, UploadObjectRequest, UploadType},
 };
-use hyper::Body;
+use hyper::{body::HttpBody, Body};
 use std::path::{Path, PathBuf};
 use tokio::time::{sleep, timeout, Duration};
 use tracing::{debug, error, info};
+
 const MAX_RETRIES: usize = 3;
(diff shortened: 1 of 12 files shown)


rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -140,25 +137,33 @@ where
...
+                );
                 if let Err(e) = self.upload_buffer(gcs_client).await {
                     error!("Failed to upload buffer: {}", e);
                     return Err(e);
                 }
                 self.last_upload_time = Instant::now();
             }
+        }
 
-            if self.last_upload_time.elapsed() >= self.upload_interval {
```
yuunlimm, line 151: this wasn't working for structs other than transactions when there aren't no structs to process for the batch we are processing. so it never triggers an upload even though it has past the interval duration provided in the config. so moving this out of the for loop and verified it's working

rust/processor/src/bq_analytics/gcs_handler.rs
```diff
@@ -65,7 +68,16 @@ pub async fn upload_parquet_to_gcs(
 
         match upload_result {
             Ok(Ok(result)) => {
-                info!("File uploaded successfully to GCS: {}", result.name);
+                info!(
+                    table_name = table_name,
+                    uploaded_time_iso = now.to_rfc3339(),
```
larry-aptos, line 73: the time is built-in; you can remove this.

rust/processor/src/bq_analytics/gcs_handler.rs
```diff
@@ -65,7 +68,16 @@ pub async fn upload_parquet_to_gcs(
...
+                    "File uploaded successfully to GCS: {}",
+                    result.name
```
larry-aptos, line 75: file_name = result.name,
"File uploaded successfully to GCS"

With this, you can search file name easily.

rust/processor/src/bq_analytics/generic_parquet_processor.rs
```diff
@@ -216,7 +227,7 @@ where
 
         upload_parquet_to_gcs(
             gcs_client,
-            upload_buffer,
+            &mut upload_buffer,
```
larry-aptos, line 230: I have one question, is it possible to consume the buffer here? 

my understanding is: build buffer -> upload buffer and clear the buffer -> reuse the buffer. 

i seems that you've already given out the ownership of this buffer(`let mut upload_buffer = old_writer.into_inner()`)
 and you can do this instead

```
upload_buffer
```

rust/processor/src/gap_detectors/parquet_gap_detector.rs
```diff
@@ -2,84 +2,163 @@
...
         }
     }
+
+    pub fn update_struct_map(
+        &mut self,
+        txn_version_to_struct_count: AHashMap<i64, i64>,
+        start_version: i64,
+        end_version: i64,
+    ) {
+        for version in start_version..=end_version {
```
larry-aptos, line 50: neat; TIL.

rust/processor/src/bq_analytics/gcs_handler.rs
yuunlimm: good call!

rust/processor/src/bq_analytics/gcs_handler.rs
```diff
@@ -1,23 +1,28 @@
...
 use tokio::time::{sleep, timeout, Duration};
 use tracing::{debug, error, info};
+
 const MAX_RETRIES: usize = 3;
 const INITIAL_DELAY_MS: u64 = 500;
 const TIMEOUT_SECONDS: u64 = 300;
 pub async fn upload_parquet_to_gcs(
     client: &GCSClient,
-    buffer: Vec<u8>,
+    mut buffer: Vec<u8>,
```
larry-aptos, line 21: if you remove `clear` is mut required still?

yuunlimm: no, mut was required for `clear`

rust/processor/src/bq_analytics/gcs_handler.rs
```diff
@@ -65,7 +76,15 @@ pub async fn upload_parquet_to_gcs(
 
         match upload_result {
             Ok(Ok(result)) => {
-                info!("File uploaded successfully to GCS: {}", result.name);
+                info!(
+                    table_name = table_name,
+                    file_name = result.name,
+                    "File uploaded successfully to GCS",
+                );
+                buffer.clear();
```
larry-aptos, line 84: nit: remove clear

larry-aptos (approved):

yuunlimm:
Screenshot 2024-07-23 at 4 16 00 PM
Screenshot 2024-07-23 at 5 16 59 PM
Screenshot 2024-07-23 at 5 16 57 PM

PR #469, opened 2024-07-18
Indexer processor metrics improvements. 

sitalkedia:
- Fix some latency counters for indexer processor
- Add batch size metrics

### DIFF
--- a/rust/processor/src/utils/counters.rs
+++ b/rust/processor/src/utils/counters.rs
@@ -147,15 +147,6 @@ pub static PROCESSED_BYTES_COUNT: Lazy<IntCounterVec> = Lazy::new(|| {
     .unwrap()
 });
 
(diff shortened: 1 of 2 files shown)


grao1991 (approved):

rust/processor/src/utils/counters.rs
```diff
@@ -147,6 +147,15 @@ pub static PROCESSED_BYTES_COUNT: Lazy<IntCounterVec> = Lazy::new(|| {
     .unwrap()
 });
 
+pub static BATCH_SIZE: Lazy<HistogramVec> = Lazy::new(|| {
```
rtso, line 150: we already have a metric like this. it's `NUM_TRANSACTIONS_PROCESSED_COUNT`

sitalkedia: There is a slight semantic difference because this is `batch_end-batch_start` but for all practical purpose, this is good. Removed the counter for now.

PR #470, opened 2024-07-19
[EI-444] Migrate events processor to sdk 

rtso:
## Description
Migrate the events processor to using aptos-indexer-processor-sdk. The plan is to deploy it and monitor performance. 

## Testing 
Run events processor 
Screenshot 2024-07-31 at 5 39 00 PM


rust/.cargo/config.toml
```diff
@@ -15,3 +15,6 @@ xclippy = [
 
 [build]
 rustflags = ["--cfg", "tokio_unstable"]
+
+[net]
+git-fetch-with-cli = true
```
rtso, line 20: Is this safe to merge? @just-in-chang

just-in-chang: Not sure, don't know how CI does github authentication. Worried that it would break the docker build

rtso: ok maybe we just open source the sdk repo 🤷‍♀️

just-in-chang: We can see if it fails first

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -0,0 +1,102 @@
...
+        .connect_to(events_storer.into_runnable_step(), 10)
+        .connect_to(version_tracker.into_runnable_step(), 10)
+        .end_and_return_output_receiver(10);
+
+        // (Optional) Parse the results
+        loop {
+            match buffer_receiver.recv().await {
+                Ok(txn_context) => {
+                    if txn_context.data.is_empty() {
+                        println!("Received no transactions");
```
grao1991, line 88: nit: use logger?

rust/sdk-processor/src/steps/events_processor/events_extractor.rs
```diff
@@ -0,0 +1,76 @@
...
+    type Output = EventModel;
+    type RunType = AsyncRunType;
+
+    async fn process(
+        &mut self,
+        item: TransactionContext<Transaction>,
+    ) -> Option<TransactionContext<EventModel>> {
+        let events = item
+            .data
+            .par_iter()
```
grao1991, line 27: nit: maybe add with_min_len(64)

rust/sdk-processor/src/steps/events_processor/events_storer.rs
```diff
@@ -0,0 +1,88 @@
...
+        let execute_res = execute_in_chunks(
+            self.conn_pool.clone(),
+            insert_events_query,
+            &events.data,
+            get_config_table_chunk_size::<EventModel>("events", &per_table_chunk_sizes),
+        )
+        .await;
+        match execute_res {
+            Ok(_) => {
+                println!("Events stored successfully");
```
grao1991, line 72: nit: use logger?

rust/sdk-processor/src/steps/events_processor/events_storer.rs
```diff
@@ -0,0 +1,88 @@
...
+impl Processable for EventsStorer {
+    type Input = EventModel;
+    type Output = EventModel;
+    type RunType = AsyncRunType;
+
+    async fn process(
+        &mut self,
+        events: TransactionContext<EventModel>,
+    ) -> Option<TransactionContext<EventModel>> {
+        let per_table_chunk_sizes: AHashMap<String, usize> = AHashMap::new();
```
grao1991, line 62: the chunk size should be part of the event processor config.

rust/sdk-processor/src/utils/database.rs
```diff
@@ -0,0 +1,329 @@
...
+        res?
+    }
+
+    Ok(())
+}
+
+pub async fn execute_with_better_error<U>(
+    pool: ArcDbPool,
+    query: U,
+    mut additional_where_clause: Option<&'static str>,
```
grao1991, line 163: let's get rid of this, just like what @CapCap is doing in https://github.com/aptos-labs/aptos-indexer-processors/pull/288?

rust/sdk-processor/src/utils/database.rs
```diff
@@ -0,0 +1,329 @@
...
+    let debug_string = diesel::debug_query::<Backend, _>(&final_query).to_string();
+    tracing::debug!("Executing query: {:?}", debug_string);
+    let res = final_query.execute(conn).await;
+    if let Err(ref e) = res {
+        tracing::warn!("Error running query: {:?}\n{:?}", e, debug_string);
+    }
+    res
+}
+
+async fn execute_or_retry_cleaned<U, T>(
```
grao1991, line 234: we should rethink about the retry strategy here.

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -0,0 +1,51 @@
...
+) -> Result<Option<u64>> {
+    let mut conn = conn_pool.get().await?;
+
+    match ProcessorStatusQuery::get_by_processor(
+        indexer_processor_config.processor_config.name(),
+        &mut conn,
+    )
+    .await?
+    {
+        Some(status) => Ok(Some(status.last_success_version as u64 + 1)),
```
grao1991, line 48: why +1 here?

rust/sdk-processor/src/config/indexer_processor_config.rs
```diff
@@ -0,0 +1,58 @@
...
+use sdk_server_framework::RunnableConfig;
+use serde::{Deserialize, Serialize};
+
+pub const QUERY_DEFAULT_RETRIES: u32 = 5;
+pub const QUERY_DEFAULT_RETRY_DELAY_MS: u64 = 500;
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct IndexerProcessorConfig {
+    pub processor_config: ProcessorConfig,
```
grao1991, line 17: not necessary in this PR, but I think we should consider supporting running multiple processors here.

rust/sdk-processor/src/config/indexer_processor_config.rs
```diff
@@ -0,0 +1,58 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use super::processor_config::ProcessorConfig;
+use crate::processors::events_processor::EventsProcessor;
+use anyhow::Result;
+use aptos_indexer_processor_sdk::aptos_indexer_transaction_stream::TransactionStreamConfig;
+use sdk_server_framework::RunnableConfig;
+use serde::{Deserialize, Serialize};
+
+pub const QUERY_DEFAULT_RETRIES: u32 = 5;
```
grao1991, line 11: are these constants used?

rust/sdk-processor/src/config/indexer_processor_config.rs
```diff
@@ -0,0 +1,58 @@
...
+            .next()
+            .unwrap_or("unknown");
+        before_underscore[..before_underscore.len().min(12)].to_string()
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct DbConfig {
+    pub postgres_connection_string: String,
```
grao1991, line 48: nit: just call it `connection_string`? we should make sdk to be orthogonal to the database choice when possible.

rust/sdk-processor/src/config/indexer_processor_config.rs
```diff
@@ -0,0 +1,58 @@
...
+
+pub const QUERY_DEFAULT_RETRIES: u32 = 5;
+pub const QUERY_DEFAULT_RETRY_DELAY_MS: u64 = 500;
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct IndexerProcessorConfig {
+    pub processor_config: ProcessorConfig,
+    pub transaction_stream_config: TransactionStreamConfig,
+    pub db_config: DbConfig,
```
grao1991, line 19: should this even be a top level config? theoretically the processor might not depend on the database, only the progress tracker. so probably make it a ProgressTrackerConfig?

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        Ok(Self {
+            config,
+            db_pool: conn_pool,
+        })
+    }
+
+    pub async fn run_processor(self) -> Result<()> {
+        // (Optional) Run migrations
+        run_migrations(
```
grao1991, line 42: IMO schema should be managed outside of processor.

rust/.cargo/config.toml
rtso: the build isn't even running lol

rust/sdk-processor/src/config/indexer_processor_config.rs
rtso: Can you explain what you mean and why? If you want to extract different types of data using the same TransactionStream, you could use the fanout method in ProcessorBuilder. But I think we shouldn't support multiple processors in the config, that fundamentally changes how we do deployment

rust/sdk-processor/src/config/indexer_processor_config.rs
rtso: This is only used by our processors, hence why it's not in the aptos-indexer-processor-sdk crate. This IndexerProcessorConfig only serves as an example and you can add/remove fields as needed.

rust/sdk-processor/src/processors/events_processor.rs
rtso: Do you have a proposal on how deployments would work in that case?

rust/sdk-processor/src/config/indexer_processor_config.rs
grao1991: What I'm saying is we should provide the `fanout to multiple processors after receiving the grpc stream` functionality as part of the sdk/library. We should make it available at config level so it's much easier to use, otherwise if it requires code change, it will be less likely for people to adopt to it.

rtso: Got it. This sounds like a separate feature built on top of SDK which will require SDK + config + deployment changes, almost like config as code which @banool  and @just-in-chang have sort of been looking into for the low code indexing.

grao1991: and for our own deployment, if you look at the profiling result, the stream part is taking a significant portion of CPU for most of the processors, and it takes a lot of network bandwidth, too.

grao1991: > This sounds like a separate feature built on top of SDK which will require SDK + config + deployment changes

I think it should be on top of `aptos_indexer_processor_sdk`, but as part of `rust/sdk-processor`.

rtso: > but as part of rust/sdk-processor

I don't necessarily agree with this. This has implications on processor performance, like if 1 processor is behind or stuck all of them will be behind.

rust/sdk-processor/src/config/indexer_processor_config.rs
grao1991: should we expect most of external people still reuse whatever we provide, they may add processors, but whatever common/framework code we provide here will likely be reused.

even for us, from an interface design point of view, this should look like:

```
struct IndexerProcessorConfig {
  progress_tracker_config: ProgressTrackerConfig,
  processor_config: ...
  ...
}

struct ProgressTrackerConfig {
  db_config: PostgresDbConfig,  // we should probably even make it an enum here to allow using things other than pg here.
}

struct EventsProcessorConfig {
  event_store_config: EventsStoreConfig,
  ...
}

struct EventsStoreConfig {
    db_config: PostgresDbConfig,  // we should probably even make it an enum here to allow using things other than pg here.
}

struct ParquetDefaultProcessorConfig {
  // no db_config here
}

```

rtso: I think we should keep things simple, without creating so many configs that all contain the same thing which makes deployment and code more complicated: 

```
struct IndexerProcessorConfig {
  db_config: Option<DbConfig>, // could be enum
  parquet_config: Option<ParquetConfig>,
  processor_config: ...
  ...
}
```

rust/sdk-processor/src/utils/database.rs
rtso: I'll track this in a task for now

rust/sdk-processor/src/config/indexer_processor_config.rs
rtso: I like your idea of making it an enum to support different DBs

rust/sdk-processor/src/utils/database.rs
rtso: Added as a task

rust/sdk-processor/src/utils/starting_version.rs
rtso: // Add +1 to start from the version after the last successful version

rust/sdk-processor/src/config/indexer_processor_config.rs
rtso: They will be used once we migrate more processors, I can remove them for now

rust/sdk-processor/src/config/indexer_processor_config.rs
grao1991: I understand there will be some implications, what I'm saying is we should provide the options, and allow the user to choose whichever they think better. There were people explicitly asking for this, and I do see the pros would outweigh the cons in many cases.

rust/sdk-processor/src/processors/events_processor.rs
grao1991: Just have another binary/script/github workflow/etc.. Schema changes are in general considered very risky, they should be explicitly acknowledged and carefully monitored, rather than shipping with binary implicitly.

rust/sdk-processor/src/utils/starting_version.rs
grao1991: If that's the case, I think you probably want to rename the method name to be something like `get_starting_version_from_db`, and also consider return 0 instead of None when no progress is in the database. Or instead you can choose keep the current name, and handle `+1` at the caller side.
It looks a little bit weird to me that you are doing `+1` here but doing the None -> 0 outside. IMO they should be together.

rust/sdk-processor/src/config/indexer_processor_config.rs
rtso: It can be built with the components offered by SDK. We can provide an example of sharing 1 stream with multiple "processors" once we work on the documentation.

rust/sdk-processor/src/processors/events_processor.rs
rtso: Good points, I'll create a future task to track this since it will require more testing, deployment changes and communication to partners who are running the processors.

Let's limit the scope of this PR as only for migrating existing processor to SDK.

rust/sdk-processor/src/config/db_config.rs
```diff
@@ -0,0 +1,54 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct PostgresConfig {
+    pub connection_string: String,
+    // Size of the pool for writes/reads to the DB. Limits maximum number of queries in flight
+    #[serde(default = "PostgresConfig::default_db_pool_size")]
+    pub db_pool_size: u32,
+    // Number of rows to insert, per chunk, for each DB table. Default per table is ~32,768 (2**16/2)
+    #[serde(default = "AHashMap::new")]
+    pub per_table_chunk_sizes: AHashMap<String, usize>,
```
grao1991, line 47: I don't think this should be part of PostgresConfig. It should probably be at each processor level. Although we are using a naive batching strategy based on the # of rows right now, in theory it should be smarter and different processors could choose to do it in a very different way. (e.g. maybe I want to based on the size, maybe I want to chunk the rows in a unordered way to better align with my db schema, etc.)

Also the comment for the default value is out of date.

rust/sdk-processor/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+
+#[cfg(unix)]
+#[global_allocator]
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
+
+const RUNTIME_WORKER_MULTIPLIER: usize = 2;
+
+fn main() -> Result<()> {
+    let num_cpus = num_cpus::get();
+    let worker_threads = (num_cpus * RUNTIME_WORKER_MULTIPLIER).max(16);
```
grao1991, line 14: I understand this is copied from the current code, but I don't see any reason to do RUNTIME_WORKER_MULTIPLIER, or max(16) here.

rust/sdk-processor/src/steps/events_processor/events_extractor.rs
```diff
@@ -0,0 +1,81 @@
...
+            .with_min_len(MIN_TRANSACTIONS_PER_RAYON_JOB)
+            .map(|txn| {
+                let mut events = vec![];
+                let txn_version = txn.version as i64;
+                let block_height = txn.block_height as i64;
+                let txn_data = match txn.txn_data.as_ref() {
+                    Some(data) => data,
+                    None => {
+                        warn!(
```
grao1991, line 40: When do we expect this to happen? Should this be error! or even panic!?

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -0,0 +1,120 @@
...
+                .map_err(|e| {
+                    anyhow::anyhow!(
+                        "Failed to create connection pool for PostgresConfig: {:?}",
+                        e
+                    )
+                })?;
+
+                Ok(Self {
+                    config,
+                    postgres_config: postgres_db_config.clone(),
```
grao1991, line 51: do you need clone here?

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -0,0 +1,120 @@
...
+
+pub struct EventsProcessor {
+    pub config: IndexerProcessorConfig,
+    pub postgres_config: PostgresConfig,
+    pub db_pool: ArcDbPool,
+}
+
+impl EventsProcessor {
+    pub async fn new(config: IndexerProcessorConfig) -> Result<Self> {
+        let db_config = config.db_config.clone();
```
grao1991, line 34: do you really need clone here?

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -0,0 +1,120 @@
...
+            ..self.config.transaction_stream_config
+        })
+        .await?;
+        let transaction_stream_with_input = RunnableStepWithInputReceiver::new(
+            input_receiver,
+            transaction_stream.into_runnable_step(),
+        );
+        let events_extractor = EventsExtractor {};
+        let events_storer = EventsStorer::new(self.db_pool.clone(), self.postgres_config.clone());
+        let timed_buffer = TimedBuffer::new(Duration::from_secs(1));
```
grao1991, line 83: why do we need a 1 second timed buffer here?

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -0,0 +1,120 @@
...
+        .connect_to(timed_buffer.into_runnable_step(), 10)
+        .connect_to(events_storer.into_runnable_step(), 10)
+        .connect_to(version_tracker.into_runnable_step(), 10)
+        .end_and_return_output_receiver(10);
+
+        // (Optional) Parse the results
+        loop {
+            match buffer_receiver.recv().await {
+                Ok(txn_context) => {
+                    if txn_context.data.is_empty() {
```
grao1991, line 104: why do we even allow empty data? any chance we make it a warn! or error!?

rust/sdk-processor/src/steps/events_processor/events_storer.rs
```diff
@@ -0,0 +1,102 @@
...
+            self.conn_pool.clone(),
+            insert_events_query,
+            &events.data,
+            get_config_table_chunk_size::<EventModel>("events", &per_table_chunk_sizes),
+        )
+        .await;
+        match execute_res {
+            Ok(_) => {
+                debug!(
+                    "Events versions {} to {} stored successfully",
```
grao1991, line 81: For logging, I generally recommend to log range as  `[start_version, end_version]` to make it more clear the end_version is inclusive.

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -0,0 +1,120 @@
...
+        );
+
+        // Connect processor steps together
+        let (_, buffer_receiver) = ProcessorBuilder::new_with_runnable_input_receiver_first_step(
+            transaction_stream_with_input,
+        )
+        .connect_to(events_extractor.into_runnable_step(), 10)
+        .connect_to(timed_buffer.into_runnable_step(), 10)
+        .connect_to(events_storer.into_runnable_step(), 10)
+        .connect_to(version_tracker.into_runnable_step(), 10)
```
grao1991, line 97: Is there any step that process txns out of order? If not can we have the version tracker to enforce this?

rust/sdk-processor/src/main.rs
rtso: What do you suggest?

rust/sdk-processor/src/steps/events_processor/events_extractor.rs
rtso: This happens when processing block epilogue transactions, it doesn't have transaction data.

rust/sdk-processor/src/processors/events_processor.rs
rtso: We actually don't need it

rust/sdk-processor/src/processors/events_processor.rs
rtso: We don't have this now, but max was saying there's a way to create worker tasks that each pull from the same channel and run the same step.

rust/sdk-processor/src/processors/events_processor.rs
rtso: We don't process events for certain transaction types like block epilogue transactions. For transactions with no extracted data, the data would be empty and that is fine. But it still needs to be passed through the pipeline to track the processed versions.

rust/sdk-processor/src/main.rs
grao1991: I think it's ok to use just `num_cpus`.

grao1991 (approved):

PR #473, opened 2024-07-22
add missing index causing query to timeout

yuunlimm:
### Description

There were sevs for missing data in the analytics pipeline, and it was due to that query was timing out. The query failing due to that indices were missing. 

### Test Plan
We are running this outside of diesel migration to avoid locks.


rtso:
I don't think this needs a migration, as @just-in-chang found that there was never a migration that dropped the indices in the first place. The SQL LGTM!

just-in-chang:
File names have a typo 😆

yuunlimm:
will close this PR! will run the scrip locally

PR #475, opened 2024-07-29
add missing indexing delete reousrces

yuunlimm:
### Description 
parquet token_v2_ownerships table was missing data. turns out it wasn't indexing some of changes like transfer and delete.

Note that we have some stale data in our DB where we have rows for self transfer, but we shouldn't index self transfer with events index. 

### Test Plan
tested it not indexes the data that we were missing previously.
Screenshot 2024-07-29 at 2 03 33 PM


ying-w (approved):

ying-w:
we're going to be missing addresses on some burn events

PR #485, opened 2024-08-16
Fix current_collection_ownership_v2_view

rtso:
## Description 
Fix collections view to use current tables in the join 

## Testing
Screenshot 2024-08-16 at 2 36 55 PM

Query is fixed in Petra.

### DIFF
--- a/rust/processor/src/db/postgres/migrations/2024-08-16-213505_fix_collections_view/up.sql
+++ b/rust/processor/src/db/postgres/migrations/2024-08-16-213505_fix_collections_view/up.sql
@@ -1,0 +1,18 @@
+-- Your SQL goes here
+CREATE OR REPLACE VIEW current_collection_ownership_v2_view as
+select owner_address,
+  creator_address,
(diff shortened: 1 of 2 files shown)


rust/processor/src/db/postgres/migrations/2024-08-16-213505_fix_collections_view/up.sql
```diff
@@ -0,0 +1,19 @@
+-- Your SQL goes here
+DROP VIEW IF EXISTS current_collection_ownership_v2_view;
```
yuunlimm, line 2: do we need this?

yuunlimm (approved):

rust/processor/src/db/postgres/migrations/2024-08-16-213505_fix_collections_view/up.sql
rtso: nope i'll remove

PR #487, opened 2024-08-20
support new signature type NoAccountSignature

larry-aptos:


larry-aptos:
abandoned; upstream won't include this change against proto.

PR #488, opened 2024-08-20
Yuunlimm/create the processor test case

yuunlimm:
## Description

originally from: https://github.com/aptos-labs/aptos-indexer-processors/pull/464

* The test framework for indexer processors. 
* Added test cases for Event processor which is to validate functionality of the event processor. The tests verify that specific types of events are correctly captured and processed based on predefined transactions. Here is a summary of each test case:

### 1. test_faucet_funded_event:  
- Purpose: To validate that the system correctly processes events when an account is funded from a faucet.
- Transactions:
  - ACCOUNT_A_GET_FUNDED_FROM_FAUCET - Simulates a scenario where Account A receives funds from a faucet.
- Assertions:
  - Ensures that the correct number of events is recorded.
  - Checks that only one WithdrawEvent is generated (withdrawal from the faucet).
  - Validates that two DepositEvent events are generated (one deposit into Account A).

### 2. test_transfer_event
- Purpose: To validate the processing of events when funds are transferred between accounts that are associated with multiple transactions.
- Transactions
  - ACCOUNT_A_GET_FUNDED_FROM_FAUCET - Account A is funded.
  - ACCOUNT_A_TRANSFER_TO_ACCOUNT_B - Account A transfers funds to Account B.
  - ACCOUNT_B_TRANSFER_TO_ACCOUNT_B - Account B transfers funds to itself.
- Assertions
  - checks that three `WithdrawEvent` evnets are generated (one from the faucet, one from Account A, and one from AccountB) 

### 3. test_validator_txn_events
- Purpose: To verify the event processing for a validator transaction.
- Transactions:
  - VALIDATOR_TRANSACTION - Simulates a transaction involving a validator.
- Assertions:
  - Confirms that 20 events are generated.
  - Validates that 19 `DistributeRewardsEvent` events are processed.

### 4. test_generated_user_script_events
- Purpose: To check the processing of events generated by a user script transaction.
- Transactions:
  - GENERATED_USER_SCRIPT_TRANSACTION - Simulates a transaction created by a user-defined script.
- Assertions:
  - Ensures that only one event is generated.
  - Validates that the event is of type FeeStatement.

## Usage
Currently using a limited set of `hard-coded` transactions from indexer-trasnaction-tests under aptos-core. dependency: https://github.com/aptos-labs/aptos-core/pull/14347

* To use the test transactions, just `use aptos_indexer_test_transactions::ACCOUNT_A_GET_FUNDED_FROM_FAUCET`.
* To test against these transaction, just

```
        let test_context = TestContext::new(&[
            ACCOUNT_A_GET_FUNDED_FROM_FAUCET
        ])
        .await
        .unwrap();
        let processor_config = TestProcessorConfig {
            config: processor::processors::ProcessorConfig::CoinProcessor,
        };
        assert!(test_context
            .run(processor_config, |conn: &mut PgConnection| {
            // put your expected result here.
           }
        );
```

### Test Plan
ran `cargo test` 
Screenshot 2024-08-20 at 3 36 01 PM

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,0 +1,28 @@
+name: Run Integration Tests
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  Integration-tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set branch
+        run: echo "BRANCH_NAME=${GITHUB_REF##*/}" >> $GITHUB_ENV
+
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Install Dependencies and Run Linter
+        uses: ./.github/actions/dep_install_and_lint
+        with:
+          working-directory: rust
+
+      - name: Run Integration Tests
+        run: cargo test --manifest-path integration-tests/Cargo.toml
+        working-directory: rust
--- a/rust/integration-tests/src/event_processor/event_processor_tests.rs
+++ b/rust/integration-tests/src/event_processor/event_processor_tests.rs
@@ -1,0 +1,255 @@
+#[cfg(test)]
+mod test {
+    use crate::{TestContext, TestProcessorConfig};
+    use aptos_indexer_test_transactions::{
+        ACCOUNT_A_GET_FUNDED_FROM_FAUCET, ACCOUNT_A_TRANSFER_TO_ACCOUNT_B,
+        ACCOUNT_B_TRANSFER_TO_ACCOUNT_B, GENERATED_USER_SCRIPT_TRANSACTION, VALIDATOR_TRANSACTION,
+    };
+    use diesel::{
+        pg::PgConnection,
+        query_dsl::methods::{FilterDsl, SelectDsl},
+        BoolExpressionMethods, ExpressionMethods, QueryResult, RunQueryDsl,
+    };
+    use processor::schema::events::dsl::*;
+
+    const WITHDRAW_EVENT: &str = "0x1::coin::WithdrawEvent";
+    const DEPOSIT_EVENT: &str = "0x1::coin::DepositEvent";
+    const FEE_STATEMENT_EVENT: &str = "0x1::transaction_fee::FeeStatement";
+    const DISTRIBUTE_REWARDS_EVENT: &str = "0x1::stake::DistributeRewardsEvent";
+
+    // Test Case: Validate the transaction that funds an account A from the faucet parsed into the correct events.
+    // - Verifies that the faucet funding results in one WithdrawEvent and two DepositEvents.
+    #[tokio::test]
+    async fn test_faucet_funding_transaction_with_correct_events() {
+        let test_context = TestContext::new(&[ACCOUNT_A_GET_FUNDED_FROM_FAUCET])
+            .await
+            .unwrap();
+        let processor_config = TestProcessorConfig {
+            config: processor::processors::ProcessorConfig::EventsProcessor,
+        };
+        let expected_transaction = test_context.transaction_batches[0].clone();
+
+        assert!(test_context
+            .run(processor_config, move |conn: &mut PgConnection| {
+                // Load and validate events
+                let withdraw_events = load_transaction_events(conn, 0, WITHDRAW_EVENT)
+                    .expect("Failed to load WithdrawEvent");
+                assert_eq!(withdraw_events.len(), 1);
+
+                validate_fee_statement_event(
+                    &withdraw_events[0],
+                    expected_transaction.version as i64,
+                    "0x832fd5e456b7f43c4ef27978766ee5242a8a393d10ef5bf662d40f774a6f2deb",
+                    3,
+                    WITHDRAW_EVENT,
+                );
+
+                let deposit_events = load_transaction_events(conn, 0, DEPOSIT_EVENT)
+                    .expect("Failed to load DepositEvent");
+                assert_eq!(deposit_events.len(), 2);
+
+                validate_fee_statement_event(
+                    &deposit_events[0],
+                    expected_transaction.version as i64,
+                    "0x832fd5e456b7f43c4ef27978766ee5242a8a393d10ef5bf662d40f774a6f2deb",
+                    2,
+                    DEPOSIT_EVENT,
+                );
+
+                Ok(())
+            })
+            .await
+            .is_ok());
+    }
+
+    // Test Case: Validate that a validator transaction generates the correct events.
+    // - Verifies that a validator transaction results in 20 events, with 19 being DistributeRewardsEvents.
+    #[tokio::test]
+    async fn test_validator_transaction_distribute_rewards_events() {
+        let test_context = TestContext::new(&[VALIDATOR_TRANSACTION]).await.unwrap();
+        let expected_transaction = test_context.transaction_batches[0].clone();
+
+        assert!(test_context
+            .run(
+                default_processor_config(),
+                move |conn: &mut PgConnection| {
+                    // Load and validate events
+                    let distributed_rewards_events =
+                        load_transaction_events(conn, 0, DISTRIBUTE_REWARDS_EVENT)
+                            .expect("Failed to load DistributeRewardsEvents");
+                    assert_eq!(distributed_rewards_events.len(), 19);
+
+                    validate_fee_statement_event(
+                        &distributed_rewards_events[0],
+                        expected_transaction.version as i64,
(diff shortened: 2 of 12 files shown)


rust/integration-tests/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+            .await
+            .unwrap();
+        format!("postgres://postgres:postgres@{host}:{port}/postgres")
+    }
+
+    // `run` functions takes a closure that is executed after the test context is created.
+    // The closure will be executed multiple times with different permutations of the transactions.
+    // For example:
+    //   test.run(async move | context | {
+    //       // Runs after every permutatation
```
larry-aptos, line 87: maybe update the document here a little since we're using diesel select function.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,30 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  Integration-tests:
+    runs-on: ubuntu-latest
+
+    steps:
```
larry-aptos, line 14: nit: this part is reused; create a shared step for better readabiliy

rust/integration-tests/src/event_processor/event_processor_tests.rs
```diff
@@ -0,0 +1,137 @@
...
+                    .load::<(i64, i64, String)>(conn);
+                assert_eq!(result.unwrap().len(), 5);
+
+                let withdraw_events = events
+                    .select((transaction_version, event_index, type_))
+                    .filter(type_.eq("0x1::coin::WithdrawEvent"))
+                    .load::<(i64, i64, String)>(conn);
+
+                // Withdraw from Faucet, Withdraw from AccountA, Withdraw from AccountB
+                assert_eq!(withdraw_events.unwrap().len(), 1);
```
larry-aptos, line 37: nit: update comment

rust/integration-tests/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+    core::{IntoContainerPort, WaitFor},
+    runners::AsyncRunner,
+    ContainerAsync, GenericImage, ImageExt,
+};
+
+mod event_processor;
+
+/// The test context struct holds the test name and the transaction batches.
+pub struct TestContext {
+    pub transaction_batches: Vec<Transaction>,
```
rtso, line 20: Instead of just a Vec<Transaction>, we could have it return Map<TransactionVersion, Transaction>. That way, the test can assert between the json txn and the DB rows on specific txn data.

rust/integration-tests/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+    where
+        F: Fn(&mut PgConnection) -> anyhow::Result<()> + Send + Sync + 'static,
+    {
+        let transactions = self.transaction_batches.clone();
+        let db_url = self.get_db_url().await;
+        let mut conn = PgConnection::establish(&db_url)
+            .with_context(|| format!("Error connecting to {}", db_url))?;
+        let db_pool = new_db_pool(&db_url, None).await.unwrap();
+
+        for perm in transactions.iter().permutations(transactions.len()) {
```
rtso, line 106: Could we have a comment here explaining why we're doing the permutations, ie to test that transactions processed out of order will result in the same output in DB.

rust/integration-tests/src/event_processor/event_processor_tests.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[tokio::test]
+    async fn test_faucet_funded_event() {
+        let test_context = TestContext::new(&[ACCOUNT_A_GET_FUNDED_FROM_FAUCET])
+            .await
+            .unwrap();
+        let processor_config = TestProcessorConfig {
+            config: processor::processors::ProcessorConfig::EventsProcessor,
+        };
+        assert!(test_context
+            .run(processor_config, |conn: &mut PgConnection| {
```
rtso, line 24: Might be also good to pass in a reference to the test transactions so they can be used inside of the test cases.

rust/integration-tests/src/lib.rs
yuunlimm: not quite following why we have to hardcode values in the test case.  could you elaborate more?

rtso: What I was suggesting is to provide the input test transactions to each test case so that you could have assertions like:
```
assert(txn.events.0.event_address, db.event_address)
```
But actually I think it might be better to hardcode the asserted values in the test case, because the proto schema and output could change

yuunlimm: Thanks for elaborating it! I see the value in hardcoding the asserted values to ensure stability, especially considering that the proto schema and output might change over time. However, I also think that providing some fields directly from the input transactions can be beneficial, as it allows us to verify that the processing logic correctly handles data anyways.

so to balance this out, i propose this:

1. For fields that are unlikely to change frequently or are critical, I'll hardcode the expected values in the test cases like indexed_type, address, creation_number.

2. For fields that are easily derived from the input transactions, I'll pull the expected values directly from the input transactions. This way, we can still validate that the processing logic correctly parses and stores these fields.

how does that sound to you?

rtso: Sounds good! I like giving the test cases the option of whether to hardcode vs derive from input txn's. Deriving from input txn's also makes the test cases more versatile. For example, the same test can be used to test transfers A -> B txn and also C -> D because the test case can derive the from/to address instead of hardcoding the transfer addresses in the test.

.github/workflows/dep_install_and_lint.yaml
```diff
@@ -0,0 +1,24 @@
+name: Install Dependencies and Run Linter
```
rtso, line 1: Seems duplicated from `.github/actions/dep_install_and_lint/action.yaml`, maybe we can use the action in this workflow?

yuunlimm: yes, this should be deleted!

rust/integration-tests/src/event_processor/event_processor_tests.rs
```diff
@@ -0,0 +1,255 @@
...
+                    .filter(|(_, _, _, _, _account_address)| _account_address.eq(account_a_address)) // Ensure correct creation number
+                    .for_each(|(_, _, _, _data, _)| {
+                        // Ensure that any key starting with "amount" has the expected value
+                        _data
+                            .as_object()
+                            .unwrap()
+                            .iter()
+                            .filter(|(k, _)| k.starts_with("amount"))
+                            .for_each(|(_, v)| {
+                                assert_eq!(v.as_str().unwrap(), "100000000");
```
larry-aptos, line 191: does this mean all deposit amount is 100000000?

rust/integration-tests/src/event_processor/event_processor_tests.rs
```diff
@@ -0,0 +1,255 @@
...
+                            });
+                    });
+
+                // 2 Deposit evnets from Faucet, 1 from AccountA, and 1 from AccountB
+                assert_eq!(deposit_events.len(), 4);
+
```
larry-aptos, line 197: do you mean 4 deposit events?

rust/processor/Cargo.lock
```diff
@@ -1,2382 +0,0 @@
-# This file is automatically @generated by Cargo.
```
larry-aptos, line 1: why the deletion btw?

rust/integration-tests/src/event_processor/event_processor_tests.rs
yuunlimm: for it's testing against specific event filtered by `account_a_address`. I am thinking of adding a feature of using input json to extract values so then we can easily compare all data

rust/processor/Cargo.lock
yuunlimm: oh this hsould'been added back. it was causing merge conflict.  will add back

rust/integration-tests/src/event_processor/event_processor_tests.rs
larry-aptos: probably you can add in comments like `checking deposit into account A is 100000000`

rust/integration-tests/src/event_processor/event_processor_tests.rs
yuunlimm: 2 from faccet, 2 from account a, and 2 from account b. will make it clearer !

larry-aptos: oh i see

larry-aptos (approved):

PR #490, opened 2024-08-23
use runs-on

yuunlimm:
### Description
we were hitting rate limit on docker hub, so we replace this with runs-on. 

also remove the env variable that we don't need.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -9,12 +9,9 @@ on:
 
 jobs:
   Integration-tests:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


larry-aptos (approved):

.github/workflows/integration-tests.yaml
```diff
@@ -9,12 +9,9 @@ on:
 
 jobs:
   Integration-tests:
-    runs-on: ubuntu-latest
+    runs-on: runs-on,cpu=2,image=cpu-linux-x64,run-id=${{ github.run_id }}
```
geekflyer, line 12:
```suggestion
    runs-on: runs-on,runner=2cpu-linux-x64,run-id=${{ github.run_id }}
```

PR #494, opened 2024-08-28
[2/2][FA migration] Complete fa migration and rename table

bowenyang007:
1. Swap Hasura to use `current_unified_fungible_asset_balances`
2. Add a killswitch on `current_unified_fungible_asset_balances_to_be_renamed` 
    1. In addition, make a migration dropping columns. We want to split dropping columns separately from adding columns back so that we can add the columns without going through diesel migration again. 
    2. Deploy the change with KS turned on. This should create no downtime since column dropping is quick.

**[This PR]**
3. Add generated columns back in dbeaver
4. Create new migration with added columns. Then remove KS. 
5. Rename `current_fungible_asset_balances` to `current_fungible_asset_balances_legacy` 
6. Rename `current_unified_fungible_asset_balances_to_be_renamed` to `current_fungible_asset_balances`

### DIFF
--- a/rust/processor/src/db/postgres/migrations/2024-08-28-015552_fa_migration_fix_2/up.sql
+++ b/rust/processor/src/db/postgres/migrations/2024-08-28-015552_fa_migration_fix_2/up.sql
@@ -1,0 +1,14 @@
+-- Your SQL goes here
+ALTER TABLE current_unified_fungible_asset_balances_to_be_renamed
+ADD COLUMN IF NOT EXISTS asset_type VARCHAR(1000) GENERATED ALWAYS AS (COALESCE(asset_type_v1, asset_type_v2)) STORED;
+ALTER TABLE current_unified_fungible_asset_balances_to_be_renamed
+ADD COLUMN IF NOT EXISTS token_standard VARCHAR(10) GENERATED ALWAYS AS (
+    CASE
+      WHEN asset_type_v1 IS NOT NULL THEN 'v1'
+      ELSE 'v2'
+    END
+  ) STORED;
+ALTER TABLE current_fungible_asset_balances
+  RENAME TO current_fungible_asset_balances_legacy;
+ALTER TABLE current_unified_fungible_asset_balances_to_be_renamed
+  RENAME TO current_fungible_asset_balances;
--- a/rust/processor/src/db/postgres/migrations/2024-08-28-015552_fa_migration_fix_2/up.sql
+++ b/rust/processor/src/db/postgres/migrations/2024-08-28-015552_fa_migration_fix_2/up.sql
@@ -1,9 +1,9 @@
 -- Your SQL goes here
 ALTER TABLE current_unified_fungible_asset_balances_to_be_renamed
-ADD COLUMN IF NOT EXISTS asset_type VARCHAR(1000) GENERATED ALWAYS AS (COALESCE(asset_type_v1, asset_type_v2)) STORED;
+ADD COLUMN IF NOT EXISTS asset_type VARCHAR(1000) NOT NULL GENERATED ALWAYS AS (COALESCE(asset_type_v1, asset_type_v2)) STORED;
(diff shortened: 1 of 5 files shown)


rust/processor/src/db/postgres/schema.rs
```diff
@@ -423,6 +423,35 @@ diesel::table! {
...
     current_fungible_asset_balances (storage_id) {
+        #[max_length = 66]
+        storage_id -> Varchar,
+        #[max_length = 66]
+        owner_address -> Varchar,
+        #[max_length = 66]
+        asset_type_v2 -> Nullable<Varchar>,
+        #[max_length = 1000]
+        asset_type_v1 -> Nullable<Varchar>,
+        is_primary -> Nullable<Bool>,
```
rtso, line 434: This shouldn't be nullable to match with the legacy schema, but I see some columns that are null is db

rust/processor/src/db/postgres/schema.rs
```diff
@@ -423,6 +423,35 @@ diesel::table! {
...
+        last_transaction_version_v2 -> Nullable<Int8>,
+        last_transaction_version -> Nullable<Int8>,
+        last_transaction_timestamp_v1 -> Nullable<Timestamp>,
+        last_transaction_timestamp_v2 -> Nullable<Timestamp>,
+        last_transaction_timestamp -> Nullable<Timestamp>,
+        inserted_at -> Timestamp,
+        #[max_length = 1000]
+        asset_type -> Nullable<Varchar>,
+        #[max_length = 10]
+        token_standard -> Nullable<Varchar>,
```
rtso, line 449: This shouldn't be nullable either

rust/processor/src/db/postgres/schema.rs
rtso: @bowenyang007

rust/processor/src/db/postgres/schema.rs
```diff
@@ -423,6 +423,35 @@ diesel::table! {
...
+        owner_address -> Varchar,
+        #[max_length = 66]
+        asset_type_v2 -> Nullable<Varchar>,
+        #[max_length = 1000]
+        asset_type_v1 -> Nullable<Varchar>,
+        is_primary -> Nullable<Bool>,
+        is_frozen -> Bool,
+        amount_v1 -> Nullable<Numeric>,
+        amount_v2 -> Nullable<Numeric>,
+        amount -> Nullable<Numeric>,
```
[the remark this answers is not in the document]
rtso: @bowenyang007

rust/processor/src/db/postgres/migrations/2024-08-28-015552_fa_migration_fix_2/up.sql
```diff
@@ -0,0 +1,15 @@
+-- Your SQL goes here
+ALTER TABLE current_unified_fungible_asset_balances_to_be_renamed
+ADD COLUMN IF NOT EXISTS asset_type VARCHAR(1000) NOT NULL GENERATED ALWAYS AS (COALESCE(asset_type_v1, asset_type_v2)) STORED;
```
rtso, line 3: We need to redo this migration because it has already been applied to the mainnet and testnet db.

rust/processor/src/db/postgres/schema.rs
rtso: Synced offline, this will be fixed in next PR

bowenyang007: this isn't related to this PR. only popped up due to renaming table. We'll address it next PR.

rust/processor/src/db/postgres/schema.rs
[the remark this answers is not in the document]
rtso: Synced offline, this will be fixed in next PR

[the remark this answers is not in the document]
bowenyang007: this isn't related to this PR. only popped up due to renaming table. We'll address it next PR.

rtso (approved):

PR #495, opened 2024-08-29
Simplify account_transaction_processor and use rayon to speed it up.

grao1991:
Tested locally using devnet data. See ~10x improvement for the processing part.

### DIFF
--- a/rust/processor/src/db/common/models/account_transaction_models/account_transactions.rs
+++ b/rust/processor/src/db/common/models/account_transaction_models/account_transactions.rs
@@ -13,11 +13,8 @@ use crate::{
     schema::account_transactions,
     utils::{counters::PROCESSOR_UNKNOWN_TYPE_COUNT, util::standardize_address},
 };
-use ahash::AHashMap;
-use aptos_protos::transaction::v1::{
-    transaction::TxnData, write_set_change::Change, DeleteResource, Event, Transaction,
-    WriteResource,
-};
+use ahash::AHashSet;
+use aptos_protos::transaction::v1::{transaction::TxnData, write_set_change::Change, Transaction};
 use field_count::FieldCount;
 use serde::{Deserialize, Serialize};
 
@@ -39,7 +36,7 @@ impl AccountTransaction {
     /// We will also consider transactions that the account signed or is part of a multi sig / multi agent.
     /// TODO: recursively find the parent account of an object
     /// TODO: include table items in the detection path
-    pub fn from_transaction(transaction: &Transaction) -> AHashMap<AccountTransactionPK, Self> {
+    pub fn get_accounts(transaction: &Transaction) -> AHashSet<String> {
         let txn_version = transaction.version as i64;
         let txn_data = match transaction.txn_data.as_ref() {
             Some(data) => data,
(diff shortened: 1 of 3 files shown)


rust/processor/src/db/common/models/account_transaction_models/account_transactions.rs
```diff
@@ -13,13 +13,10 @@ use crate::{
...
 };
-use ahash::AHashMap;
-use aptos_protos::transaction::v1::{
-    transaction::TxnData, write_set_change::Change, DeleteResource, Event, Transaction,
-    WriteResource,
-};
+use aptos_protos::transaction::v1::{transaction::TxnData, write_set_change::Change, Transaction};
 use field_count::FieldCount;
 use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
```
larry-aptos, line 19: nit: use AHash; my impression is it has better perf.

grao1991: Done

rust/processor/src/processors/account_transactions_processor.rs
```diff
@@ -101,20 +102,24 @@ impl ProcessorTrait for AccountTransactionsProcessor {
...
-        }
-        let mut account_transactions = account_transactions
-            .into_values()
-            .collect::<Vec<AccountTransaction>>();
-
-        // Sort by PK
-        account_transactions.sort_by(|a, b| {
-            (&a.transaction_version, &a.account_address)
-                .cmp(&(&b.transaction_version, &b.account_address))
-        });
```
rtso, line 117: Need to make sure to sort by PK to prevent db insertions from locking the same range of rows.

grao1991: The `account_transaction` is already sorted by version.

and BTW I'm not sure if the sorting actually help,
and BTW before my change we were not sorting by PK, the PK is (account_address, transaction_version), not (transaction_version, account_address)

rust/processor/src/processors/account_transactions_processor.rs
```diff
@@ -101,20 +102,24 @@ impl ProcessorTrait for AccountTransactionsProcessor {
...
+        let account_transactions: Vec<_> = transactions
+            .into_par_iter()
+            .enumerate()
+            .map(|(i, txn)| {
+                let transaction_version = start_version as i64 + i as i64;
```
rtso, line 109: We should get the version from `txn` instead of doing addition here which might be wrong.

rust/processor/src/processors/account_transactions_processor.rs
rtso: I'm not sure about the frequency of potential deadlocks, but this is a good opportunity to fix it here to sort by actual PK (account_address, transaction_version) :)

grao1991: Why do we think it will cause deadlock? It doesn't make sense.

and specifically for this case, IIUC we do have another index on (transaction_version, account_address), and postgres implementation doesn't really differentiate primary v.s. secondary index (other than enforcing unique and not null)..

rtso: If there are concurrent tasks writing to the same rows and the writes are unsorted,
1. Task 1 locks row C then tries to lock row A
2. Task 2 locks row A then tries to lock row C

That will result in deadlock

grao1991: they all touch different rows, no?

rtso (approved):

PR #500, opened 2024-09-05
[FA migration] Hasura changes for migration

bowenyang007:


rtso (approved):

hasura-api/metadata-json/README.md
```diff
@@ -1,4 +1,7 @@
 # Hasura Metadata
 This directory contains the metadata we use for Hasura. The "metadata" is what Hasura calls the file that configures pretty much everything about a Hasura instance, including table schemas, relations, DB connection configuration, how metrics are exported, etc.
 
+> [!WARNING]
+> (09/25/2024) We added a new unified_transition.json to facilitate Fungible Asset Migration. Please use this file in Hasura while the new table is backfilling.
```
rtso, line 5: This migration only affects people currently running the FA processor, so would release communication in the regular channels be enough? 

This statement might be confusing to new folks who want to run the FA processor, and we shouldn't maintain 2 metadatas here indefinitely

bowenyang007: yea true. after migration i plan to remove this.

bowenyang007: as well as the actual metadata file.

PR #501, opened 2024-09-06
[Processor] Simplify some logic to reduce unnecessary duplicated work.

grao1991:
This PR is trying to move some common logic/type/constant to resources.rs, avoid duplicate/unnecessary parsing/deserialization logic when converting between generic write resource and a specific resource type.

The main motivation is to improve the performance, no other behavior change is expected.
To give an idea on the performance impact, when running simple token minting workload, the processing time of token_v2_processor is 5x less than before.

This PR only handles FA and TokenV2 resources, leave other resources like Coin and Token for future PRs.

### DIFF
--- a/rust/processor/src/db/common/models/resources.rs
+++ b/rust/processor/src/db/common/models/resources.rs
@@ -1,0 +1,212 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::db::common::models::{
+    default_models::move_resources::MoveResource,
+    fungible_asset_models::v2_fungible_asset_utils::{
+        ConcurrentFungibleAssetBalance, ConcurrentFungibleAssetSupply, FungibleAssetMetadata,
+        FungibleAssetStore, FungibleAssetSupply,
+    },
+    object_models::v2_object_utils::{ObjectCore, Untransferable},
+    token_v2_models::v2_token_utils::{
+        AptosCollection, Collection, ConcurrentSupply, FixedSupply, PropertyMapModel,
+        TokenIdentifiers, TokenV2, UnlimitedSupply,
+    },
+};
+use anyhow::Result;
+use aptos_protos::transaction::v1::WriteResource;
+use const_format::formatcp;
+
+pub const COIN_ADDR: &str = "0x0000000000000000000000000000000000000000000000000000000000000001";
+pub const TOKEN_ADDR: &str = "0x0000000000000000000000000000000000000000000000000000000000000003";
+pub const TOKEN_V2_ADDR: &str =
+    "0x0000000000000000000000000000000000000000000000000000000000000004";
+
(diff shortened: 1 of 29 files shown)


bowenyang007:
This looks really good! Did we test this? Also don't forget to deploy before merging. For testing, maybe @rtso and @yuunlimm can help? We should at least compare the before and after, but it could also be a great case for an integration test looking for diffs.

rust/processor/src/db/common/models/resources.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn type_str() -> &'static str {
+        TYPE_FUNGIBLE_ASSET_SUPPLY
+    }
+}
+
+impl V2FungibleAssetResource {
+    pub fn from_write_resource(write_resource: &WriteResource) -> Result<Option<Self>> {
+        let type_str = MoveResource::get_outer_type_from_write_resource(write_resource);
+        Ok(Some(match type_str.as_str() {
+            TYPE_CONCURRENT_FUNGIBLE_ASSET_BALANCE => {
```
rtso, line 108: Use `ConcurrentFungibleAssetBalance::type_str()` here? Same for the other ones.

grao1991: It unfortunately doesn't compile, even it's compile time constant..

we can probably still write `x if x == ConcurrentFungibleAssetBalance::type_str() => {...}` but I feel it's ugly..

grao1991:
> This looks really good! Did we test this? Also don't forget to deploy before merging. For testing, maybe @rtso and @yuunlimm can help? We should at least compare the before and after, but it could also be a great case for an integration test looking for diffs.

I don't have a good way to get a full coverage for everything. I only manually sampled some rows and didn't see any difference before and after.

rtso:
The general code structure looks good to me, but I'm hesitant to stamp this until we do adequate testing because it's such a big change. We can either write some integration tests (@yuunlimm or @larry-aptos do either of you have bandwidth to write these tests?) or we can deploy to devnet and manually check each resource is indexed.

grao1991:
Rebased my PR on top of tests. PTAL @rtso @yuunlimm @larry-aptos

rust/processor/src/db/common/models/resources.rs
```diff
@@ -0,0 +1,212 @@
...
+
+impl<'a, T> FromWriteResource<'a> for T
+where
+    T: TryFrom<&'a WriteResource, Error = anyhow::Error> + Resource,
+{
+    fn from_write_resource(write_resource: &'a WriteResource) -> Result<Option<Self>> {
+        if MoveResource::get_outer_type_from_write_resource(write_resource) != Self::type_str() {
+            return Ok(None);
+        }
+        Ok(Some(write_resource.try_into()?))
```
rtso, line 62: Can we log the txn_version and wsc_index in case this fails? That will help us debug errors in parsing logic

rtso (approved):

PR #503, opened 2024-09-10
Fix current_aptos_names view

rtso:
## Description
There's a bug in the ANS contract where the primary name and the ANS don't point to the same registered address if a primary name token gets transfered. This caused the `current_aptos_names` view to show the wrong primary name because it joins only on `token_name`. We need to add a constraint to the left join on `current_ans_lookup_v2.registered_address = current_ans_primary_name_v2.registered_address`

Screenshot 2024-09-10 at 6 07 41 PM
Screenshot 2024-09-10 at 6 07 27 PM

## Testing
Run the SQL and see that is_primary = false 
Screenshot 2024-09-10 at 6 04 46 PM

### DIFF
--- a/rust/processor/src/db/postgres/migrations/2024-09-10-215454_ans_registered_addr/up.sql
+++ b/rust/processor/src/db/postgres/migrations/2024-09-10-215454_ans_registered_addr/up.sql
@@ -1,0 +1,46 @@
+-- Your SQL goes here
+CREATE OR REPLACE VIEW current_aptos_names AS
+SELECT cal.domain,
+  cal.subdomain,
+  cal.token_name,
+  cal.token_standard,
+  cal.registered_address,
+  cal.expiration_timestamp,
+  greatest(
+    cal.last_transaction_version,
+    capn.last_transaction_version
+  ) as last_transaction_version,
+  coalesce(not capn.is_deleted, false) as is_primary,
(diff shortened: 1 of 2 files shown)


rust/processor/src/db/postgres/migrations/2024-09-10-215454_ans_registered_addr/up.sql
```diff
@@ -0,0 +1,46 @@
...
+  concat(cal.domain, '.apt') as domain_with_suffix,
+  c.owner_address as owner_address,
+  cal.expiration_timestamp >= CURRENT_TIMESTAMP as is_active,
+  cal2.expiration_timestamp as domain_expiration_timestamp,
+  b.token_data_id as token_data_id,
+  cal.subdomain_expiration_policy as subdomain_expiration_policy
+FROM current_ans_lookup_v2 cal
+  LEFT JOIN current_ans_primary_name_v2 capn ON cal.token_name = capn.token_name
+  AND cal.token_standard = capn.token_standard
+  AND capn.registered_address = cal.registered_address
```
rtso, line 23: This is the only line that changed.

bowenyang007: add a comment in the file that this is the only line that changed?

bowenyang007 (approved):

PR #511, opened 2024-09-16
skip updating processor status with null values for timestamp

yuunlimm:
### Test Plan
Tested with parquet_ans_processor
when there is no transaction processed, log and do not update. 
Screenshot 2024-09-16 at 12 59 15 PM
Screenshot 2024-09-16 at 12 59 13 PM

when there is transaction processed, update the processor status
Screenshot 2024-09-16 at 1 06 04 PM


larry-aptos (approved):

yuunlimm:
> nit fix lint.
> 
> for existing records, will have it null until next time there are transaction processed I assume

yes that's correct!

PR #518, opened 2024-09-20
Add a feature to generate testing transactions 

yuunlimm:
### Description
linear task: https://linear.app/aptoslabs/issue/EI-521/export-generated-jsons-to-processors-repo-without-cargo-dependency

THe way we import new transactions when we are testing, we have to create a PR every time we generate new transaction JSONs to be used in the processor tests. This allows us to skip that step.

Previous Steps
1. generate txn jsons in aptos-core repo
2. create a pr
3. copy commit hash and paste it into processor repo
4. write test cases

New Steps
1. generate txn jsons in aptos-core repo (you would have to specify the output folder to the folder in processor repo here (e.g. `aptos-indexer-processor/rust/testing-transactions/json_transactions/` ).
2. write test cases

### Test Plan
used the new txn generated from this change and tests still pass
Screenshot 2024-09-20 at 2 22 20 PM

Screenshot 2024-09-20 at 2 22 23 PM


larry-aptos (approved):
this is the existing crate under aptos-core and we move here. looks good.

larry-aptos:
Can you link the linear ticket here? i think we have a task tracking this effort. thanks!

PR #522, opened 2024-09-25
Add a DB schema output diff testing.

yuunlimm:
### Description 
This is to introduce automated diff testing for all processors to detect any issues with unknown schema/indexing logic change  on non-current tables

detailed page[https://www.notion.so/aptoslabs/Schema-Diff-Feature-in-Testing-Framework-10a8b846eb7280b78960ff9403c9b0d8?pvs=4](https://www.notion.so/Schema-Diff-Feature-in-Testing-Framework-10a8b846eb7280b78960ff9403c9b0d8?pvs=21)

and by introducing this, we will have two testing approaches available:
1. DB output diff testing
2. Scenario Testing, which is to capture and test behaviors of more complex use cases. 

### Test Plan
cargo test passed


rust/integration-tests/src/diff_test_helper/processors/fungible_asset_processor.rs
```diff
@@ -0,0 +1,21 @@
...
+};
+use anyhow::Result;
+use diesel::{pg::PgConnection, query_dsl::methods::FilterDsl, ExpressionMethods, RunQueryDsl};
+use processor::schema::fungible_asset_activities::dsl::*;
+use serde_json::Value;
+
+pub struct FungibleAssetProcessorTestHelper;
+
+impl ProcessorTestHelper for FungibleAssetProcessorTestHelper {
+    fn load_data(&self, conn: &mut PgConnection, txn_version: &str) -> Result<Value> {
```
larry-aptos, line 12: do you know if the query result gets a different order? i wonder if that's going to break the test. if not can we force the order or order it before testing.

yuunlimm: unless the results are different, doesn't json_pretty re-order automatically? I can check. and if not, wecan force the ordered result when we load

larry-aptos: let's make sure the orders won't be a problem.

larry-aptos: oh in this case, only a single entry will be selected?

yuunlimm:
@larry-aptos  added additional order_by to diesel query!

rust/integration-tests/src/models/queryable_models.rs
```diff
@@ -0,0 +1,76 @@
...
+/**
+* Event model
+* this is created // b/c there is inserated_at field which isn't defined in the Event struct, we can't just load the events directly without specifying the fields.
+        // TODO: make this more generic to load all fields, then we should be able to run tests for all processor in one test case.
+
+*/
+#[derive(Clone, Debug, Deserialize, FieldCount, Identifiable, Insertable, Serialize, Queryable)]
+#[diesel(primary_key(transaction_version, event_index))]
+#[diesel(table_name = events)]
+pub struct Event {
```
larry-aptos, line 20: this looks good! Also this can benefit write query building if we need custom reading endpoint.  

another way i was thinking about is raw query like `select row_to_json(table_name) from table_name` and define `JsonResult`. And serialize it into expected db file.

larry-aptos: this can lower the complexity of the work here.

rust/integration-tests/src/diff_tests/all_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+                                    eprintln!(
+                                        "[ERROR] Error handling JSON for processor {} and transaction version {}: {}",
+                                        processor_name, txn_version, e
+                                    );
+                                    return Err(e);
+                                }
+                            };
+
+                            remove_inserted_at(&mut json_data);
+                            // Remove the Err handling around assert_json_eq!, as it doesn't return a Result type
```
larry-aptos, line 112: can we add a todo here about the json diff; this can be used for testing transaction diff but also diffing here.

rust/integration-tests/src/diff_tests/all_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+            get_expected_generated_txns,
+        )
+        .await;
+    }
+
+    // Helper function to reduce duplicate code for running tests on all processors
+    async fn run_processor_tests<F>(
+        processor_configs: Vec<TestProcessorConfig>,
+        test_context: &TestContext,
+        get_expected_json_path_fn: F,
```
larry-aptos, line 71: is it due to the lifetime that you pass closure here instead of fn? 

```
get_expected_json_path_fn: fn(&str, &str) -> String
```

rust/integration-tests/src/diff_tests/all_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+
+                            println!(
+                                "[INFO] Test passed for processor {} and transaction version: {}",
+                                processor_name, txn_version
+                            );
+                            Ok(())
+                        },
+                    )
+                    .await
+                    .unwrap();
```
larry-aptos, line 123: if one test breaks, it early returns right? do we want to finish all and get a list of results?

rust/integration-tests/src/diff_tests/all_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+                },
+            },
+            Err(e) => {
+                eprintln!("[ERROR] Failed to read file at {}: {}", path, e);
+                Err(anyhow::anyhow!("Failed to read file: {}", e))
+            },
+        }
+    }
+
+    fn get_processor_map() -> HashMap<String, Arc<Box<dyn ProcessorTestHelper>>> {
```
larry-aptos, line 145: this looks weird.. why `Arc` here?

rust/integration-tests/src/diff_tests/all_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+                        },
+                    )
+                    .await
+                    .unwrap();
+            }
+        }
+    }
+
+    // Helper function to read and parse JSON files
+    fn read_and_parse_json(path: &str) -> Result<serde_json::Value, anyhow::Error> {
```
larry-aptos, line 129: nit: anyhow::Result<serde_json::Value>

rust/integration-tests/src/lib.rs
```diff
@@ -111,19 +112,96 @@ impl TestContext {
         // Iterate over all permutations of the transaction batches to ensure that the processor can handle transactions in any order.
```
larry-aptos, line 112: if we don't do permutation, let's update the comment.

rust/Cargo.toml
```diff
@@ -33,8 +33,8 @@ anyhow = "1.0.86"
 aptos-indexer-processor-sdk = { git = "https://github.com/aptos-labs/aptos-indexer-processor-sdk.git", rev = "9ecd252ccff53023664562001dd04c2886488c0d" }
 aptos-indexer-processor-sdk-server-framework = { git = "https://github.com/aptos-labs/aptos-indexer-processor-sdk.git", rev = "9ecd252ccff53023664562001dd04c2886488c0d" }
 aptos-protos = { git = "https://github.com/aptos-labs/aptos-core.git", rev = "5c48aee129b5a141be2792ffa3d9bd0a1a61c9cb" }
-aptos-system-utils = { git = "https://github.com/aptos-labs/aptos-core.git", rev = "202bdccff2b2d333a385ae86a4fcf23e89da9f62" }
-aptos-indexer-test-transactions = { git = "https://github.com/aptos-labs/aptos-core.git", rev = "202bdccff2b2d333a385ae86a4fcf23e89da9f62" }
+aptos-system-utils = { git = "https://github.com/aptos-labs/aptos-core.git", rev = "82a8e2bfff1e822781462da83f3abd6cda3f44eb" }
```
larry-aptos, line 36: required by linter?

rust/integration-tests/src/diff_tests/all_tests.rs
yuunlimm: yeah we can early return! will change

rust/integration-tests/src/diff_tests/all_tests.rs
yuunlimm: actually, that should work too.

rust/integration-tests/src/diff_tests/all_tests.rs
yuunlimm: the closure in test_context.run requires a 'static lifetime, and the processor_map is a local variable so it doesnt have a 'static lifetime. so I used arc b/c the value itself doesn't need to be copied or cloned.. and I dont' have to implement clone for the trait and Box<dyn ProcessorTestHelper>..

rust/integration-tests/src/diff_tests/all_tests.rs
yuunlimm: yeah that makes sense

rust/integration-tests/src/models/queryable_models.rs
yuunlimm: oh yeah, we can improve this as we add more processors and tables!

rust/integration-tests/src/diff_tests/all_tests.rs
```diff
@@ -0,0 +1,182 @@
...
+        .await;
+    }
+
+    // Helper function to reduce duplicate code for running tests on all processors
+    async fn run_processor_tests(
+        processor_configs: Vec<TestProcessorConfig>,
+        test_context: &TestContext,
+        get_expected_json_path_fn: fn(&str, &str) -> String,
+    )
+    // ) where
```
larry-aptos, line 73: nit: remove

rust/integration-tests/src/diff_tests/all_tests.rs
```diff
@@ -0,0 +1,182 @@
...
+    //     F: Fn(&str, &str) -> String + Send + Sync + 'static + std::marker::Copy,
+    {
+        let processor_map = get_processor_map();
+        for processor_config in processor_configs {
+            let processor_name = processor_config.config.name();
+            let test_type = TestType::Diff(DiffTest);
+
+            if let Some(test_helper) = processor_map.get(processor_name).cloned() {
+                // let test_helper = test_helper.clone();
```
larry-aptos, line 82: remove

rust/integration-tests/src/diff_tests/all_tests.rs
larry-aptos: i feel if you make the map static, and use the reference in these places, it could work. 

not a blocking issue

rust/integration-tests/src/diff_test_helper/processors/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod event_processor;
```
larry-aptos, line 1: do you need headers here?

PR #525, opened 2024-09-27
test cross repo ci trigger

yuunlimm:


larry-aptos:
looks good; do we want to merge? or you will include more changes?

yuunlimm:
> looks good; do we want to merge? or you will include more changes?

I would just like to make sure that it's trigger first, then we can remove this and modify the existing one!

larry-aptos:
lol the aggressive linter

yuunlimm:
will use temp public repo to test

PR #528, opened 2024-10-01
allow cross repo workflow trigger (#526)

yuunlimm:
### description 

this is to allow aptos-core repo to trigger processor intregration test to support a case where json files are updated in aptos-core, we pass in commit hash along with the event, and it updates dependency using toml-cli. 

### Test Plan
PR level test pass: https://github.com/aptos-labs/aptos-indexer-processors/actions/runs/11131978344
dispatch passed: https://github.com/aptos-labs/aptos-indexer-processors/actions/runs/11131969466

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,7 +1,14 @@
 name: Run Integration Tests
 
 on:
+  repository_dispatch:
+    types: [json-change-detected]  # Custom event type to trigger the workflow
   workflow_dispatch:
+    inputs:
+      commit_hash:
+          description: 'Commit hash to use for the dependency update'
+          required: true
+          default: 'main'
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yaml
```diff
@@ -1,25 +1,59 @@
 name: Run Integration Tests
 
 on:
+  repository_dispatch:
+    types: [json-change-detected]  # Custom event type to trigger the workflow
```
rtso, line 5: nit: is it possible to make this more specific?

.github/workflows/integration-tests.yaml
```diff
@@ -1,25 +1,59 @@
...
+
+      # Update aptos-system-utils dependency using toml-cli
+      - name: Update aptos-system-utils dependency
+        if: ${{ github.event_name == 'repository_dispatch' || github.event_name == 'workflow_dispatch' }}
+        run: |
+          echo "Updating aptos-system-utils dependency in Cargo.toml to use commit hash ${{ github.event.inputs.commit_hash }}"
+          toml set rust/Cargo.toml workspace.dependencies.aptos-system-utils.rev "${{ github.event.inputs.commit_hash }}" > Cargo.tmp && mv Cargo.tmp rust/Cargo.toml
+
+      # Update aptos-indexer-test-transactions dependency using toml-cli
+      - name: Update aptos-system-utils dependency
```
rtso, line 41: forgot to update name

.github/workflows/integration-tests.yaml
```diff
@@ -1,25 +1,59 @@
...
+        if: ${{ github.event_name == 'repository_dispatch' || github.event_name == 'workflow_dispatch' }}
+        run: |
+          echo "Updating aptos-system-utils dependency in Cargo.toml to use commit hash ${{ github.event.inputs.commit_hash }}"
```
rtso, line 44: forgot to update this

rtso (approved):

.github/workflows/integration-tests.yaml
yuunlimm: how specific are you thinking? I think we can though, it's all custom

rtso: maybe something like `test-txn-json-change-detected`?

PR #533, opened 2024-10-02
update cargo.lock to fix lint failure

yuunlimm:
### Description

https://github.com/aptos-labs/aptos-indexer-processors/actions/runs/11153152318/job/31000121810

cargo lock doesn't get updated, so that triggered workflow fails at lint.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -54,6 +54,11 @@ jobs:
       - name: Show Cargo.toml After Update
         run: cat rust/Cargo.toml  # Correct path to the Cargo.toml file
 
+      # Ensure Cargo.lock is updated with the latest dependencies
+      - name: Update Dependencies
+        run: cargo update
+        working-directory: rust
+
       - name: Install Dependencies and Run Linter
         uses: ./.github/actions/dep_install_and_lint
         with:


.github/workflows/integration-tests.yaml
```diff
@@ -54,6 +54,11 @@ jobs:
       - name: Show Cargo.toml After Update
         run: cat rust/Cargo.toml  # Correct path to the Cargo.toml file
 
+      # Ensure Cargo.lock is updated with the latest dependencies
+      - name: Update Dependencies
+        run: cargo update
```
larry-aptos, line 59: this seems very risky... how to prevent new dependency from breaking changes..

yuunlimm: hmm yeah. but this doesn't override our cargo.lock, so it shouldn't be as risky. 

maybe I need to understand why it's building with the previous integration tests first.

larry-aptos: oh i see i read it in the wrong way..
i thought it's upgrade lol

larry-aptos (approved):

.github/workflows/integration-tests.yaml
yuunlimm: ohhhhh haha tyty!!

PR #536, opened 2024-10-03
update commit hash for imported txn

yuunlimm:
### Description
update the aptos-indexer-test-transactions dependency. 

it was causing the workflow failed when it's triggered with the latest change from aptos-core. because the test was referring to the old const variable, which was deprecated. 
failed job: https://github.com/aptos-labs/aptos-indexer-processors/actions/runs/11155221954/job/31005811598


just-in-chang (approved):

larry-aptos (approved):
looks good.

let's put the json diff with config as the follow-up.

PR #539, opened 2024-10-03
[parquet] Change move modules and clean up default processor

bowenyang007:
* Move module bytecode should always be there. 
* Given move module is only used in parquet, I cleaned up that file and it cascaded into refactoring the entire default processor.


yuunlimm:
Diff tests passed

yuunlimm (approved):

PR #541, opened 2024-10-04
Update to support mapping function from txn bytes 

yuunlimm:
### Description
Added a feature to add a mapping function for any scripted txn to its file name. This is used by the CLI command within the intergation-test repo to generate expected db output. 

rust code it generates looks like below:

```

                pub const IMPORTED_MAINNET_TXNS_123123_TEST_ACTIVITY: &[u8] = include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/json_transactions/imported_mainnet_txns/123123_test_activity.json"));
                
                pub const IMPORTED_MAINNET_TXNS_308783012_FA_TRANSFER: &[u8] = include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/json_transactions/imported_mainnet_txns/308783012_fa_transfer.json"));
                pub const ALL_IMPORTED_MAINNET_TXNS: &[&[u8]] = &[IMPORTED_MAINNET_TXNS_123123_TEST_ACTIVITY,IMPORTED_MAINNET_TXNS_308783012_FA_TRANSFER,];

                pub const IMPORTED_TESTNET_TXNS_5979639459_COIN_REGISTER: &[u8] = include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/json_transactions/imported_testnet_txns/5979639459_coin_register.json"));
                
                pub const IMPORTED_TESTNET_TXNS_1255836496_V2_FA_METADATA_: &[u8] = include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/json_transactions/imported_testnet_txns/1255836496_v2_fa_metadata_.json"));
                
                pub const IMPORTED_TESTNET_TXNS_5992795934_FA_ACTIVITIES: &[u8] = include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/json_transactions/imported_testnet_txns/5992795934_fa_activities.json"));
                
                pub const IMPORTED_TESTNET_TXNS_278556781_V1_COIN_REGISTER_FA_METADATA: &[u8] = include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/json_transactions/imported_testnet_txns/278556781_v1_coin_register_fa_metadata.json"));
                
[...]

PR #542, opened 2024-10-05
Yuunlimm/cli expected db

yuunlimm:
### Description
Whenever we have a new txn(s) to test with processor, we need an expected db output to compare with what we have indexed in the local db as a part of integ tests we ran. This is a cli that generated the db output so that we can manually review when the new txn is added to list of txn inputs. 

Dependent PR: https://github.com/aptos-labs/aptos-indexer-processors/pull/541

### Test Plan
Integration test passed.


rust/integration-tests/src/config.rs
```diff
@@ -0,0 +1,195 @@
...
+                    //     .collect(); // Collect into a HashMap of version -> name
+                    // test_context.transaction_batches.iter().map(|txn| {
+                    //
+                    //     // let version = txn.version.to_string();
+                    //     // let name = test_context
+                    //     //     .transaction_names
+                    //     //     .get(&version)
+                    //     //     .unwrap_or_else(|| "unknown".to_string());
+                    //     // (version, name)
+                    // }).collect::<HashMap<String, String>>(
```
rtso, line 121: Delete this?

rust/integration-tests/src/config.rs
```diff
@@ -0,0 +1,195 @@
...
+            },
+            TestProcessorConfig {
+                config: processor::processors::ProcessorConfig::TokenV2Processor(
+                    TokenV2ProcessorConfig {
+                        query_retries: 3,
+                        query_retry_delay_ms: 1000,
+                    },
+                ),
+            },
+        ];
```
rtso, line 54: Would this work for custom processors?

rtso (commented):
Should we split the output of `expected_db_output_files` into a separate folder outside of `integration-tests`? I'd imagine in the future we want to move `integration-tests` into the SDK, and the `aptos-indexer-processors` repo only contains business logic for Aptos core processors.

rust/integration-tests/src/config.rs
yuunlimm: Yes, the current structure would support adding a new custom processor, but will need to do a few steps to integrate the new processor into the framework. 

assume you have a custom processor
1.  Integrate the Custom Processor by adding your custom processor to the processor_configs vector along with the other processors like this:
```
let processor_configs = vec![
    TestProcessorConfig {
        config: processor::processors::ProcessorConfig::EventsProcessor,
    },
    TestProcessorConfig {
        config: processor::processors::ProcessorConfig::FungibleAssetProcessor,
    },
    TestProcessorConfig {
        config: processor::processors::ProcessorConfig::TokenV2Processor(
            TokenV2ProcessorConfig {
                query_retries: 3,
                query_retry_delay_ms: 1000,
            },
        ),
    },
    TestProcessorConfig {
        config: processor::processors::ProcessorConfig::MyCustomProcessor(
            MyCustomProcessorConfig {
                query_retries: 3,
                query_retry_delay_ms: 500,
            },
        ),
    },
];
```
2. In the get_processor_map() function, ensure that your new custom processor is added to the processor map with the correct name:

```
pub fn get_processor_map() -> HashMap<String, Arc<Box<dyn ProcessorTestHelper>>> {
    let mut processor_map = HashMap::new();
    // Adding existing processors
    processor_map.insert(
        "TokenV2Processor".to_string(),
        Arc::new(Box::new(TokenV2ProcessorTestHelper)),
    );
    // Adding your new custom processor
    processor_map.insert(
        "MyCustomProcessor".to_string(),
        Arc::new(Box::new(MyCustomProcessorTestHelper)),
    );
    processor_map
}
```
3. When running the run function, the custom processor will create JSON output just like the other processors. The code handling this will already work for the new processor, as it will follow the pattern that we have for other processors like TokenV2Processor. The output folder and file structure will be automatically handled by the closure inside the run method.
and we don't need to modify the logic inside run as it is processor-agnostic.

i think there is a few steps I missed, but this should work with the custom processor as well

rtso: Hmm.. is there a way to split out the config definition out from the tooling code? If the users have to update this vector, this model reverts back to the fork/clone model which SDK moves away from.

yuunlimm:
> Should we split the output of `expected_db_output_files` into a separate folder outside of `integration-tests`? I'd imagine in the future we want to move any test tooling like `integration-tests` into the SDK, and the `aptos-indexer-processors` repo only contains business logic for Aptos core processors.

agree! also, it makes to extract the config part out so that clients can easily configure and run tests without having to modify the code. I am working on the SDK integration. let me have that and this working together!

rust/integration-tests/src/config.rs
yuunlimm: agreed! let me see how I can transform this to SDK way!

yuunlimm: Synced offline: we want to provide an argument to cargo test so that they can decide to generate db output by passing in `cargo test -- --diff`, or run without `--diff` flag which is a default behavior of just running tests

PR #545, opened 2024-10-08
sdk processor test

yuunlimm:
### Description
- added tests for sdk processor
- separated expected db output folders between non-sdk and sdk processors


rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        // Initialize and run the EventsProcessor
+        let events_processor = init_events_processor(db_url.clone()).await;
+        events_processor.run_processor().await.expect("Failed to run processor");
+
+        // Connect to Postgres and verify the events
+        let mut conn = PgConnection::establish(&db_url)
+            .with_context(|| format!("Error connecting to {}", db_url)).expect("Failed to connect to database");
+
+        let events_result = events
```
yuunlimm, line 140: somehow this doesn't return any events. @larry-aptos do you have any insights how I can debug this?

rtso: It might be that the transaction didn't process yet. Since with SDK all the work is split between separate tasks, it takes some time to process. We can add `tokio::time::sleep(Duration::from_millis(250)).await` to wait to finish processing before checking the DB.

larry-aptos:
review as i am investigating the test issue.

rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        // Start the Mock GRPC server
+        tokio::spawn(async move {
+            println!("Starting Mock GRPC server");
+            mock_grpc_server.run().await;
+        });
+    }
+
+    /// Helper function to initialize the Events Processor.
+    async fn init_events_processor(db_url: String) -> EventsProcessor {
```
larry-aptos, line 80: we may want to explore some generic solutions.

rust/integration-tests/src/sdk/mock_grpc.rs
```diff
@@ -0,0 +1,89 @@
...
+        //     println!("Transaction not found, returning default response");
+        //     return Err(Status::not_found("Transaction not found"));
+        // }
+
+        let result = match transaction {
+            Some(t) => t.clone(),
+            None => {
+                // just return what we have in self.transactions
+                println!("Transaction not found, returning first transaction");
+                self.transactions[0].clone()
```
yuunlimm, line 63: here is the hack i used to not fail `check_or_update_chain_id` @rtso

rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
```diff
@@ -0,0 +1,105 @@
...
+                    let json_data = serde_json::to_string_pretty(&all_events)
+                        .expect("Failed to serialize events");
+                    assert_eq!(
+                        all_events.len(),
+                        5,
+                        "Expected 1 event, got {}",
+                        all_events.len()
+                    );
+                    println!("json data: {}", json_data);
+                    Ok(())
```
yuunlimm, line 98: I assume for the custom processor, they will have their own validation logic.

rust/integration-tests/src/sdk/mock_grpc.rs
rtso: yeah for tests, we need to mock a lot of these functions to skip the setup stuff

rust/integration-tests/src/diff_test_helper/processors/event_processor.rs
```diff
@@ -7,19 +7,26 @@ use diesel::{
...
 use processor::schema::events::dsl::*;
 use serde_json::Value;
+use test_utils::ProcessorTestHelper;
 
 // Example implementation for the EventsProcessor
 pub struct EventsProcessorTestHelper;
 
 impl ProcessorTestHelper for EventsProcessorTestHelper {
     fn load_data(&self, conn: &mut PgConnection, txn_version: &str) -> Result<Value> {
+        let events_res = diesel::sql_query("SELECT * FROM events")
```
rtso, line 17: Sycned offline: we don't need this trait and we can use test helper functions instead!

rust/integration-tests/src/diff_tests/all_tests.rs
rtso, line 1: Since we're going to reuse `cargo test` to run all processor tests, should we delete this file now?

rust/integration-tests/src/sdk/mock_grpc.rs
rtso: We can mock both `TransactionStream::get_chain_id` and `check_or_update_chain_id` since tests don't care about those.

rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
rtso: There's a mapping between processor -> config and I wonder if we can reuse it here to know which config each processor uses to instantiate the default of that config.

rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
```diff
@@ -0,0 +1,105 @@
...
+            db_pool_size: 100,
+        };
+
+        let db_config = DbConfig::PostgresConfig(postgres_config);
+
+        let indexer_processor_config = IndexerProcessorConfig {
+            processor_config,
+            transaction_stream_config,
+            db_config,
+        };
```
rtso: This can be in a default helper function

rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
rtso: How can we read the raw transaction? If I want it to compare with the DB output

rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
yuunlimm: if you are thinking of that for SDK migration, I think we can reuse that mapping for us. 
 but i don't think that we can support that for the custom processors.

rust/integration-tests/src/sdk/sdk_test_context.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        // Start the Mock GRPC server
+        tokio::spawn(async move {
+            println!("Starting Mock GRPC server");
+            mock_grpc_server.run().await;
+        });
+    }
+
+    /// Helper function to create TransactionStreamConfig and DbConfig, now part of SdkTestContext.
+    pub async fn create_transaction_and_db_config(
```
rtso, line 82: We can make this take in a `ProcessorConfig`, so the function can return a full `IndexerProcessorConfig`?

rust/integration-tests/src/sdk/sdk_test_context.rs
```diff
@@ -0,0 +1,165 @@
...
+        &self,
+        starting_version: Option<u64>,
+        ending_version: Option<u64>,
```
rtso, line 85: If there's non-sequential transactions, how would this work? I guess we could make this take in a vec of transactions and get the first, last versions. But we'd have to mock out the gap check in TransactionStreamInternal

rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
yuunlimm: i assume client will beusing the txn generator tool to generate txn json files, then they can look at the json file

rust/integration-tests/src/sdk_tests/sdk_test_helpers.rs
yuunlimm: addressed in the latest commit!

rust/integration-tests/src/sdk/sdk_test_context.rs
yuunlimm: do we expose `ProcessorConfig` through the  processor SDK?

rust/integration-tests/src/sdk/mock_grpc.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Default)]
+pub struct MockGrpcServer {
+    pub transactions: Vec<TransactionsResponse>,
+    pub chain_id: u64,
+}
+
+type ResponseStream = Pin<Box<dyn Stream<Item = Result<TransactionsResponse, Status>> + Send>>;
+
+#[tonic::async_trait]
+impl RawData for MockGrpcServer {
```
rtso, line 18: Instead of mocking the GRPC stream directly, might be easier to mock TransactionStream step's `poll` to return each transaction in the test. Then mock `should_continue_polling = false` to stop the processor.

rust/integration-tests/src/sdk/sdk_test_context.rs
yuunlimm: as a workaourdn, we could probably match starting_version and endpoint_version for each transaction stream config. let me think a bit more on this

rust/integration-tests/src/sdk/mock_grpc.rs
rtso: You can mock a sequence of values which is pretty cool. https://github.com/aptos-labs/aptos-indexer-processors-sdk/blob/343a2d01dce9f8bca76b90d506f61c1a07943e4b/aptos-indexer-processors-sdk/sdk/src/common_steps/transaction_stream_step.rs#L242

rtso (commented):
Also should we move this to a `testing-framework` crate in SDK?

yuunlimm:
> Also should we move this to a `testing-framework` crate in SDK?

yeah, should we leave the current implementation in the processor repo as it is for now. and update the tests once we start migrating?

rust/integration-tests/src/sdk/mock_grpc.rs
yuunlimm: ohh that's intersting, let me look into it

rust/integration-tests/src/sdk/sdk_test_context.rs
rtso: I think if we mock `TransactionStreamStep`, it should get around this so we don't need to specify a start/end version here.

yuunlimm: yeah that would be nicer. dumb question: how do we inject this step into the processor?

rust/integration-tests/src/sdk/sdk_test_context.rs
rtso: oh hm we actually do not expose the ProcessorConfig or DbConfig, so this is specific only to our processors. maybe we just create a TransactionStreamConfig default so that custom processors can define their own config?

rtso:
Yeah we can leave it here for now, but let's move this PR over to the SDK!

rust/integration-tests/src/sdk/sdk_test_context.rs
yuunlimm: will sdk support other db in the next release ?

yuunlimm: if we don't expose dbConfig then, it would be more diffcult for them to write all the helper functions like establishing connection for their db and write heavier validation.

rust/sdk-processor/src/db/postgres/schema.rs
rtso, line 1: Are these schema and migrations intended change? the SDK processors use the same schema/migrations as the old processors, so we probably shouldn't redefine them here.

rust/integration-tests/src/sdk_tests/sdk_tests.rs
```diff
@@ -0,0 +1,134 @@
...
+        let mut db_value = test_context
+            .run(
+                &events_processor,
+                txn_version,
+                true,   // TODO: enable a command line flag
+                custom_output_path.clone(),
+                move|db_url| {
+                    // Custom validation logic
+                    let mut conn = PgConnection::establish(&db_url).expect("Failed to establish DB connection");
+                    let test_helper = Box::new(EventsProcessorTestHelper) as Box<dyn ProcessorTestHelper>;
```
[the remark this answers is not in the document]
yuunlimm: oh this isn't required at all, we need to tune this for our internal tests

rust/sdk-processor/src/db/postgres/schema.rs
yuunlimm: oh i just copied what we have in the example repo, so probably don't need here.

rust/integration-tests/src/sdk_tests/sdk_tests.rs
```diff
@@ -0,0 +1,134 @@
+#[cfg(test)]
+mod tests {
+    use std::fs;
```
yuunlimm, line 3: @rtso @dermanyang probably only file worth taking a look for now

rust/integration-tests/src/sdk_tests/sdk_tests.rs
```diff
@@ -0,0 +1,134 @@
...
+            transaction_stream_config,
+            db_config,
+        };
+
+        let events_processor = EventsProcessor::new(indexer_processor_config)
+            .await
+            .expect("Failed to create EventsProcessor");
+
+
+        let custom_output_path = Some("custom_output_path/test_results.json".to_string());
```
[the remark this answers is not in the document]
yuunlimm: we have a predefined one for us in the run function, but clients can pass in their's if they want to

rust/integration-tests/src/sdk_tests/sdk_tests.rs
```diff
@@ -0,0 +1,134 @@
...
+    use crate::diff_test_helper::processors::event_processor::EventsProcessorTestHelper;
+    use crate::diff_test_helper::ProcessorTestHelper;
+    use crate::diff_tests::{remove_inserted_at, remove_transaction_timestamp};
+
+    #[tokio::test]
+    async fn test_run() {
+        // Step 1: set up an input transaction that will be used
+        let imported_txns = [IMPORTED_TESTNET_TXNS_1255836496_V2_FA_METADATA_];
+
+        let txn_version = 1255836496; // TOOD: more elegant way to get this
```
yuunlimm, line 28: I wonder if this can be improved so that they don't have to delcar a txn_version variable with hard-coded value.

rust/integration-tests/src/sdk_tests/sdk_tests.rs
dermanyang: are we planing on adding all tests into this file?

rust/integration-tests/src/sdk_tests/sdk_tests.rs
```diff
@@ -0,0 +1,134 @@
...
+                move|db_url| {
+                    // Custom validation logic
+                    let mut conn = PgConnection::establish(&db_url).expect("Failed to establish DB connection");
+                    let test_helper = Box::new(EventsProcessorTestHelper) as Box<dyn ProcessorTestHelper>;
+                    let json_data = match test_helper.load_data(&mut conn, &txn_version.to_string()) {
+                        Ok(events_json) => {
+                            events_json
+                        }
+                        Err(e) => {
+                            println!(
```
dermanyang, line 77: `eprintln`? to be consistent with other error prints?

rust/integration-tests/src/sdk_tests/sdk_tests.rs
```diff
@@ -0,0 +1,134 @@
...
+                println!(
+                    "[ERROR] Error handling JSON for processor {} and transaction version {}: {}",
+                    processor_name, txn_version, e
+                );
+                panic!("Failed to read and parse JSON");
+            }
+        };
+
+        // TODO: we need to enhance json diff, as we might have more complex diffs.
+        remove_inserted_at(&mut db_value);
```
dermanyang, line 108: should these be grouped into a teardown method?

rust/integration-tests/src/sdk_tests/sdk_tests.rs
yuunlimm: no there will be tests for sdk event processor only. renamed to events_processor_tests

rust/integration-tests/src/sdk_tests/test_cli_flag_util.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    // Only pass the arguments that come after "--", if it exists
+    let custom_args: Vec<String> = match clap_args_position {
+        Some(position) => raw_args[position + 1..].to_vec(), // Slice after `--`
+        None => Vec::new(), // If no `--` is found, treat as no custom args
+    };
+    println!("Custom arguments: {:?}", custom_args);
+
+    // Manually parse the "--diff" flag
+    let diff_flag = custom_args.contains(&"--diff".to_string());
```
yuunlimm, line 27: will probably change it to --generate-output or somethign else. current one is misleading. we are not diffing but generating

rust/integration-tests/src/sdk_tests/test_cli_flag_util.rs
```diff
@@ -0,0 +1,43 @@
+#[allow(dead_code)]
+#[derive(Debug, Clone)]
+pub struct TestArgs {
```
rtso, line 3: I think we can use `clap` crate to help parse the CLI args so you don't have to parse the args manually. We use this for the processors to parse the config file.

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -53,8 +53,92 @@ impl EventsProcessor {
             },
         }
     }
+    //
```
rtso, line 56: Delete this?

rust/integration-tests/src/sdk_tests/test_cli_flag_util.rs
yuunlimm: yeah i tried but it seems like it doesn't support args parsing with `cargo test` see https://github.com/rust-lang/cargo/issues/8916

PR #550, opened 2024-10-15
- support querying and validating multiple tables for existing proces…

yuunlimm:
### Description

- support querying and validating multiple tables for existing processors.
- added more tables handled by processors
- add support cli args

https://github.com/aptos-labs/aptos-indexer-processors/pull/545/files Part 1


rust/integration-tests/src/cli_parser.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[allow(dead_code)]
+#[derive(Debug, Clone)]
+pub struct TestArgs {
+    pub generate_output: bool,
+    pub output_path: Option<String>,
+}
+
+#[allow(dead_code)]
+pub fn parse_test_args() -> TestArgs {
```
dermanyang, line 26: does this helpfully complain if insufficient/bad args are provided?

yuunlimm:
test passed locally. 
```
test diff_tests::all_tests::test::test_all_scripted_txns_schema_output_for_all_processors ... ok
test diff_tests::all_tests::test::test_all_testnet_txns_schema_output_for_all_processors ... ok
test diff_tests::all_tests::test::test_all_mainnet_txns_schema_output_for_all_processors ... ok 

```

rust/integration-tests/src/cli_parser.rs
yuunlimm: one thing is that we would have to passing addition `--` to bypass failing unrecognized argument from cargo test

```
1. cargo test
2. cargo test diff_tests
3. cargo test diff_tests -- --nocapture --generate-output
4. cargo test diff_tests -- -- --generate-output <-- like this 
5. cargo test -- --nocapture
```

rust/integration-tests/src/diff_tests/mod.rs
```diff
@@ -25,25 +25,37 @@ pub fn remove_transaction_timestamp(value: &mut Value) {
 }
 
 #[allow(dead_code)]
```
rtso, line 27: Could we put all these in a `mod tests`

rust/integration-tests/src/diff_tests/all_tests.rs
```diff
@@ -171,4 +212,55 @@ mod test {
             },
         ]
     }
+
+    // Helper function to generate output files for each table
```
rtso, line 216: These helpers are already part of the SDK testing framework, we could reuse them rather than redefine here

rust/integration-tests/src/cli_parser.rs
```diff
@@ -0,0 +1,52 @@
...
+});
+
+// Function to fetch global test args
+#[allow(dead_code)]
+pub fn get_test_config() -> (bool, Option<String>) {
+    let test_args = TEST_CONFIG.lock().unwrap().clone();
+    (test_args.generate_output, test_args.output_path)
+}
+
+pub fn parse_test_args() -> TestArgs {
```
rtso, line 24: Could we also reuse the one defined in the SDK?

yuunlimm: which function are you referring to?

rtso: Oh I thought parse_test_args is part of the SDK. I feel like it should be part of the testing framework SDK, wdyt?

yuunlimm: yeah also thought about that whether we should let them decide to use their own way of passing those args (generate-files and custom_path), b/c they might want more customzied way. but we can expose first and gather their feedback

yuunlimm: https://github.com/aptos-labs/aptos-indexer-processor-sdk/pull/72

rust/integration-tests/src/diff_tests/mod.rs
yuunlimm: this prevents us from importing : https://github.com/aptos-labs/aptos-indexer-processor-sdk/pull/73

PR #553, opened 2024-10-16
Update existing processor tests with the latest dependency change

yuunlimm:
moving https://github.com/aptos-labs/aptos-indexer-processors/pull/550 into graphite

PR #554, opened 2024-10-16
Add Sdk tests

yuunlimm:
https://github.com/aptos-labs/aptos-indexer-processors/pull/545


yuunlimm:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-indexer-processors/554?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#554** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-indexer-processors/554?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#553** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-indexer-processors/553?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @yuunlimm and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

rust/integration-tests/src/sdk_tests/temp.rs
```diff
@@ -0,0 +1,140 @@
...
+        processors::fungible_asset_processor::FungibleAssetProcessor,
+    };
+    use std::{collections::HashSet};
+    use crate::sdk_tests::run_processor_test;
+    use crate::sdk_tests::get_test_config;
+    use crate::sdk_tests::DEFAULT_OUTPUT_FOLDER;
+    use std::collections::HashMap;
+
+    #[tokio::test]
+    async fn fa_processor_db_output_diff_test() {
```
rtso, line 38: What's the difference between `fa_processor_db_output_diff_test` here and the one in `fungible_asset_processor`?

rust/integration-tests/src/sdk_tests/events_processor_tests.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    // This test cases runs the events processor and validates the output of all available transactions proto jsons
+    #[tokio::test]
+    async fn testnet_events_processor_db_output_diff_test() {
+        let (diff_flag, custom_output_path) = get_test_config();
+        let output_path = custom_output_path
+            .unwrap_or_else(|| DEFAULT_OUTPUT_FOLDER.to_string() + "imported_testnet_txns");
+
+        // Step 1: set up an input transaction that will be used
+        let transaction_batches = ALL_IMPORTED_TESTNET_TXNS
```
rtso, line 32: Instead of running the events processor test on every imported transaction, would it be better to let the user decide?

I liked the devex we had before where for each cargo test, you defined which transactions you want to use.  Running each processor test on all imported transactions would create a bunch of empty files / no-ops and aren't super useful.

dermanyang: +1 what's the reason to run `ALL_IMPORTED_TESTNET_TXNS`? Can this just be `EVENTS_PROCESSOR_IMPORTED_TESTNET_TXNS`?

rust/integration-tests/src/sdk_tests/temp.rs
dermanyang, line 1: What's the context behind this file being `temp`?

rust/integration-tests/src/sdk_tests/events_processor_tests.rs
yuunlimm: they are the next milestones, that test case is there to make sure that we can process them without any issues. 
more test cases with purpose and context could be added as a part of the next milestone: 90% coverage

rust/integration-tests/src/sdk_tests/temp.rs
yuunlimm: this file should've included. removed

rust/integration-tests/src/sdk_tests/events_processor_tests.rs
rtso: Testing each processor with `ALL_IMPORTED_TESTNET_TXNS` can get us to a high % test coverage goal for sure, but you lose the context associated with the transaction and the test, and I'm not sure how debuggable these are. 

I think we should be importing transactions and testing intentionally as part of first milestone. The bulk of the work is just identifying which transactions to use, right? Since we have this framework now it should be really easy to add tests.

dermanyang: > that test case is there to make sure that we can process them without any issues.
Can we instead then do `[EVENTS_PROCESSOR_IMPORTED_TESTNET_TXNS + IRRELEVANT_TXN]` with a comment so that this intent is clear?

yuunlimm: you are absolutely right, most of the effort now lies in identifying which transactions to use for testing specific scenarios. and using `ALL_IMPORTED_TESTNET_TXNS` can certainly help us reach the initial test coverage milestone faster. It’s a _valuable proof of concept_ that shows our system is correctly indexing large sets of real-world data, even if it lacks the intentionality and debuggability of more focused tests.


here’s how I think we can position this:

1. ALL_IMPORTED_TESTNET_TXNS as a Milestone:
- It serves as a strong baseline, proving that we can handle and index transactions correctly. Even though it doesn't focus on specific scenarios, it's a great starting point when **we didn’t have any coverage.**
- By running all imported transactions, we can show that our system processes real-world data as expected. This is useful as an overall system health check.

2. Next Milestones with More Intentional Tests:
- Once we have the initial proof of correct indexing, the next milestone can focus on intentional testing, which will bring in more context and debugability.
- We can target specific edge cases, error conditions, and SEVs that have been critical in the past, aligning this with the goal of reaching 90% coverage.

This gives us a strong base with test coverage and indexing proof, while also allowing room to refine and add higher-quality tests in the next phases. Would this align with your thoughts?

dermanyang: @yuunlimm I hear ya here . Let's leave a TODO for the next milestone so we don't forget to update this later?

rust/integration-tests/src/sdk_tests/test_cli_flag_util.rs
rtso, line 1: This is not needed / used anywhere

rtso (approved):
Stamping to unblock, let's add the todo that @dermanyang suggested!

PR #559, opened 2024-10-21
Update SDK Tests using the latest change in mock grpc

yuunlimm:
### Description
update tests based on the latest change in sdk testing framework where it stubs out the txn version in mock grpc to avoid having gaps.

TODO:
- naming structure of files for diffing all txns, whether we keep using the original version as the folder name(preferred) or not.

### Test Plan
Screenshot 2024-10-21 at 8 33 10 AM


yuunlimm:
* **#559** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-indexer-processors/559?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @yuunlimm and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

rust/integration-tests/src/sdk_tests/events_processor_tests.rs
```diff
@@ -49,23 +49,41 @@ mod tests {
...
+        pub const ALL_IMPORTED_TESTNET_TXNS: &[&[u8]] = &[
+            IMPORTED_TESTNET_TXNS_1_GENESIS,
+            IMPORTED_TESTNET_TXNS_2_NEW_BLOCK_EVENT,
+            IMPORTED_TESTNET_TXNS_3_EMPTY_TXN,
+            IMPORTED_TESTNET_TXNS_278556781_V1_COIN_REGISTER_FA_METADATA,
+            IMPORTED_TESTNET_TXNS_1255836496_V2_FA_METADATA_,
+            IMPORTED_TESTNET_TXNS_5523474016_VALIDATOR_TXN,
+            IMPORTED_TESTNET_TXNS_5979639459_COIN_REGISTER,
+            IMPORTED_TESTNET_TXNS_5992795934_FA_ACTIVITIES,
+        ];
```
rtso, line 85: For the intentional tests, what are your thoughts on splitting these transactions up into separate test cases? Right now the test name is `testnet_events_processor_db_output_diff_test` but we could make it more contextual and include what scenario it's actually testing.

rust/integration-tests/src/sdk_tests/mod.rs
```diff
@@ -7,15 +7,19 @@ use aptos_indexer_testing_framework::{
...
+};
 
 #[cfg(test)]
 pub mod events_processor_tests;
 #[cfg(test)]
 pub mod fungible_asset_processor_tests;
 
 #[allow(dead_code)]
-pub const DEFAULT_OUTPUT_FOLDER: &str = "sdk_expected_db_output_files/";
+pub const DEFAULT_OUTPUT_FOLDER: &str = "sdk_expected_db_output_files2";
```
rtso, line 22: Did you mean to change this?

rust/integration-tests/sdk_expected_db_output_files2/imported_testnet_txns/fungible_asset_processor/8/coin_supply.json
```diff
@@ -1,10 +1,10 @@
 [
   {
-    "transaction_version": 21,
+    "transaction_version": 8,
```
rtso, line 3: Maybe we even stub out transaction_version, transaction_timestamp, and transaction_epoch, since these are not relevant to the test. Can add as a TODO

rtso (approved):
Minor comments, stamping to unblock!

rust/integration-tests/src/sdk_tests/events_processor_tests.rs
```diff
@@ -137,10 +156,7 @@ mod tests {
             .map(|txn| serde_json::from_slice(txn).expect("Failed to deserialize transaction"))
             .collect();
 
-        let starting_version = transaction_batches
-            .first()
-            .expect("No transactions found")
-            .version;
+        let starting_version = 1;
```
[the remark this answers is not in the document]
yuunlimm: yeah that makes sense, I also thought it would be a devEx gap. then, it would also makes more sense to name the folder with original version for us too, when the custom file name isn't specified.

rust/integration-tests/src/sdk_tests/events_processor_tests.rs
yuunlimm: yeah i can split this into a separate test case.

rust/integration-tests/src/sdk_tests/mod.rs
yuunlimm: I wanted to show what the new folder structure would look like so that we can compare. However, since we are going to keep the original version as a folder name, you can ignore this!

rust/integration-tests/sdk_expected_db_output_files2/imported_testnet_txns/fungible_asset_processor/8/coin_supply.json
yuunlimm: yeah wanted to hear your opinion on these, wasn't sure if they would really care about those.

rust/integration-tests/src/sdk_tests/events_processor_tests.rs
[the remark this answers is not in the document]
yuunlimm: one downside of stubbing out txn_versions would confuse users when they write their validation logic. we would have to tell them to use version 1, 2, 3, to query db.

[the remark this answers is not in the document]
rtso: hmm what if instead of stubbing, the mock grpc just returns the transactions sorted in order of version?

rust/integration-tests/sdk_expected_db_output_files2/imported_testnet_txns/fungible_asset_processor/8/coin_supply.json
rtso: I think your point of using the txn_version to debug a failed test is valid, so we should probably keep for now.

rust/integration-tests/src/sdk_tests/events_processor_tests.rs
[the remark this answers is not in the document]
yuunlimm: so then user uses [txn3, txn1, txn5] as an input, then grpc retruns [txn1, txn3, txn5]?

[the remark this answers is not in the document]
rtso: Yeah exactly. I guess we have to choose between 
1. whether mock grpc allows gaps in the transaction batches 
2. stubbing out transaction versions -> you lose contextual information about the transaction 
3. require that scenario test txn's have no gaps (this is hard since there's no guarantee when submitting txn's that there won't be a gap)

Taking a step back, I think the goal here is to be able to test multiple txn's in a single test, so I think #1 could work here. The risk I see with #1 is if the user is writing a multi-txn test and their processor panics if it sees gaps within a txn batch. I think this is a minimal risk; the user could fallback to using single-txn test case.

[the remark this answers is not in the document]
yuunlimm: yeah the goal of this test is to processors not to test grpc, so it makes sense to allow mock grpc with gaps.

PR #560, opened 2024-10-22
Migrate Events Processor to Use New Version Tracker Impl

dermanyang:
### Copied changes
Apply recent changes improving restart behavior from `aptos-indexer-processor-example` repo. Namely:
1. Refactor the `events_processor` to use `VersionTracker` step instead of the `LatestVersionProcessedTracker` object ([original PR](https://github.com/aptos-labs/aptos-indexer-processor-example/pull/8))
2. Create a `BackfillProcessorStatus` table and write to it when `backfill_alias` is set in the config ([original PR](https://github.com/aptos-labs/aptos-indexer-processor-example/pull/6))
3. Add new column, `BackfillStatus`, and improve `starting_version` selection on restart ([original PR](https://github.com/aptos-labs/aptos-indexer-processor-example/pull/10))
See original PRs for more details and discussion. Design [here] (https://www.notion.so/aptoslabs/Improve-Processor-Restart-and-Gap-Detection-Behavior-9e884917f4df4f5690d09f8a59581aab).

### Other changes
-  Grouped the new migration/schema with the other migrations/schemas under `rust/processor` rather than `rust/sdk-processor`. But added the model under `rust/sdk-processor/db/common/models`. 
- Updated certain `execute...()` functions to return `Result<usize, ProcessorError>` instead of `QueryResult<usize>`. A change that has been in the example repo for a while. 
- Small changes to keep the new integration tests happy given the change to the SDK processor config. 

## Remaining Work
- Migrate FA processor to use this too + delete `LatestVersionProcessedTracker` 
- Merge this [infra PR](https://github.com/aptos-labs/internal-ops/pull/4395) to leave the pods in `Complete` when the indexer reaches the `end_version` instead of crash looping lol

#### New Cargo Tests for Starting Version
image

#### Regular Indexer Start
Starts from last check pointed version:
image
 Starts from max(last check pointed, starting_version):
image

#### Backfill Indexer
image

#### Non-SDK Default Processor Sanity Check
image


rust/processor/src/db/postgres/schema.rs
```diff
@@ -837,8 +851,8 @@ diesel::table! {
         supply_aggregator_table_key_v1 -> Nullable<Text>,
         #[max_length = 10]
         token_standard -> Varchar,
-        is_token_v2 -> Nullable<Bool>,
         inserted_at -> Timestamp,
+        is_token_v2 -> Nullable<Bool>,
```
dermanyang, line 855: This needed to be changed around otherwise whenever the migrations are re-ran, the ordering of the produced struct's constructor changes causing some issues.

yuunlimm: to add a bit of context: this is reverting the change I made manually. this order should be always the same unless we change in sql file.

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -1,52 +1,278 @@
...
 
-    // If latest_processed_version is not stored in DB, return the default 0
-    Ok(0)
+    // If nothing checkpointed, return the `starting_version` from the config, or 0 if not set.
+    Ok(latest_processed_version.unwrap_or(
+        indexer_processor_config
+            .transaction_stream_config
+            .starting_version
+            .unwrap_or(0),
+    ))
```
rtso, line 37: This changes the existing priority of the flags for regular processors which can make it hard if you're developing locally and want to start from a specific version. Could we keep the order for regular processor?

dermanyang: you mean the use case where you want `starting_version` to have higher precedence than last_processed_version? 

That's a trade-off we have to make to have better restart behavior. Otherwise, the regular processor will always restart at the starting_version if that's ever set

dermanyang: I could add the old `starting_version_override` config option lol. But rename it to `starting_version_dev_override` to make it clear its for development purposes so its easier for local development. Wdyt?

rtso: What if we keep the priority the same for regular processors to match with existing devex, and establish a different priority only for backfill processors?

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -1,52 +1,278 @@
...
+            &mut conn,
+        )
+        .await
+        .context("Failed to query backfill_processor_status table.")?;
+
+        // Return None if there is no checkpoint or if the backfill is old (complete).
+        // Otherwise, return the checkpointed version + 1.
+        return Ok(
+            backfill_status.and_then(|status| match status.backfill_status {
+                BackfillStatus::InProgress => Some(status.last_success_version as u64 + 1),
```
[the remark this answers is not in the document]
dermanyang: I have to leave in the `+1`s unfortunately since you can't factor it out from the different logical branches. I renamed the function to `get_starting_version_from_db` to correct the misnomer.

rust/sdk-processor/src/steps/common/processor_status_saver.rs
```diff
@@ -0,0 +1,144 @@
...
+                let status = BackfillProcessorStatus {
+                    backfill_alias: backfill_alias.clone(),
+                    backfill_status,
+                    last_success_version: lst_success_version,
+                    last_transaction_timestamp: end_timestamp,
+                    backfill_start_version: backfill_start_version.unwrap_or(0) as i64,
+
+                    backfill_end_version: backfill_end_version
+                        .unwrap_or(last_success_batch.metadata.end_version)
+                        as i64,
```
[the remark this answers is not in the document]
dermanyang: good call out. This way, it'll be clear that the BFP will run forever rather than incorrectly indicating that the `last_processed_version` is equal to the `backfill_end_version` and status being `in_progress`.

rust/sdk-processor/src/processors/events_processor.rs
```diff
@@ -106,10 +105,12 @@ impl ProcessorTrait for EventsProcessor {
         let events_storer = EventsStorer::new(self.db_pool.clone(), processor_config);
         let order_step = OrderByVersionStep::new(
             starting_version,
-            Duration::from_secs(UPDATE_PROCESSOR_STATUS_SECS),
+            Duration::from_secs(DEFAULT_UPDATE_PROCESSOR_STATUS_SECS),
+        );
```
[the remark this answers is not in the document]
dermanyang: removed.

rust/sdk-processor/src/db/common/models/backfill_processor_status.rs
```diff
@@ -0,0 +1,86 @@
...
+    // #[diesel(rename = "in_progress")]
+    InProgress,
+    // #[diesel(rename = "complete")]
+    Complete,
+}
+
+impl ToSql<Text, Pg> for BackfillStatus {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> serialize::Result {
+        match *self {
+            BackfillStatus::InProgress => out.write_all(b"in_progress")?,
```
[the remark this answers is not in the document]
dermanyang: done

rust/sdk-processor/src/utils/starting_version.rs
dermanyang: > head processors in prod should never need to use starting_version in the config

how come? if we made a logic change, wouldn't the steps be something like
1. create a backfill processor from 0 to HEAD
2. create a regular processor from HEAD onwards

to do #2, wouldn't we use `starting_version=HEAD` where HEAD is just some recent version?

rtso: Good point! I was thinking about the current processors already running in prod, but if you're a new processor and you set `starting_version=HEAD` then going forward, it should only need to refer to the latest version from DB. 

And to my point of developing locally, they can also use the `backfill_processor` config to start from arbitrary version!

rtso (approved):

rust/sdk-processor/src/utils/starting_version.rs
dermanyang: awesome. thanks for the discussion! yeah, or even run diesel reset to wipe the state

PR #563, opened 2024-10-24
remove null address from delegator count

ying-w:
When a delegator first receives a delegation in the UI it will show as having 3 active delegators.

ex. https://explorer.aptoslabs.com/txn/1828020008/changes?network=mainnet

The 3 table items with balance are

1. reward address
2. NULL address
3. delegator

I've changed the tooltip to reflect that reward address will be included https://github.com/aptos-labs/explorer/pull/836

This logic change removes the NULL address from being counted

Context on NULL address https://github.com/aptos-labs/aptos-core/blob/114d361d6cc7119f88eec11ae562c5ec6797275a/aptos-move/framework/aptos-framework/sources/delegation_pool.move#L1568

### DIFF
--- a/rust/processor/src/db/postgres/migrations/2024-10-23-232703_num_active_delegator_per_pool/down.sql
+++ b/rust/processor/src/db/postgres/migrations/2024-10-23-232703_num_active_delegator_per_pool/down.sql
@@ -1,0 +1,1 @@
+DROP VIEW IF EXISTS num_active_delegator_per_pool;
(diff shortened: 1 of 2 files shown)


bowenyang007 (approved):
looks good. @larry-aptos please help deploy thanks!

rust/processor/src/db/postgres/migrations/2024-10-23-232703_num_active_delegator_per_pool/down.sql
```diff
@@ -0,0 +1 @@
+DROP VIEW IF EXISTS num_active_delegator_per_pool;
```
bowenyang007, line 1: nit, down should just be the previous query right? Or is this a new view?

larry-aptos:
sure! I’ll deploy tmr morning.

PR #571, opened 2024-10-28
Reset LastSuccessfulVersion to 0 if Restarting on Stauts=Complete

dermanyang:
## Purpose
If the backfillStatus is Complete on a particular row, while the processor will correctly start to backfill from 0 again, the table will not reflect the current status due to the monotonic increasing invariant set on the table writes. A workaround was to simply remove that invariant. 

This PR adds back that invariant to avoid possible edge cases and instead resets the latestSuccessVersion to 0 in the `starting_version` logic upon encountering the `Complete` state.

## Testing
Manually tested expected behavior:
image

image

image

### DIFF
--- a/rust/sdk-processor/src/steps/common/processor_status_saver.rs
+++ b/rust/sdk-processor/src/steps/common/processor_status_saver.rs
@@ -140,7 +140,7 @@ impl ProcessorStatusSaver for ProcessorStatusSaverEnum {
                             backfill_processor_status::backfill_end_version
                                 .eq(excluded(backfill_processor_status::backfill_end_version)),
                         )),
-                    None,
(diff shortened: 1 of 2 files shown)


yuunlimm:
qq: what specific edge cases led to the removal of the invariant initially?

dermanyang:
@yuunlimm I initially removed the `WHERE` clause assuming that given correct processing order and `starting_version` logic (namely, choosing the MAX of start_ver/latest_processing or restarting given a Completed state) would be sufficient. 

But I remembered that we don't actually enforce in-order events so it would be possible to have weird in_progress states that don't actually reflect the processor's progress lol. So I added it back for sanity.

rtso (approved):

yuunlimm (approved):

PR #576, opened 2024-10-29
Rewrite Stake Processor with SDK

dermanyang:
## Purpose
- Refactor existing stake processor to locate all parsing logic in `parse_stake_data`.
- Rewrite stake processor using the sdk
- add new, non-default ProcessorConfig called StakeProcessorConfig that contains additional fields to the default one. Namely, query_retries, query_retry_delay_ms.

## Testing
image

### DIFF
--- a/rust/sdk-processor/src/steps/stake_processor/stake_storer.rs
+++ b/rust/sdk-processor/src/steps/stake_processor/stake_storer.rs
@@ -185,6 +185,6 @@ impl AsyncStep for StakeStorer {}
 
 impl NamedStep for StakeStorer {
     fn name(&self) -> String {
-        "FungibleAssetStorer".to_string()
+        "StakeStorer".to_string()
     }
 }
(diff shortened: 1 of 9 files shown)


rust/sdk-processor/src/steps/stake_processor/stake_storer.rs
```diff
@@ -0,0 +1,190 @@
...
+            Vec<DelegatorPoolBalance>,
+            Vec<CurrentDelegatorPoolBalance>,
+            Vec<CurrentDelegatedVoter>,
+        )>,
+    ) -> Result<Option<TransactionContext<Self::Output>>, ProcessorError> {
+        let per_table_chunk_sizes: AHashMap<String, usize> = self
+            .processor_config
+            .default_config
+            .per_table_chunk_sizes
+            .clone();
```
rtso, line 82: Now that I think about it, I think we should move this part to the processor so we avoid the clone every time process is called?

rust/sdk-processor/src/steps/stake_processor/stake_storer.rs
```diff
@@ -0,0 +1,190 @@
...
+            metadata: input.metadata,
+        }))
+    }
+}
+
+impl AsyncStep for StakeStorer {}
+
+impl NamedStep for StakeStorer {
+    fn name(&self) -> String {
+        "FungibleAssetStorer".to_string()
```
rtso, line 188: Change name

rtso (approved):

PR #577, opened 2024-10-29
Add Default Processor Tests for table_items, current_table_items, and block_metadata_transactions

dermanyang:
## Purpose
as above

### Old Coverage Stats:
image

### New Coverage Stats:
image

### Current Table Items
image

image

### Table Items
image
image

### BMT
image

image


rust/integration-tests/src/diff_test_helper/default_processor.rs
```diff
@@ -32,22 +35,14 @@ pub fn load_data(
...
-    // let cti_result = cti_dsl::current_table_items
-    //     .filter(cti_dsl::last_transaction_version.eq_any(&txn_versions))
-    //     .then_order_by(cti_dsl::last_transaction_version.asc())
-    //     .load::<CurrentTableItem>(conn)?;
-    // result_map.insert(
-    //     "current_table_items".to_string(),
-    //     serde_json::to_value(&cti_result)?,
-    // );
-
-    // TODO: revisit we need to find out a way to find correct table metadata
```
dermanyang, line 44: do we still need this table metadata? It's not part of the default processor

rtso (approved):

rust/integration-tests/src/diff_test_helper/default_processor.rs
yuunlimm: The `create_table_item` event, which generates table_metadata for a table_item, is part of the default processor. To test this, we need to identify the correct transaction version containing the `create_table_item` event, as `table_metadata` records doesn't have a transaction_version field. For the test case, we should either pull all rows to verify the presence of the expected data since filtering by transaction version isn’t an option for us here

PR #585, opened 2024-11-02
Migrate NFT Metadata Processor to SDK

just-in-chang:
also took out some clones


dermanyang (approved):

yuunlimm:
is this tested? if then, could you attach any snapshots?

PR #604, opened 2024-11-13
Add README to SDK Dir and Update the Legacy README

dermanyang:
### Purpose
Include information about which coin type mappings are supported. See [thread](https://aptos-org.slack.com/archives/C03MN5F7WUV/p1731394170711589)

Keeping the new readme pretty bare. Let me know if you guys think another format is better


rust/sdk-processor/README.md
```diff
@@ -0,0 +1,13 @@
+# Indexer GRPC Parsers (SDK)
+
+Indexer GRPC parser is to indexer data processor that leverages the indexer grpc data. These processors have been (re)-written using the new SDK.
```
rtso, line 3: I'd reword this a bit. Also "Indexer GRPC" is a name used internally. Externally it's known in the docs as "Transaction Stream", and we need to update this everywhere lol 

```suggestion
Processors that index data from the Aptos Transaction Stream (GRPC). These processors have been (re)-written using the new Indexer SDK.
```

rtso (approved):

rtso:
Thanks for making this change!

PR #627, opened 2024-12-06
fix token claims

rtso:
## Description 
When a token gets claimed, processor needs to know the previous owner for token_activities_v2. Currently the processor gets it from PendingClaimsResource but in this case the resource wasn't found in the transaction and breaks the processor. This PR tries to get it previous owner from the TokenClaim event instead of the table metadata. 

## Testing
Screenshot 2024-12-06 at 12 44 57 PM


rtso:
* **#627** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-indexer-processors/627?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-indexer-processors/627?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

dermanyang (approved):
code lgtm. can you explain the testing screenshot please? i don't understand how it tests the new logic

rtso:
@dermanyang token v2 processor is failing in devnet with version 19922017 so to test the fix, I ran it with the fix locally and checked that 1) the processor didn't panic and 2) the data is correct. also added an integration test for this case.

PR #629, opened 2024-12-09
Migrate Parquet User Transaction Processor

dermanyang:
### Purpose
Migrate user transaction processor to use the SDK

Note: models were not refactored to use the raw model pattern as `Signature` need a schema update first. It was difficult to use the raw pattern on just `UserTransactions` due to the dependency on `Signature` and came out pretty messy. Will leave the raw refactor until after `Signature` gets updated. 

### DQ Sanity Check
image
image


rust/processor/src/processors/parquet_processors/parquet_user_transactions_processor.rs
```diff
@@ -78,6 +78,52 @@ impl Debug for ParquetUserTransactionsProcessor {
         )
     }
 }
+pub async fn process_transactions(
```
yuunlimm, line 81: nit: new line

yuunlimm:
could you also check if the processor_status table was updated correctly?

dermanyang:
@yuunlimm confirming that `processor_Status` was updated correctly:
image

rust/processor/src/processors/parquet_processors/parquet_user_transactions_processor.rs
dermanyang: weird that the formatter didn't get this

yuunlimm: yeah formatter sometimes misses, `cargo +nightly fmt  && cargo +nightly xclippy` I run this just in case.

yuunlimm:
oh could you lastly check if it's able to pick up the last success version correctly when it restarts? other than this, lgtm

yuunlimm (approved):

PR #639, opened 2024-12-11
Return None when no entry found in processor_status

dermanyang:
### Purpose
This will now use `starting_version` when there's no checkpoint in `processor_status`.

### Testing
image

### DIFF
--- a/rust/sdk-processor/src/utils/starting_version.rs
+++ b/rust/sdk-processor/src/utils/starting_version.rs
@@ -115,7 +115,7 @@ async fn get_min_processed_version_from_db(
                     Some(status.last_success_version as u64)
                 },
                 // Handle specific cases where `Ok` contains `None` (no status found)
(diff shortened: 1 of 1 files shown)


yuunlimm (approved):

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -115,7 +115,7 @@ async fn get_min_processed_version_from_db(
                     Some(status.last_success_version as u64)
                 },
                 // Handle specific cases where `Ok` contains `None` (no status found)
-                Ok(None) => Some(0),
+                Ok(None) => None,
```
yuunlimm, line 118: should we add a log here?

dermanyang: imo fine not to since processors already print/log the starting version

PR #656, opened 2025-01-07
Add columns to `user_transactions` table

0xmaayan:
Following https://aptos-org.slack.com/archives/C03MN5F7WUV/p1734539332759199 , adding `contract_address`, `module_name`, and `function_name` columns to the `user_transactions` table to index each of the entry function string data separately. 

image

### DIFF
--- a/rust/integration-tests/sdk_expected_db_output_files/user_transaction_processor/test_single_sender_webauth/user_transactions.json
+++ b/rust/integration-tests/sdk_expected_db_output_files/user_transaction_processor/test_single_sender_webauth/user_transactions.json
@@ -10,6 +10,9 @@
     "gas_unit_price": "100",
     "timestamp": "2024-04-13T17:59:30.751467",
     "entry_function_id_str": "0x1::aptos_account::transfer_coins",
-    "epoch": 6620
+    "epoch": 6620,
+    "contract_address": "0x1",
+    "module_name": "aptos_account",
+    "function_name": "transfer_coins"
   }
 ]
--- a/rust/integration-tests/sdk_expected_db_output_files/user_transaction_processor/test_single_sender_webauth/user_transactions.json
+++ b/rust/integration-tests/sdk_expected_db_output_files/user_transaction_processor/test_single_sender_webauth/user_transactions.json
@@ -11,8 +11,8 @@
     "timestamp": "2024-04-13T17:59:30.751467",
     "entry_function_id_str": "0x1::aptos_account::transfer_coins",
     "epoch": 6620,
-    "contract_address": "0x0000000000000000000000000000000000000000000000000000000000000001",
-    "module_name": "aptos_account",
-    "function_name": "transfer_coins"
+    "entry_function_contract_address": "0x0000000000000000000000000000000000000000000000000000000000000001",
+    "entry_function_module_name": "aptos_account",
+    "entry_function_function_name": "transfer_coins"
   }
 ]
--- a/rust/integration-tests/sdk_expected_db_output_files/user_transaction_processor/test_single_sender_webauth/user_transactions.json
+++ b/rust/integration-tests/sdk_expected_db_output_files/user_transaction_processor/test_single_sender_webauth/user_transactions.json
@@ -11,7 +11,7 @@
     "timestamp": "2024-04-13T17:59:30.751467",
     "entry_function_id_str": "0x1::aptos_account::transfer_coins",
     "epoch": 6620,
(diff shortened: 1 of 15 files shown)


rust/integration-tests/sdk_expected_db_output_files/user_transaction_processor/test_single_sender_webauth/user_transactions.json
```diff
@@ -10,6 +10,9 @@
     "gas_unit_price": "100",
     "timestamp": "2024-04-13T17:59:30.751467",
     "entry_function_id_str": "0x1::aptos_account::transfer_coins",
-    "epoch": 6620
+    "epoch": 6620,
+    "contract_address": "0x1",
```
rtso, line 14: we should standardize the contract address so it's always length 64. there's a helper you can use - `standardize_address`

rust/processor/src/db/postgres/migrations/2025-01-07-000504_add_function_columns_to_user_transactions/up.sql
```diff
@@ -0,0 +1,5 @@
+-- Your SQL goes here
+ALTER TABLE user_transactions
+ADD COLUMN contract_address VARCHAR(66) NOT NULL,
```
rtso, line 3: to make sure the queries that filter by these fields are efficient, we need to add an index to the DB. example: 
```
CREATE INDEX IF NOT EXISTS ta2_owner_type_index ON token_activities_v2 (event_account_address, type);
```

rust/processor/src/utils/util.rs
```diff
@@ -147,6 +147,34 @@ pub fn get_entry_function_from_user_request(
     entry_function_id_str.map(|s| truncate_str(&s, MAX_ENTRY_FUNCTION_LENGTH))
 }
 
+pub fn get_entry_function_contract_address_from_user_request(
+    user_request: &UserTransactionRequest,
+) -> Option<String> {
+    get_entry_function_from_user_request(user_request).and_then(|s| {
```
rtso, line 153: `get_entry_function_from_user_request` truncates the string to a max length, so getting these 3 fields using this method of splitting might return weird results if the entry function is too long. 

The better way would be to map the fields directly from the `EntryFunctionId` struct in the `Transaction`:
 https://github.com/aptos-labs/aptos-core/blob/main/protos/proto/aptos/transaction/v1/transaction.proto#L318

Similar to how `entry_function_id_str` is extracted: https://github.com/aptos-labs/aptos-indexer-processors/blob/3c6291318485454afcb3fd762b78797066e646a3/rust/processor/src/utils/util.rs#L128

rust/processor/src/db/postgres/migrations/2025-01-07-000504_add_function_columns_to_user_transactions/up.sql
0xmaayan: do we need to add index each of the fields? something like
```
CREATE INDEX IF NOT EXISTS user_transactions_contract_info_index ON user_transactions (contract_address, module_name, function_name);
```

rtso: yeah a multi column index will be more efficient in this case. I imagine the query pattern will utilize a subset of these 3 columns, starting with the leading (leftmost) columns.

rtso:
nit on column naming - can we prefix the new columns with `entry_function_` to show they're all related?

0xmaayan:
> nit on column naming - can we prefix the new columns with `entry_function_` to show they're all related?

done

rust/processor/src/db/postgres/migrations/2025-01-07-000504_add_function_columns_to_user_transactions/down.sql
```diff
@@ -0,0 +1,5 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE user_transactions
```
rtso, line 2: need to also drop the index in down.sql

rtso (approved):

rust/processor/src/db/postgres/migrations/2025-01-07-000504_add_function_columns_to_user_transactions/down.sql
0xmaayan: done

PR #668, opened 2025-01-14
Retain order for property maps

CapCap:
Ensure that the property maps from tokens maintain their order, to prevent deltas between different versions due to different keys (but with same data)

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1710,7 +1710,7 @@ dependencies = [
  "futures-sink",
  "futures-util",
  "http 0.2.12",
- "indexmap 2.6.0",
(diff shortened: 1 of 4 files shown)


just-in-chang (approved):

larry-aptos (approved):

just-in-chang:
Wonder if `BTreeMap` would be better here? It'd make the order even more deterministic (ord trait instead of just insertion order)

larry-aptos:
> Wonder if `BTreeMap` would be better here? It'd make the order even more deterministic (ord trait instead of just insertion order)

I assume we want to preserve the ordering from PropertyMap; for BTreeMap, we need to define the ordering for each key

PR #688, opened 2025-01-22
[FA migration] Make storage id consistent for transaction tables

bowenyang007:
Even those are the same assets storage id is different. 
image

## Backfill
Testnet: 0
Mainnet: 0
This affects fungible_asset_activities and fungible_asset_balances (for analytics)

## Test
Storage id is now the same!
image


yuunlimm:
I think integration test might fail,

rust/processor/src/db/common/models/fungible_asset_models/raw_v2_fungible_asset_activities.rs
```diff
@@ -194,9 +197,10 @@ impl RawFungibleAssetActivity {
                 }
             };
 
-            let storage_id =
-                CoinInfoType::get_storage_id(coin_type.as_str(), owner_address.as_str());
-
+            // Storage id should be derived (for the FA migration)
+            let metadata_addr = get_paired_metadata_address(&coin_type);
+            let storage_id = get_primary_fungible_store_address(&owner_address, &metadata_addr)
```
rtso, line 202: what if the coin hasn't migrated yet? does it matter what this value is?

rust/processor/src/db/common/models/fungible_asset_models/raw_v2_fungible_asset_balances.rs
```diff
@@ -318,8 +336,10 @@ impl RawFungibleAssetBalance {
             );
             if let Some(coin_type) = coin_info_type.get_coin_type_below_max() {
                 let owner_address = standardize_address(delete_resource.address.as_str());
-                let storage_id =
-                    CoinInfoType::get_storage_id(coin_type.as_str(), owner_address.as_str());
+                // Storage id should be derived (for the FA migration)
+                let metadata_addr = get_paired_metadata_address(&coin_type);
+                let storage_id = get_primary_fungible_store_address(&owner_address, &metadata_addr)
+                    .expect("calculate primary fungible store failed");
```
rtso, line 342: this replaces all coin balances with a new primary key. how do we delete the old ones so they don't show up twice in the table?

rtso:
@yuunlimm we also should enable the cfab tests (non-legacy) so this PR has enough testing

yuunlimm:
> @yuunlimm we also should enable the cfab tests (non-legacy) so this PR has enough testing
https://github.com/aptos-labs/aptos-indexer-processors/pull/690/files#diff-fdebff6c735ac6b5581af94ffb8f889ef988a3dfa320e880a504fefa60f71fba added here

rust/processor/src/db/common/models/fungible_asset_models/raw_v2_fungible_asset_balances.rs
rtso: @yuunlimm correct me if i'm wrong but i think this is also used in the postgres processors

yuunlimm: it's the legacy table, right? then it was deprecated so the code can be removed

rtso: oh i see, the RawCurrentFungibleAssetBalance is already using `get_primary_fungible_store_address`

rust/processor/src/db/common/models/fungible_asset_models/raw_v2_fungible_asset_activities.rs
rtso: makes sense. and storage_id can be arbitrary for coin balances since it's only used for the PK of the table.

yuunlimm (approved):

PR #691, opened 2025-01-22
[FA migration] When coinstore is deleted we miss the FA activities

bowenyang007:
This is a temporary solution until we have a better event that has both the event address and the coin type so that we can map withdraw / deposit events to the proper coin type.

Caveat, if there are multiple coinstore deletions in the same account, in the same transaction, we will get this wrong. 

## Backfill
Testnet: 
Mainnet: 

## Testing
May need to add an integration test for this

Before we were missing apt withdraw
image

Now we have it!
image


rust/processor/src/processors/fungible_asset_processor.rs
```diff
@@ -571,7 +571,11 @@ pub async fn parse_v2_coin(
             // This is because v1 events (deposit/withdraw) don't have coin type so the only way is to match
             // the event to the resource using the event guid
             let mut event_to_v1_coin_type: EventToCoinType = AHashMap::new();
-
+            // When coinstore is deleted we have no way of getting the mapping but hoping that there is
+            // only 1 coinstore deletion by owner address. This is a mapping between owner address and deleted coin type
+            // This is not ideal as we're assuming that there is only 1 coinstore deletion by owner address, this should be
+            // replaced by an event (although we still need to keep this mapping because blockchain)
```
ying-w, line 577: `CoinStoreDeletion` will replace `CoinEventHandleDeletion` and include CoinStore https://github.com/aptos-labs/aptos-core/pull/15730

rtso (approved):

rust/processor/src/processors/fungible_asset_processor.rs
bowenyang007: Great. I'm not using the events now so will just wait for that event to be on testnet and add the patch.

yuunlimm (approved):

PR #692, opened 2025-01-23
add coin_store_deletion_event info to event_to_coin_type map

lightmark:
Add the to event id to coin type mapping.


rust/processor/src/processors/fungible_asset_processor.rs
```diff
@@ -683,6 +683,35 @@ pub async fn parse_v2_coin(
                 fungible_asset_activities.push(gas_event);
             }
 
+            // The CoinStoreDeletionEvent, only need for v1 with migration
+            if user_request.is_some() {
+                let coin_store_deletion_event = events.iter().find_map(|event| {
+                    let event_type = event.type_str.as_str();
+                    CoinStoreDeletionEvent::from_event(event_type, &event.data, txn_version)
```
rtso, line 690: what if there are multiple deletion events in the same txn?

lightmark: nice catch. I updated.

lightmark:
what happened here? @rtso @bowenyang007

rtso:
not sure if we still need this @bowenyang007

PR #699, opened 2025-01-27
Refactor events models

rtso:
### TL;DR
Refactored event processing to use a common RawEvent model with dedicated conversion traits for Postgres and Parquet outputs.

### What changed?
- Introduced `ParquetConvertible` and `PostgresConvertible` traits for standardized model conversions
- Consolidated event parsing into `RawEvent::from_transaction()` with a helper function: `from_event`
- Removed duplicate event processing code across processor and SDK
- Simplified event processing in both Postgres and Parquet processors by leveraging the new traits

### Testing
- Added tests for validator txn with missing events


rtso:
* **#699** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-indexer-processors/699?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-indexer-processors/699?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

rust/processor/src/db/parquet/models/event_models/parquet_events.rs
```diff
@@ -60,157 +54,24 @@ impl GetTimeStamp for Event {
...
-            });
-            Event::from_null_event(
-                txn_version,
-                block_height,
-                index as i64,
-                size_info,
-                block_timestamp,
-            )
-        })
-        .collect()
```
rtso, line 195: @yuunlimm @ying-w i removed this because if there's no events, then we don't care about event_size_info. confirming that this is intended.

rust/processor/src/db/common/models/event_models/raw_events.rs
```diff
@@ -21,12 +26,63 @@ pub struct RawEvent {
...
+                    block_height,
+                    index as i64,
+                    size_info,
+                    Some(block_timestamp),
+                )
+            })
+            .collect::<Vec<Self>>()
+    }
+
+    pub fn from_event(
```
yuunlimm, line 85: I wonder if this is being used elsewhere other than in the raw_event file, if not we can make it non public

rust/processor/src/db/parquet/models/event_models/parquet_events.rs
```diff
@@ -60,157 +54,24 @@ impl GetTimeStamp for Event {
...
+            sequence_number: self.sequence_number,
+            creation_number: self.creation_number,
+            block_height: self.transaction_block_height,
+            event_type: self.type_.clone(),
+            data: self.data.clone(),
+            event_index: self.event_index,
+            indexed_type: self.indexed_type.clone(),
+            type_tag_bytes: self.type_tag_bytes.unwrap_or(0),
+            total_bytes: self.total_bytes.unwrap_or(0),
             event_version: EVENT_VERSION,
```
yuunlimm, line 73: Hmm I thought event_v2 was already added by Aaron, I wonder what this has to be set then

rust/processor/src/db/parquet/parquet_convertible.rs
```diff
@@ -0,0 +1,12 @@
+pub trait ParquetConvertible {
+    type ParquetModelType;
+    fn to_parquet(&self) -> Self::ParquetModelType;
+}
+
+impl<T: ParquetConvertible> ParquetConvertible for Vec<T> {
```
yuunlimm, line 6: Nice!!

rust/processor/src/db/common/models/event_models/raw_events.rs
rtso: good point!

rust/processor/src/db/parquet/models/event_models/parquet_events.rs
yuunlimm: yea this shouldn't affect anything, should we add a test with that edge case just in case? and  @ying-w is there any filter job during merge for null event?

rtso: do you have a transaction that has null events and non-null event size info?

yuunlimm: you can use this one: https://explorer.aptoslabs.com/txn/554229017/changes?network=mainnet

ying-w: I'd rather have the inconsistency (either in n_events in transactions or event_size_info) so it is possible to flag that an event is missing. The above example occurs because we didn't fully backfill validator transactions for grpc (only went back to 970304805 on mainnet)

rtso: ok then we'll need to update how we get transactions.n_events. it should be based off of len(event_size_info), which is the source of truth, instead of len(events). i can make the change in this PR. @yuunlimm would you be able to help with backfill?

yuunlimm: yeah I can help with the backfil

rust/processor/src/db/parquet/parquet_convertible.rs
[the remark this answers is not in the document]
rtso: to convert models in a vector, we can do: 
```
let converted: Vec<EventPQ> = events
.into_iter()
.map(|e| e.into())
.collect()
```

rust/processor/src/db/common/models/event_models/raw_events.rs
```diff
@@ -21,12 +26,63 @@ pub struct RawEvent {
     pub total_bytes: Option<i64>,
 }
 
-pub trait EventConvertible {
-    fn from_raw(raw_item: &RawEvent) -> Self;
-}
-
 impl RawEvent {
-    pub fn from_raw_event(
+    pub fn from_transaction(txn: &Transaction, processor_name: &str) -> Vec<RawEvent> {
```
bowenyang007, line 30: this is not efficient right? let's say i want to process all events and specific events in the same processor, i'd have to loop through events twice b/c the way we're writing the code?

rtso: yeah agree, let's avoid the paradigm where models are parsed from the transaction (ie `Event::from_transaction`) and use functions instead (ie `parse_events(txn)`). for processors that are parsing multiple models the second way is more efficient.

PR #700, opened 2025-01-29
Make backfill behavior more explicit 

dermanyang:
## Purpose
Make restart behavior more explicit through the introduction of `TestingConfig`, `BootstrapConfig` and `ProcessorMode` on `IndexerProcessorConfig`. The `starting_version` on `transaction_stream_config` should no longer be used, instead there are new fields `initial_starting_version` on the backfill + bootstrap configs, and `override_starting_version` on testing config. 

### Behavior
The current live behavior of taking the higher value of `transaction_stream_config.starting_version` is moved to `bootstrap_config.initial_starting_version`. This is the default behavior of the processor if `ProcessorMode` is not specified. Just as before, if the respective starting version is not provided, the processor will begin from genesis. 

`BackfillConfig.intial_starting_version` will be respected when any of the following are true:
- there is no checkpoint in `backfill_processor_status`
- the last backfill has been complete (`last_successful_version >= backfill_config.ending_version`)
- `initial_starting_version > last_successful_version`. 
- `backfill_config.overwrite_checkpoint` is set to `true`
Otherwise, the processor will select `last_successful_version`. 

## Technical Bits
Using pattern matching on `IndexerProcessorConfig.processor_mode` for different logic in `starting_version.rs`. All new fields are Optional using Serde to default `processor_mode` to default when not provided. Serde will throw if `mode == testing` or `mode == backfill` and the respective sub-config isn't provided.

Make start and end version on `backfill_config` required. Remove `backfill_alias` field in favor of `backfill_id`. This will be appended onto the processor type as a PK in the `backfill_processor_status` table.

## Testing
Added tests for new behavior 

#### Mode not set, no checkpoint
image

#### Mode not set, checkpoint higher than bootstrap start ver (ran after the above run)
image

#### mode = backfill with no checkpoint
image

#### mode = backfill, checkpoint higher than `backfill_config.intial_starting_version`
image

#### mode = testing, checkpoint higher than `testing_config.override_starting_version`. latter is still picked
image
image


rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -51,20 +148,25 @@ pub async fn get_min_last_success_version_parquet(
...
 
     let config_starting_version = indexer_processor_config
-        .transaction_stream_config
-        .starting_version
-        .unwrap_or(0);
+        .bootstrap_config
+        .clone()
+        .map_or(0, |config| config.initial_starting_version);
 
     if let Some(min_processed_version) = min_processed_version {
```
dermanyang, line 171: @yuunlimm can I get your eyes on this logic in particular
thank you!!

rust/sdk-processor/src/config/indexer_processor_config.rs
```diff
@@ -155,5 +226,21 @@ impl RunnableConfig for IndexerProcessorConfig {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct BackfillConfig {
-    pub backfill_alias: String,
+    pub backfill_id: String,
```
yuunlimm, line 229: is this the primary key we use to query backfill process?

rust/sdk-processor/src/config/indexer_processor_config.rs
```diff
@@ -155,5 +226,21 @@ impl RunnableConfig for IndexerProcessorConfig {
...
+    pub initial_starting_version: u64,
```
yuunlimm, line 230: nit: if there isn't any other starting_version, can this be just starting_version?

rust/sdk-processor/src/db/common/models/backfill_processor_status.rs
```diff
@@ -72,14 +72,16 @@ pub struct BackfillProcessorStatusQuery {
...
 }
 
 impl BackfillProcessorStatusQuery {
     pub async fn get_by_processor(
-        backfill_alias: &str,
+        processor_type: &str,
+        backfill_id: &str,
         conn: &mut DbPoolConnection<'_>,
     ) -> diesel::QueryResult<Option<Self>> {
+        let backfill_alias = format!("{}_{}", processor_type, backfill_id);
```
yuunlimm, line 84: oh I see append backfill_id to the processor type and this is the key we use. what's reason for this change?

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -15,31 +15,128 @@ use processor::schema::backfill_processor_status;
...
 ///
-/// If this is a backfill processor and threre is an in-progress backfill, this will return
+/// If this is a backfill processor and there is an in-progress backfill, this will return
 /// the checkpointed version + 1.
 ///
 /// If this is a backfill processor and there is not an in-progress backfill (i.e., no checkpoint or
-/// backfill status is COMPLETE), this will return `starting_version` from the config, or 0 if not set.
+/// backfill status is COMPLETE), this will return `intial_starting_version` from the backfill config, or 0 if not set.
+///
+/// If the backfill status is IN_PROGRESS, and `backfill_config.overwrite_checkpoint` is `true`, this will return
```
yuunlimm, line 27: would this ever in a state where indefinite restarting and overwriting occurs, would there be a way for us to prevent this?

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -15,31 +15,128 @@ use processor::schema::backfill_processor_status;
...
+        ProcessorMode::Testing => {
+            // Always start from the override_starting_version.
+            let testing_config = indexer_processor_config.testing_config.clone().unwrap();
+            Ok(testing_config.override_starting_version)
+        },
+    }
+}
+
+fn log_ascii_warning(version: u64) {
+    println!(r#"
```
yuunlimm, line 128: lol nice!

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -15,31 +15,128 @@ use processor::schema::backfill_processor_status;
 
 /// Get the appropriate starting version for the processor.
 ///
-/// If it is a regular processor, this will return the higher of the checkpointed version,
-/// or `staring_version` from the config, or 0 if not set.
+/// If it is a regular (`mode == default`) processor, this will return the higher of the (checkpointed version,
+/// `initial_starting_version` || 0) from the bootstrap config.
```
yuunlimm, line 19: q: what purpose does intial_starting_version serve?

rust/sdk-processor/src/config/indexer_processor_config.rs
```diff
@@ -155,5 +226,21 @@ impl RunnableConfig for IndexerProcessorConfig {
...
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct BootStrapConfig {
+    pub initial_starting_version: u64,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct TestingConfig {
```
yuunlimm, line 243: q: what's expected behavior of. processor using testing config?

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -15,31 +15,128 @@ use processor::schema::backfill_processor_status;
...
+            let default_starting_version = indexer_processor_config
+                .bootstrap_config
+                .clone()
+                .map_or(0, |config| config.initial_starting_version);
+
+            Ok(status.map_or(default_starting_version, |status| {
+                std::cmp::max(status.last_success_version as u64, default_starting_version)
+            }))
+        },
+        ProcessorMode::Testing => {
```
yuunlimm, line 119: nit: should we not differentiate the behavior of processor as testing is to test the actual behavior of processors

rust/sdk-processor/src/config/indexer_processor_config.rs
dermanyang: I want to make it clear that it's the initial starting version, since if the backfill restarts with a checkpoint in `backfill_processor_status` that is larger than `initial_starting_version`, it'll pick the former

rust/sdk-processor/src/db/common/models/backfill_processor_status.rs
dermanyang: so that when monitoring the backfill processor table, its clear what the processor type is! lmk if you have a better name than `backfill_id`

rust/sdk-processor/src/utils/starting_version.rs
dermanyang: I don't think so-- unless I'm missing something.

If the job is COMPLETE and `overwrite_checkpoint` is true, the job will just restart until its done which is actually the same behavior as it is today

rust/sdk-processor/src/utils/starting_version.rs
dermanyang: this was fun haha

rust/sdk-processor/src/utils/starting_version.rs
dermanyang: it's the same purpose as it is today, i just prepended it with `initial` to make it explicit that it'll only be respected on initial startup and `get_starting_version()` will take the larger of the config's starting version and the checkpointed version

rust/sdk-processor/src/config/indexer_processor_config.rs
dermanyang: testing config is a way to get the processor to always start at the same `starting_version`. I'll add some comments around these configs in this file to be clear

rust/sdk-processor/src/utils/starting_version.rs
dermanyang: This is simlar to a 'development' mode-- so when you're writing the processor, you don't have to worry about what is checkpointed in postgres and you can reliably restart from the same version for debugging purposes

rust/processor/parser.yaml
```diff
@@ -5,6 +5,6 @@ health_check_port: 8084
 server_config:
   processor_config:
     type: default_processor
-  postgres_connection_string: postgresql://postgres:@localhost:5432/default_processor
-  indexer_grpc_data_service_address: http://127.0.0.1:50051
-  auth_token: AUTH_TOKEN
\ No newline at end of file
+    postgres_connection_string: postgresql://postgres:@localhost:5432/default_processor
```
bowenyang007, line 8: Do we need to move these into the processor config?

rust/sdk-processor/src/db/common/models/backfill_processor_status.rs
bowenyang007: regardless of what we call it the idea is that you'd have fa_1, fa_2, fa_3, ...

rust/sdk-processor/src/utils/starting_version.rs
bowenyang007: What happens when the job is complete? Does the server auto restarts?

rust/sdk-processor/src/utils/starting_version.rs
bowenyang007: the reality is that for parquet we'd always only have a single backfill at any given moment right? woudln't multiple create issues for downstream?

bowenyang007 (commented):
Where is the updated readme? 

had some questions. once confirmed should be g2g. thanks for addressing this!

rust/sdk-processor/src/utils/starting_version.rs
yuunlimm: k8b job doesn't restart once it's complete

rust/processor/parser.yaml
dermanyang: this is the legacy processor config-- removed.

rust/sdk-processor/src/utils/starting_version.rs
```diff
@@ -15,31 +15,131 @@ use processor::schema::backfill_processor_status;
...
+                &backfill_config.backfill_id,
+                &mut conn,
+            )
+            .await
+            .context("Failed to query backfill_processor_status table.")?;
+            // Return None if there is no checkpoint, if the backfill is old (complete), or if overwrite_checkpoint is true.
+            // Otherwise, return the checkpointed version + 1.
+            if let Some(status) = backfill_status_option {
+                // If status is Complete or overwrite_checkpoint is true, this is the start of a new backfill job.
+                if status.backfill_status == BackfillStatus::Complete
```
rtso, line 51: if it's already complete, do we still want to overwrite if overwrite_checkpoint = false or not set? this could lead to unintended backfills

rtso (commented):
left a few comments but devex lgtm! would you mind also updating the [README.md](https://github.com/aptos-labs/aptos-indexer-processors/blob/main/rust/sdk-processor/README.md#L0-L1)?

rust/sdk-processor/src/utils/starting_version.rs
bowenyang007: does it just kinda die?

rust/sdk-processor/README.md
```diff
@@ -40,11 +49,25 @@ If you want to index a custom contract, we recommend using the [Quickstart Guide
     - `type`: which processor to run
     - `channel_size`: size of channel in between steps
     - Individual processors may have different configuration required. See the full list of configs [here](https://github.com/aptos-labs/aptos-indexer-processors/blob/main/rust/sdk-processor/src/config/processor_config.rs#L89).
+
+- `backfill_config`
+    - `backfill_id`: appended to `processor_type` for a unique backfill identifier
+    - `initial_starting_version`: processor starts here unless there is a greater checkpointed version
+    - `ending_version`: ending version of the backfill
```
bowenyang007, line 56: this is now required correct?

rust/sdk-processor/README.md
```diff
@@ -40,11 +49,25 @@ If you want to index a custom contract, we recommend using the [Quickstart Guide
...
+    - `overwrite_checkpoint`: overwrite checkpoints if it exists, restarting the backfill from `initial_starting_version`.
```
bowenyang007, line 57: wouldn't this lead to crashlooping behavior?

rust/sdk-processor/README.md
dermanyang: the config itself is optional but the fields are required. 

```
pub struct IndexerProcessorConfig {
    ...
    pub backfill_config: Option<BackfillConfig>,
    pub bootstrap_config: Option<BootStrapConfig>,
    pub testing_config: Option<TestingConfig>,
}

pub struct BackfillConfig {
    pub backfill_id: String,
    pub initial_starting_version: u64,
    pub ending_version: u64,
    pub overwrite_checkpoint: bool,
}
...
```

rust/sdk-processor/README.md
dermanyang: we either have crash looping when someone sets `overwrite_checkpoint = true`, or we don't have a way to in progress backfills. I think if someone needs to restart an in-progress BF, they're able to pay attention and fix the wrong behavior. 

What do you think?

rust/sdk-processor/src/utils/starting_version.rs
yuunlimm: It doesn't die, but the job's status is set to `complete`

dermanyang: it'll set the job's status to complete and once the transaction_stream reaches its ending version it will terminate which will cause an exit 0 from the processor

yuunlimm (commented):
left minor questions, overall lgtm!

yuunlimm (approved):

rust/sdk-processor/README.md
```diff
@@ -23,10 +23,19 @@ If you want to index a custom contract, we recommend using the [Quickstart Guide
...
+        backfill_config:
+            backfill_id: "123"
+            initial_starting_version: 42
+            ending_version: 100042
+            overwrite_checkpoint: false
+        testing_config:
+            override_starting_version: 111111
+            ending_version: 222222
+        bootstrap_config:
+            initial_starting_version: 123456789
```
rtso, line 35: are only 1 of these configs set at a time? if so, might be confusing to include all three of these in the template. could we also add comments here for when to use each mode?

PR #701, opened 2025-01-29
fix clippy warning

rex4539:


rex4539:
Tried something different. Let's see if that works. If not, I will close the PR.

PR #702, opened 2025-02-01
fix asset_type_v1 null issue

bowenyang007:
## Summary
Not sure why diesel is being weird here and overriding field with null even though `treat_none_as_null` is false. 

## Backfill
When first v2 only apt was created.
Testnet: 2598047902
Mainnet: 1669115601

## Test
asset_type_v1 was null before
image


bowenyang007:
Not sure what's happening here. Locally these tests pass. Are the flaky or did something actually change @yuunlimm ?

rtso (approved):

rtso:
@bowenyang007 @yuunlimm they're passing bc there aren't any tests that process multiple txn's, but we should add them as fast follow

PR #705, opened 2025-02-06
Refactor and fix auth key formula in account resortation processor

heliuchuan:


rust/processor/src/db/postgres/models/account_restoration_models/account_restoration_utils.rs
```diff
@@ -28,8 +28,7 @@ impl AuthKeyScheme for Ed25519AuthKeyScheme {
     const SCHEME: u8 = 0x00;
 
     fn auth_key(&self) -> Option<String> {
-        let mut preimage = self.public_key.clone();
-        preimage.push(Self::SCHEME);
```
0xjunha, line 32: Shouldn't the preimage include SCHEME suffix here?

rust/processor/src/db/postgres/models/account_restoration_models/account_restoration_utils.rs
```diff
@@ -89,6 +88,20 @@ impl AnyPublicKeyType {
...
 }
 
+fn get_auth_key_preimage_for_public_key(
+    key_type: &AnyPublicKeyType,
+    public_key: Vec<u8>,
+) -> Vec<u8> {
+    let mut preimage = vec![key_type.to_u8()];
+    match key_type {
+        AnyPublicKeyType::Ed25519 => preimage.push(0x20),
+        AnyPublicKeyType::Secp256k1Ecdsa => preimage.push(0x41),
```
0xjunha, line 98: Why do we need `0x20` or `0x41` here?

0xjunha (commented):
Thanks for the refactor! @heliuchuan 
Utilizing `AccountSignature` and adding FeePayer support looks good to me.
Regarding auth key derivation logic, left some comments.

rust/processor/src/db/postgres/models/account_restoration_models/account_restoration_utils.rs
heliuchuan: the legacy ed25519 does not use scheme.  scheme was added to abstract to diff auth schemes.

0xjunha: https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L765
https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L744

Doesn't this mean the legacy ed25519 includes the `0x00` scheme?

rust/processor/src/db/postgres/models/account_restoration_models/account_restoration_utils.rs
0xjunha: Gotcha, just to confirm - is the length prefix part of BCS encoding

yuunlimm:
@heliuchuan could you make this change in this repo https://github.com/aptos-labs/aptos-indexer-processors-v2. we are trying to deprecate this repo starting next week

rust/processor/src/db/postgres/models/account_restoration_models/account_restoration_utils.rs
heliuchuan: It unfortunately is not standardize between schemes.  But it is for Ed25519 (which I believe is just a byproduct of the library used to serialize the pubkey) and Secp256k1Ecdsa (which is actually part of the standard way to represent the pubkey).  

Keyless does not.

rust/processor/src/db/postgres/models/account_restoration_models/account_restoration_utils.rs
heliuchuan: hmm yeah you are totally right.  i messed up my test somehow. fixed

PR #718, opened 2025-03-24
update fa metdata schema

yuunlimm:


rtso:
The migration name needs to be a direct copy of the ones we have in v2, otherwise diesel will think they're 2 different migrations and will apply them again when the processors get upgraded to the v2 repo.

rtso (changes requested):
requesting changes to fix migration naming

rust/processor/src/db/postgres/migrations/2025-01-23-164631_coin_infos_for_migration/up.sql
rtso, line 1: why deleting this?

yuunlimm: added back!

rtso (approved):

